PR #3, opened 2018-07-06
Adjustments for DB interface modifications

Timo614:
Pushing out this branch as PR 17 for the DB repo requires these changes as the interface is modified.

https://github.com/big-neon/bn-db/pull/17

The responses from the db crate are wrapped in a DatabaseError response which either returns Ok with the record or Err with the error. I've put some basic error response in place but it's just to get this branch to pass with the new changes from PR 17. Will swing back for better error responses from the API later just didn't want to leave master broken if we end up merging the db side.


Timo614:
This branch now has the tests running for the artists controller and the test db properly working. I ended up using Rcs to handle it as they worked well with the need to share the connection for the test environment. It's still somewhat rough though so will clean some of this logic up.

One thing to note about the branch. I ended up getting the test logic working for the unit tests of the actix-web logic but there are a lot of problems I ran into into the process:

- The [TestRequest](https://actix.rs/actix-web/actix_web/test/struct.TestRequest.html) logic isn't setup to handle extractors at all at this time from what I could gather going through the code. The logic uses an empty router and uses the passed in handler to the `run` method to pass a HttpRequest into the given method [src](https://github.com/actix/actix-web/blob/master/src/test.rs#L621). I ended up using little shim methods to handle the extracting to allow us to still use the API but it doesn't feel clean.
- Unfortunately TestRequest also requires static strings for path parameters. I ended up using a somewhat hacky approach of passing in a valid uuid and then swapping it out for the expected one in the closure I am using but that's definitely not ideal. Trying to figure out a better solution for this now as I've been a bit stumped looking through the code / API.

mikethetike:
I'm not entirely sure about what approach the testing should take yet. Also, not entirely sure what is the best approach for handling errors before they go to the client. I want to get the `master` back to building again and working on Travis, so I think let's push it in

Issue #7, opened 2018-07-09
A Visitor should be able to login to the platform using their email address and a password


CjS77:
From an API perspective, we should allow the following authentication procedures:
* use session cookies / JWT for stateless authentication (this ticket)
* OAuth
* Public key & signature (direct wallet authentication -- not for MVP)

Timo614:
> use session cookies / JWT for stateless authentication (this ticket)

Were you thinking of supporting both methods? So JWT with an authorization bearer header and session cookies? Or were you thinking of using JWT and then storing that in the cookie?

Actix-web has a middleware we can use to securely store session data:
https://actix.rs/book/actix-web/sec-9-middlewares.html#user-sessions

Issue #21, opened 2018-07-09
Permission enforcing middleware


CjS77:
The check whether a given user has the rights to execute a given endpoint should be implemented as middleware, so that a 403 or 401 can be returned early on in the process, and other endpoint logic can proceed with the assumption that authorization has been done already.

PR #23, opened 2018-07-17
Changes to support API login & refactored test logic

Timo614:
Login branch to deal with #7 and #11.

Changes in the branch:
- Added login and logout actions for session creation (using the session controller's restful create and destroy functions)
- Added cookie based session storage via middleware and implemented a solution for testing purposes
- Refactored the test logic to make it cleaner and reduce the somewhat hacky nature of it

The bulk of the work for this branch went toward the test refactoring as I had to find some way to surface the session logic. It's much cleaner now as I'm no longer relying on the built in `run` logic and using a closure to somewhat hack it to work.

Note: this branch will fail on travis and locally unless you change `Cargo.toml` to reference your local db `bigneon_db = { path = "../bn-db" }` and have your local db pointing to my branch bn-api-issues-7+11-sessions-modifications from my fork.


Timo614:
Added https://github.com/big-neon/bn-db/pull/24 for adding the User::find method used by the extractor I added.

I confirmed it worked locally adding it to a method. Just add it like any other extractor and if the user is logged in it'll pass the logged in user to the method.

It panics if the user is not found or if the token is missing. I think that should be fine since #21 mentions:

> other endpoint logic can proceed with the assumption that authorization has been done already

So that middleware will redirect people as needed if they are not logged in or lack the right authorization. That middleware can use this extractor to fetch the session as well:

````rust
    use actix_web::FromRequest; // trait definition
    let user = User::extract(&request);
    ...
````

Timo614:
I've made one more modification adding a response to the session::create action on login to return a subset of User fields. I added it to the db in progress branch. I selected the ones that made the most sense to me but we can edit as we iterate.

For now I didn't make the `for_display` function (name open to changes) a trait as most models won't need to hide their innards during display and can be serialized fully but since User included the hashed password I wanted to make sure I didn't include that in the API response.

Timo614:
I think this branch should be good for review if anyone can take a look.

Still waiting to hear back from Cayle at the moment regarding JWT tokens. We may want to merge this and #24 if he ends up mentioning we do want both header and cookie based auth. Had left a message in #7 as I wasn't sure. I think it shouldn't be too bad to merge the two branches together although we may want to use this one as a base given the large test changes (the other branch made some modifications as well but ended up completely changing how we run them in this one to remove some of the hackiness).

mikethetike:
I agree with @Timo614 that we should use this PR as a base, but I think that we will most likely need to  use JWT's instead of cookie based auth for the android and iOs apps.

PR #24, opened 2018-07-17
Auth middleware token for #21 

mikethetike:
This puts the calls in place to obtain a JWT via username and password. There should be other mechanisms in place at a later stage, but this work allows us to create a token that can be inspected in the middleware, and then as @Timo614 suggested, made available via an extractor.


mikethetike:
I've merged in @Timo614 's changes and re-organized the routing to apply the middleware. I am not yet checking any permissions, just that the user has a valid token.

src/bin/server.rs
```diff
@@ -1,9 +1,12 @@
 extern crate bigneon_api;
+extern crate dotenv;
 
 use bigneon_api::config::{Config, Environment};
 use bigneon_api::server::Server;
+use dotenv::dotenv;
 
 fn main() {
+    dotenv().ok();
```
Timo614: Is this for some issue that arose with it? Or leftover from an earlier commit?

Not a big deal either way just quickly looking through and saw it so wondering if it arose out of a problem of some sort.

src/routing.rs
```diff
@@ -1,23 +1,25 @@
...
-        .resource("/login", |r| r.method(Method::POST).with(sessions::create))
-        .resource("/logout", |r| {
-            r.method(Method::DELETE).with(sessions::destroy)
+pub fn routes(config: &Config, app: App<AppState>) -> App<AppState> {
+    let mw = AuthMiddleware::new(&config.token_secret);
+    let mw2 = AuthMiddleware::new(&config.token_secret);
+
+    app.resource("/artists/{id}", |r| {
+        r.middleware(mw);
+        r.method(Method::GET).with(artists::show);
```
Timo614: I guess it's more of a @CjS77 question but were we planning on making any API methods public? Or all behind authorization? Just confirming for the behavior we want here but otherwise looks good.

src/routing.rs
```diff
@@ -1,23 +1,25 @@
...
-        })
-        .resource("/artists/{id}", |r| {
-            r.method(Method::DELETE).with(artists::destroy)
-        })
-        .resource("/login", |r| r.method(Method::POST).with(sessions::create))
-        .resource("/logout", |r| {
-            r.method(Method::DELETE).with(sessions::destroy)
+pub fn routes(config: &Config, app: App<AppState>) -> App<AppState> {
+    let mw = AuthMiddleware::new(&config.token_secret);
+    let mw2 = AuthMiddleware::new(&config.token_secret);
```
Timo614: Are we doing the two here since it moves into the middleware call below? May be worth using [clone](https://doc.rust-lang.org/std/clone/trait.Clone.html) and just having one of these that we call clone on for each of the middleware we need it for.

src/bin/server.rs
mikethetike: It didn't seem to be reading the .env file when I was running it to test via Postman. Otherwise I think you have to set the environment variables manually

src/routing.rs
mikethetike: I think it'll be more clear once we have more methods. It's not set in stone yet

src/routing.rs
mikethetike: Correct, it is because they get moved into each resource. I'll look at cloning them.

Issue #10, opened 2018-07-09
A User should be able to reset their account password


CjS77:
This should follow standard 3 step process:
- Request reset link
- Reset link gets emailed to user's email address
- User uses link to reset password

(Future requirement): 2FA integration

PR #34, opened 2018-07-21
Password resets

Timo614:
Associated bn-db PR: https://github.com/big-neon/bn-db/pull/32 
Fixes: #10

This branch's primary focus is adding the ability to send emails from the API via SMTP (or to look into sent mails for test) and to create a forgot password flow.

Given this is an API I went with the following (let me know if you'd like me to adjust):

* `WHITELISTED_DOMAINS` env var is used with a comma separated list of approved domains (could be useful with CORS later too)
* A password request is made passing in email and a redirect URI
  * If the redirect URI is valid for the whitelist and email exists, a token that expires in 1 day is given and an email sent directing people to that redirect URI
  * If the redirect URI is invalid an error response is returned
  * If the email does not exist an okay response is still returned to avoid people using the API to scan for emails
* A consume password token request is made with the new password and the token, it checks the validity and if it is valid removes the token and associates the new password, it also logs in the user

Future / maybe this card if needed now:

* Push email handlng into a background thread so that the response doesn't have to wait

Some other modifications (refactoring out of need for the most part):

* Controller responses return a HttpResponse now with a proper status code and content type
* Added a 404 default route with some basic error json and content type
* New naming of some structs for artists / the existing setup wasn't scalable as we had the parameters in the models scope so only making the main models pub use. Let me know what you think. It's not ideal but not sure of a better call as we need `UserEditableAttributes` or something like that for each model to deserialize the restful update request. Using the `NewArtist` etc is fine for the new one though.
* Ended up adding an optional app name and domain. They are used for sending mail.
* Switched Rc to Arc for the test DB -- should allow it to work with integration tests hitting the API (I think?) while Rc would likely error out as it's not thread safe
* Test request provides a method `test_transport` which can be used to return the mail transport downcasted back to a TestTransport (so the sent Emails can be inspected), I used it in the forgot password tests to confirm emails were going out

The following curl commands work for testing (you'll need to approve the domain name in the whitelist env var and have a valid email you'll receive messages at). I confirmed it locally with an smtp server I have handy.

```
curl -H "Content-Type: application/json" -X POST -d '{"email":"tim@example.com", "reset_url": "http://www.bigneon.com/finish_reset"}' http://localhost:8088/password_reset -i
curl -H "Content-Type: application/json" -X PUT -d '{"password_reset_token":"d6c18e28-cd59-4a4c-81a1-51bec4e88fe1", "password": "examplePassword"}' http://localhost:8088/password_reset -i
```

Note: this references my fork for the bn-db with the latest commit in a branch over there for some needed changes to support this. Prior to merging it db should first be merged and then this one should be pointed back to master otherwise it may point to my fork in master.


Timo614:
There are a few cases where I should be logging the error prior to returning a friendlier error to the end user. I'll adjust those Monday when I rebase. We're planning on outputting errors directly to the log right? Or something more structured / in place in case we end up using sentry, etc in the future?

Timo614:
I've rebased this over the latest master and it's pointed to the latest commit on my rebased bn-db fork.

I'll quickly go through today and add any missing tests / wrap this branch up.

Timo614:
I fixed the fmt issues with travis (my machine had an older version of it). The failing tests are a result of how we have the db setup to run migrations from master. I could switch travis temporarily to use my own fork for installing the migrations if we think it's worth it (although it's not the most ideal process I think since it opens the door to mistakes during merging). Not sure of a better call there though.

Timo614:
For the time being I've updated the travis.yml file as well to point to my revision so the migrations work / the tests can pass.

https://github.com/big-neon/bn-api/pull/34/commits/89bbc24f9a3944ee2ac2cc2ed80b0652d8478cb7

This and the change to Cargo.toml will need to be reverted prior to merge.

PR #38, opened 2018-07-24
Point bn-db to master from PR

Timo614:
Just pointing back to master from the PR. Was set to point to my bn-db fork for the model changes there for the reset pw flow.

Should be safe to merge. Tests are currently broken but it's unrelated as we're in the middle of merging another feature (the db side is merged but api side pending so should clear up once we merge it).


Timo614:
We also fixed this in #37 so going to close this in favor of that PR.

Issue #40, opened 2018-07-25
Create CORS headers on actix


mikethetike:
These should be configurable.

PR #41, opened 2018-07-25
#40 Add CORS

mikethetike:
Added CORS processing so that the front end can use it.


mikethetike:
Added default CORS origin of http://localhost:3000

src/config.rs
```diff
@@ -90,6 +92,7 @@ impl Config {
                 .map(String::from),
         );
 
+        let allowed_origins = env::var(&ALLOWED_ORIGINS).unwrap_or("*".to_string());
```
Timo614: May be worth defaulting to nothing in the event it's not provided just out of precaution. If we drop cookie login altogether (which seems like we may do in the future) we may be fine since third parties wouldn't have access to the JWT but probably just safer to err on locking it down if not provided out of precaution either way.

src/config.rs
```diff
@@ -100,19 +103,20 @@ impl Config {
...
+            cookie_secret_key,
+            database_url,
+            domain,
+            environment,
+            mail_from_name,
+            mail_from_email,
+            mail_transport,
+            token_secret,
+            token_issuer,
+            whitelisted_domains,
```
Timo614: I wonder if whitelisted_domains is redundant now with allowed_origins as we could parse the domains out of that. It's used for determining if a redirect_uri is valid given the password reset flow. I guess one benefit may be that we'll have consumers who don't need CORS and the access_origins in that case may be confusing. I wonder if it's worth defaulting any CORS approved domain to be a whitelisted one too since I'd think at least in that case it's always true (anyone with CORS access should be an acceptable domain to redirect to).

Not a huge deal either way.

PR #42, opened 2018-07-25
Sw proposed unit tests

SWvheerden:
Please have a look and comment


tests/tests.rs
```diff
@@ -14,5 +14,6 @@ extern crate tokio_core;
 extern crate uuid;
 
 mod functional;
+mod functional_base;
```
mikethetike: I think these should be a sub-folder under `functional` rather. i.e. `/functional/base`

tests/functional_base/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod base_organizations;
```
mikethetike: This is already namespaced, so it can just be named `organizations`

PR #45, opened 2018-07-26
#44 Create get currently logged in user's details API call

mikethetike:


mikethetike:
I've also cleaned up the middleware in the routing

src/controllers/users.rs
```diff
@@ -0,0 +1,15 @@
...
+use auth::user::User as AuthUser;
+use bigneon_db::models::User;
+use server::AppState;
+
+pub fn current_user((state, user): (State<AppState>, AuthUser)) -> HttpResponse {
+    let connection = state.database.get_connection();
+    match User::find(&user.id, &*connection) {
+        Ok(u) => HttpResponse::Ok().json(&u.for_display()),
+        Err(_e) => {
+            HttpResponse::InternalServerError().json(json!({"error": "An error has occurred"}))
```
SWvheerden: rather use:
Err(e) => HttpResponse::from_error(ConvertToWebError::create_http_error(&e)),

mikethetike: Updated

PR #46, opened 2018-07-26
#17 ⁃ A system administrator should be able to assign a user as an organizations owner

mikethetike:


tests/functional/base/organizations.rs
```diff
@@ -155,3 +155,83 @@ fn update() {
     let updated_organization: Organization = serde_json::from_str(&body).unwrap();
     assert_eq!(updated_organization.name, new_name);
 }
+
+#[test]
+fn update_owner_id_fails_when_not_admin() {
```
SWvheerden: I would create the test on the new pattern. By adding a match to check for fail/succeed you can reuse the function, and test for all roles.

mikethetike:
@Timo614 I 100% agree. In fact, I think we should be doing PATCH updates instead of entire PUT replaces. I'm not 100% sure of the easiest way to do this in rust, but I think we should do this soon so that new developers don't make mistakes as you mentioned.

mikethetike:
I'm going to copy what you did for the Artist attributes

mikethetike:
Waiting for big-neon/bn-db/pull/39

src/controllers/organizations.rs
```diff
@@ -57,27 +57,44 @@ pub fn update(
...
-    let updated_organization: Organization = organization_parameters.into_inner();
-    match organization_response {
-        Ok(_organization) => {
-            let organization_update_response = updated_organization.update(&*connection);
+    match Organization::find(&parameters.id, &*connection) {
+        Ok(organization) => match organization
+            .update(organization_parameters.into_inner(), &*connection)
+        {
+            Ok(updated_organization) => HttpResponse::Ok().json(&updated_organization),
+            Err(_e) => HttpResponse::BadRequest().json(json!({"error": "An error has occurred"})),
```
SWvheerden: Hiding returned database error. Rather use the database to web error wrapper.

src/controllers/organizations.rs
```diff
@@ -57,27 +57,44 @@ pub fn update(
...
+        },
+        Err(_e) => HttpResponse::NotFound().json(json!({"error": "Organization not found"})),
```
SWvheerden: same here, true error suppressed

src/controllers/organizations.rs
```diff
@@ -57,27 +57,44 @@ pub fn update(
...
+    if !user.is_in_role(Roles::Admin) {
+        return application::unauthorized();
+    }
+    let connection = state.database.get_connection();
+    match Organization::find(&parameters.id, &*connection) {
+        Ok(organization) => match organization
+            .set_owner(json.into_inner().owner_user_id, &*connection)
+        {
+            Ok(updated_organization) => HttpResponse::Ok().json(&updated_organization),
+            Err(_e) => HttpResponse::BadRequest().json(json!({"error": "An error has occurred"})),
```
SWvheerden: suppressed error

src/controllers/organizations.rs
```diff
@@ -57,27 +57,44 @@ pub fn update(
...
+        },
         Err(_e) => HttpResponse::NotFound().json(json!({"error": "Organization not found"})),
```
SWvheerden: suppressed error

PR #54, opened 2018-07-27
Sw org owner can remove users from organization

SWvheerden:


SWvheerden:
This wont build until PR 38 iin bn-db is merged

PR #56, opened 2018-07-27
#48 ⁃ A Visitor should be able to create an account with their email address and a password

mikethetike:
Kept it very simple. Email verification and password strength to be done at a later stage


src/controllers/auth.rs
```diff
@@ -71,3 +74,18 @@ pub fn token(data: (State<AppState>, Json<LoginRequest>)) -> Result<AccessToken,
             .unwrap(),
     })
 }
+
+pub fn register((state, request): (State<AppState>, Json<RegisterRequest>)) -> HttpResponse {
```
SWvheerden: move to user controller

src/routing.rs
```diff
@@ -72,6 +72,9 @@ pub fn routes(app: &mut CorsBuilder<AppState>) -> App<AppState> {
             r.method(Method::GET).with(users::current_user)
         })
         .resource("/auth/token", |r| r.method(Method::POST).with(auth::token))
+        .resource("/auth/register", |r| {
```
SWvheerden: change to user

mikethetike:
Moved to `users` controller

PR #59, opened 2018-07-29
Sw find users via email

SWvheerden:


SWvheerden:
this branch requires the PR from bn-db as well

Issue #31, opened 2018-07-20
Create a API method to refresh the JWT token


mikethetike:

Timo614:
Grabbing this one.

PR #61, opened 2018-07-30
Refresh tokens

Timo614:
The DB branch includes some changes here which are required: https://github.com/big-neon/bn-db/pull/42
Fixes: #31

Note: Travis.yml and Cargo.toml are both currently pointing to my fork and the latest db revision in its refresh tokens branch.

This branch adds long lasting refresh tokens and severely limits the lifetime of access tokens provided. 

Primary modifications:
- Removed the roles cache from the AccessToken claims -- since this gates access (both to remove and add) seems worth always having the most up to date version. It's a hit to the DB but only when we're using the middleware and it seems worth it in this case and unavoidable.
- Added a token refresh endpoint to return a new access token when valid
- Refresh tokens do not expire (I can add an expiration of a year or some other long lasting period if we feel strongly for one) but do keep track of their issue date and if the password of a given user is changed they are invalidated across all sessions.
- Access tokens expire after 15 minutes

Example curl commands:

````
curl -H "Content-Type: application/json" -X POST -d '{"email":"test@domain.com", "password": "password"}' http://localhost:8088/auth/token -i
curl -H "Content-Type: application/json" -X POST -d '{"refresh_token": "eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiIyMzc4YWQ0ZS1jZGNjLTQyMmEtOWVhZi05MWU0YmJiYzc2ZWIiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiaXNzdWVkIjoxNTMyOTg0MDA4fQ.9G09KuGuSUhYfYb5ZmoLg/6r/eawpiT1tG/vspCBQjw"}' http://localhost:8088/auth/token/refresh -i
````

Let me know if you think we should modify any of this flow. I spent some time looking into JWT refresh tokens but there's no real standard for implementing it so open to suggestions.


Timo614:
Will rebase this and the bn-db branches today and pull in the latest changes.

Timo614:
Hmm this doesn't cover the (somewhat edge case) scenario:

1. User has their account credentials leaked so changes their password
2. Another person logged in as that user still has up to 15 minutes where they have access given the expiration of the access tokens where they can then change the password as well

I think it's probably fine though as a given user won't realize the password changed until they go to refresh and their given token is already expired (so at that point they can no longer modify the password anyhow) and if a user's credentials were leaked the attacker could at any time already change their password.

Since we can't really revoke access tokens perhaps the best solution to this edge case is simply to make the account edit flows force re-enter of customer authentication credentials (email, password) which would then prevent still active access tokens from adjusting passwords for users in that state.

PR #62, opened 2018-07-31
Update actix-web version to 0.7.2

### LINKED ISSUE #57: upgrade actix-web to latest version
upgrading actix-web from 0.6 to 0.7 breaks the auth class

Timo614:
This branch fixes #57 by upgrading the actix-web version to 0.7.2.

Two changes had to be made to do such:

1. Some modifications to the middleware trait signature to remove the request mutability
2. The router logic in actix-web has been modified and any access to it has been made crate only so I needed to change how we handle path extraction for tests.

Right now the test request method in actix-web still doesn't support extractors. In addition with the latest changes the interface exposes no way to pass a custom router to it for use in the extraction. 

I'd have just mocked the data by wrapping the parameters struct in the Path struct from actix-web but there's no way to create an instance of it outside of the extraction process.

As such instead a static parameter (the only way currently to pass in path parameters to the test request) is used to mock out an id parameter and then when the extractor has run I have the test logic override the id. It's not ideal by any means just not really sure of any better alternative. This is all changing fast given the API hasn't reached the 1.0.0 milestone so it's likely it'll continue to evolve.

May be worth us making an issue to switch over to the [integration test approach](https://github.com/actix/actix-website/blob/master/content/docs/testing.md#integration-tests). There may be some complications with handling the routing in a DRY way but aside from that I can't think of anything offhand that would make it too difficult to get in place of this. Didn't do it in this card as it's outside of the scope here and it's worth getting actix-web on the latest versions so we don't rely too heavily on the older API as we keep iterating on it. Easy enough to swap out the tests later if we think it's worth it.


Timo614:
Rebased over master to resolve conflicts.

PR #63, opened 2018-07-31
Check result of parsing auth token to avoid panic

### LINKED ISSUE #55: Panic when making an API call with an invalid token.
Happens when an API call is made with an invalid or missing token.

Validating on the frontend so it doesn't happen for the web but could affect other apps using the API at a later stage.

`api_1   | thread 'arbiter:"3393fd3b-c1f4-45c6-b6d2-235aa4d8e9d2":"actor"' panicked at 'called 'Result::unwrap()' on an `Err` value: Utf8(FromUtf8Error { bytes: [158, 233, 101], error: Utf8Error { valid_up_to: 0, error_len: Some(1) } })', libcore/result.rs:945:5
`

`api_1   | note: Run with `RUST_BACKTRACE=1` for a backtrace.`

Timo614:
Simple change to fix #55 

Previously we were unwrapping the result from the token parsing so an error there would trigger a panic that led to no API response. This branch just handles the error condition with a match statement. The behavior can be seen via curl in and out of the branch:

```
curl -X GET http://localhost:8088/organizations  -H "Authorization: Bearer invalid.token" -I
```


Timo614:
This is failing as the bn-db branch master is pointing to no longer exists (we merged it).

Once #62 is merged (I updated it to point to master) I'll quickly rebase this (should have no conflicts) and tests will all pass.

PR #76, opened 2018-08-03
API for creating a user from Javascript Facebook Login SDK

mikethetike:
This creates an API method at `/external/facebook/web_login` to create a user from the access token obtained from using the Javascript Facebook SDK Login function. 

There are other methods to support the Manual Facebook Login flow entirely from the server. This creates an access token to facebook that will last longer, but I'm struggling to get it working. Something might have changed with Facebook's side. I'd like to get the web section of it merged in so that it can be tested with the web front end, and the manual (server) flow can be completed later.


src/helpers/facebook_client.rs
```diff
@@ -0,0 +1,176 @@
...
+impl FacebookClient {
+    pub fn new(client_id: String, client_secret: String) -> FacebookClient {
+        FacebookClient {
+            client_id,
+            client_secret,
+        }
+    }
+
+    pub fn create_login_redirect_for(&self, redirect_uri: Url, scopes: Vec<&str>) -> String {
+        format!("https://www.facebook.com/v3.1/dialog/oauth?client_id={}&redirect_uri={}&response_type=code&scope={}",
```
SWvheerden: should we not move this to a constant file or env file?

src/helpers/facebook_client.rs
```diff
@@ -0,0 +1,176 @@
...
+                self.client_id, redirect_uri, scopes.join(","))
+    }
+
+    pub fn verify_auth_code(
+        &self,
+        code: &str,
+        original_redirect_uri: Url,
+    ) -> Result<FacebookAccessToken, FacebookError> {
+        let uri: Uri = format!(
+            "https://graph.facebook.com/v3.1/oauth/access_token?client_id={}&redirect_uri={}&client_secret={}&code={}",
```
SWvheerden: should we not move this to a constant file or env file?

src/helpers/facebook_client.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        let access_token = json_fut.wait();
+        access_token.map_err(|e| FacebookError { cause: Box::new(e) })
+    }
+
+    pub fn get_user_id(&self, access_token: &str) -> Result<String, FacebookError> {
+        let fb_client_token = self.get_app_api_token()?;
+
+        let uri: Uri = format!(
+            "https://graph.facebook.com/debug_token?input_token={}&access_token={}",
```
SWvheerden: constant

src/helpers/facebook_client.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        let debug_info = json_fut.wait();
+        debug_info
+            .map(|r| r.data.user_id)
+            .map_err(|e| FacebookError { cause: Box::new(e) })
+    }
+
+    fn get_app_api_token(&self) -> Result<String, FacebookError> {
+        let uri: Uri = format!(
+            "https://graph.facebook.com/oauth/access_token?client_id={}&client_secret={}&grant_type=client_credentials",
```
SWvheerden: constant

src/controllers/external/facebook.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    let existing_user = match existing_user {
+        Ok(e) => e,
+        Err(e) => return application::internal_server_error(e.description()),
+    };
+
+    let user = match existing_user {
+        Some(u) => {
+            info!("Found existing user with id: {}", &u.user_id);
+            User::find(&u.user_id, &*connection).unwrap()
```
Timo614: Looks good, we have the foreign constraints on the user_id column so can't think of anything unwrapping would suppress outside of significant db issues and those would be api wide. :+1:

src/controllers/external/facebook.rs
```diff
@@ -0,0 +1,128 @@
...
+    unimplemented!()
+}
+
+// TODO: Not covered by tests
+pub fn web_login(
+    (state, auth_token): (State<AppState>, Json<FacebookWebLoginToken>),
+) -> HttpResponse {
+    let connection = state.database.get_connection();
+    info!("Finding user");
+    let existing_user = ExternalLogin::find_user(&auth_token.user_id, "facebook.com", &*connection);
```
Timo614: Probably not worth thinking about for this card but wondering how often support related agents of a ticketing site would be asked to link fb accounts to the user's main accounts. I know some websites out there just combine the two acting as if they're just separate authentication processes for the same account while others keep them separate. Just imagining people reaching out in confusion on their mobile phones at an event when their facebook login doesn't show the same tickets they were expecting. 

Definitely the sort of thing that could come later too though if that did actually become a problem (and probably a bit preemptive for any account combination features now anyhow). :+1:

src/controllers/external/facebook.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    let user = match existing_user {
+        Some(u) => {
+            info!("Found existing user with id: {}", &u.user_id);
+            User::find(&u.user_id, &*connection).unwrap()
+        }
+        None => {
+            info!("User not found, creating user");
+            match User::create_from_external_login(
+                auth_token.user_id.clone(),
```
Timo614: Looking over the code this is just passed in as json posted data that's deserialized along with an access token to use with facebook? May be worth us asking facebook with the access token via `https://graph.facebook.com/me?access_token=` and not trusting any input. 

Worried someone may be able to trick the system for authorization of another account if we're getting this information from the client (wasn't sure how the flow works as you mentioned in the card that SSL related issue you ran into so couldn't run through it but just from a quick look at the code).

src/controllers/password_resets.rs
```diff
@@ -42,23 +42,26 @@ pub fn create(
...
-            Ok(user) => {
-                let result = mailers::user::password_reset_email(
-                    &state.config,
-                    &user,
-                    &parameters.reset_url,
-                ).deliver();
+    let user = match User::find_by_email(&parameters.email, &*connection) {
+        Ok(user) => match user {
+            Some(user) => user,
+            None => return HttpResponse::Created().json(request_pending_message),
```
Timo614: How are we planning on messaging users who purchase tickets with an external login? Just via the tooling itself? Wondering about confirmation emails, etc. and if it would have some form of email input for receipts in those cases.

Just thinking by not having an email on file we force support of the system into an odd place. Users may forget they have accounts that were with a third party and support will have no way to lookup users besides via their external ID which may be problematic as a form of identification for end users who may not be able to easily find it.

I think still having them return a generic message here is the right call as otherwise a script could scan the system for active emails pretty easily but wondering if not having an email on file is going to create a large support burden. I guess in that case they could always first ask if the user if they had a facebook account and to try that. If we had it on file we could then at least send a message informing the user they were a facebook login and to use that means of authorization at the time of the pw reset email. Not sure on the ideal behavior -- probably not worth doing ASAP just thinking it may be something to keep an eye out for when this system is put in use and as other forms of communication are added (email confirmation, internal reminders / communication, etc.).

Timo614:
Looks good to me -- just wondering about what direction we should take future communication with the third parties for the most part and some minor questions from the flow. :+1:

src/controllers/external/facebook.rs
mikethetike: The linking of accounts did come to mind. I agree that it's a different ticket though

src/controllers/external/facebook.rs
mikethetike: Good point. I'll add it

src/controllers/password_resets.rs
mikethetike: True. Let me look at obtaining the email from Facebook.

src/controllers/auth.rs
```diff
@@ -59,6 +61,11 @@ pub fn token((state, login_request): (State<AppState>, Json<LoginRequest>)) -> H
         Err(_e) => return application::unauthorized_with_message(login_failure_messaging),
     };
 
+    let user = match user {
+        Some(u) => u,
+        None => return application::unauthorized_with_message("Email or password incorrect"),
```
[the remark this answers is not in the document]
mikethetike: Regarding the Option<> for `find_by_email`, I personally prefer errors to be used only when there is an actual error, but I'm happy to go with the general feeling of the team. We would need to add a mapping on the bn-db side though, because currently it will return a `ErrorCode::QueryError` for no results.

PR #85, opened 2018-08-07
Return HTTP 409 Conflict when registering a user with an existing email address

mikethetike:
Relies on big-neon/bn-db/pull/49.

If the user's email exists, an HTTP 409 response will be generated. Note: this only checks if a DuplicateKeyError is received from the bn-db library. If, at another point, there is another unique key on the users table, then it will also generate the same error, and the API will not distinguish between duplicate emails and another unique field.


src/controllers/users.rs
```diff
@@ -59,6 +59,9 @@ pub fn register((state, request): (State<AppState>, Json<RegisterRequest>)) -> H
     ).commit(&*connection)
     {
         Ok(_u) => HttpResponse::Ok().finish(),
-        Err(e) => HttpResponse::from_error(ConvertToWebError::create_http_error(&e)),
+        Err(e) => match e.code {
+            3400 => HttpResponse::new(StatusCode::CONFLICT),
```
Timo614: I think you can use the `ConvertToWebError` with actix-web's Error `ErrorConflict` if we want to encapsulate it in the `create_http_error` logic.

src/errors/database_error.rs
```diff
@@ -16,6 +16,7 @@ impl ConvertToWebError for DatabaseError {
             3100 => error::ErrorInternalServerError("Could not insert record"),
             3200 => error::ErrorInternalServerError("Could not update record"),
             3300 => error::ErrorInternalServerError("Could not delete record"),
+            3400 => error::ErrorInternalServerError("Duplicate record exists"),
```
Timo614: Ah actually I see why you handled it above now.

So we're returning an internal server error by default but using conflict in the event that it happens for the user record with registration. Wondering if it's worth just defaulting to the `ErrorConflict` in this case.

mikethetike: The main reason was that it's easier to check for a status code 409 in the frontend than it is to get a 500 and then check another field.

Timo614: https://github.com/actix/actix-web/blob/master/src/error.rs#L809 it returns status code 409 when you use ErrorConflict.

mikethetike: Ahhh. I just assumed they were all 500s

PR #86, opened 2018-08-07
Denormalize roles into scopes

mikethetike:
By using scopes instead of the roles, we can have a finer level at which we grant or deny permissions. 

I used the term `scopes` instead of `permissions` to match the commonly used Oauth claim. At a future stage, we can use these scopes to create API tokens.


mikethetike:
> It uses roles when assigning,
I'm not sure where you are referring to?

The idea is that roles are stored in the database, but only scopes are checked in the api.

mikethetike:
> What I feel we must only hard map the roles when assigning a role to the user as a certain set of default scopes. But we must store scopes and not roles in the database.

This would make an edit roles screen in the frontend very difficult.

SWvheerden:
I agree that It will complicate things in the front end a bit. But I believe the compromise will be worth it in the long run for a more fine grain control. 

They way I see we could do is is almost almost like how most image sites add tags to images. We can perhaps even group them together into a role if the user has all the scopes from a certain role and just use colour to differentiate  between the two.

PR #92, opened 2018-08-08
Find all organizations in existance if you are an admin

SWvheerden:
This uses the same api call request as find all organizations. 
There is a clause added that that if you call with admin rights it will search for all, not just where you are an admin or member


src/controllers/organizations.rs
```diff
@@ -13,11 +13,26 @@ pub struct PathParameters {
 }
 
 pub fn index((state, user): (State<AppState>, User)) -> HttpResponse {
+    if user.is_in_role(Roles::Admin) {
+        return index_for_orgs((state, user));
```
Timo614: Are we ever going to have admins who are also organization members? If not this is probably fine otherwise we may want to introduce another route for the full listing for admins versus the list of the ones they're part of. Otherwise the frontend will have no way to figure out which organizations the admin has joined as they will all appear to be part of that list.

We may want to also allow the show action for admins as it looks like that's locked down as well.

src/controllers/organizations.rs
```diff
@@ -13,11 +13,26 @@ pub struct PathParameters {
...
     let connection = state.database.get_connection();
-    let organization_response = Organization::all(user.id(), &*connection);
+    let organization_response = Organization::all_linked_to_user(user.id(), &*connection);
+    match organization_response {
+        Ok(organizations) => HttpResponse::Ok().json(&organizations),
+        Err(e) => HttpResponse::from_error(ConvertToWebError::create_http_error(&e)),
+    }
+}
+
+pub fn index_for_orgs((state, user): (State<AppState>, User)) -> HttpResponse {
```
mikethetike: This method should have a better name. It's not clear from the name what makes this method different from `index`.

tests/functional/organizations.rs
```diff
@@ -6,23 +6,28 @@ mod index_tests {
     use super::*;
     #[test]
     fn index_org_member() {
-        organizations::index(Roles::OrgMember);
+        organizations::index(Roles::OrgMember, true);
+        organizations::index_for_admin(Roles::OrgMember, true);
```
mikethetike: I think it will be better to create a new `mod index_for_admin` for the `index_for_admin` tests to match the way the other methods have been tested

src/controllers/organizations.rs
SWvheerden: I am also of the opinion that for now we can filter on the front end. But currently the roles are used as scopes. So if an admin is also an org owner, the admin scope will override the OrgOwner scope. 

But that being said we should keep in mind that this will be an issue we should keep en eye out, perhaps we could move this to a different api call or and adding limitations so we can filter on the backend and not the front end. I am going to move this discussion to an issue, we so can continue it there, as this needs more discussion and input. But philipr-za requires this PR in so he can work on this. I dont want to loose this issue.

PR #94, opened 2018-08-09
Check if org venue link exists prior to creation

### LINKED ISSUE #91: When linking venues to organizations validate it's not already linked.
Path: `/venues/${venueId}/organizations`
I'll be doing client side validation but should be validated in the backend as well.

Timo614:
Fixes #91
Related / required: https://github.com/big-neon/bn-db/pull/52

This branch simply prevents additional links from being made via the use of a unique index on the association and a method that checks to see if it exists to give a conflict status code if it already is present.

Currently pointing travis and the cargo.toml files to my revision in my fork. I also had to make some modifications to the code to handle some code that has been merged in the bn-db side but still is pending on the bn-api so I'll remove that extra logic when I rebase later on (was just to get the tests passing).

I adjusted the tests for this method to use the new pattern we're using with the base tests. I left the others alone as it's out of scope here.


Timo614:
Going to quickly rebase these branches.

.travis.yml
```diff
@@ -30,7 +30,7 @@ env:
 before_script:
   - export PATH="$PATH:$HOME/.cargo/bin"
   - rustup component add rustfmt-preview
-  - cargo install --git https://github.com/big-neon/bn-db.git --force
+  - cargo install --git https://github.com/timo614/bn-db.git --rev 2b06ae7b28953b6e5b890bc22e2ffab270a4d8b9 --force
```
SWvheerden: Timo614, can you please point this to the https://github.com/big-neon/bn-db.git so we can merge this in.

Timo614: Sure I'll rebase it again and leave it pointed to master (tests will break in the meantime but should be fine once we merge it in).

Timo614:
I'll rebase this again

Issue #58, opened 2018-07-27
Organizations/events endpoints too open


Timo614:
We're forcing a OrgMember role to access the endpoints but we're not filtering the data down to those that the given user has access to.

https://github.com/big-neon/bn-api/blob/master/src/controllers/organizations.rs#L30

Just need something in the logic first to confirm the user can access the resource to edit as they may not be the org owner or member of that given org.

The events one has this issue via the organization_id in the new event and update events:
https://github.com/big-neon/bn-api/blob/master/src/controllers/events.rs#L69
https://github.com/big-neon/bn-db/blob/master/src/models/events.rs#L30

Could allow another org owner who doesn't belong to the organization to assign events to it. May be others just quickly saw those when looking through.

mikethetike:
Let's investigate Venues as well

mikethetike:
It might be an idea to return 401 when the user does not have the correct role/scope to perform the action, and 403 when they do have the role but don't have access to that entity (e.g. someone else's organization)

Timo614:
Grabbing.

Sounds good I'll use 403 (Forbidden) if the user is authenticated but lacks access and I'll adjust the controller actions to all properly confirm their access.

Timo614:
@zmelamed a few quick questions:

- What level of access should be able to create an organization private artist? We currently allow org members to create events, should org members be able to create private artists? Or just the org owner?
- If you have the OrgOwner role should you be able to lookup any user in the system? That's how it is now but wondering if we should lock it down to just members of the given user's organization. I guess it depends on how those endpoints are used and locking it down would prevent an organization from being able to search up a given user to add to their org. I'm just concerned about the privacy implications of giving the org owners full access to random user data via the lookups.
- Do we plan on supporting multiple OrgOwners in the future? The current scope setup doesn't work really well with that so we'd have to likely have some sort of role associated with the organization_users table so we can confirm that the user has the role for the given organization (otherwise if they were granted OrgOwner status then they'd have it across all of their organizations given the scope setup we've been using).

zmelamed:
- Any member that can create an event can create the artist for an event, not just the OrgOwner

- Orgs should only be able to view users and users activity related to that org and that orgs events. In regards to adding an OrgMember to an org: An org should be able to add a user as a member of their org by email invite, not by selecting them from a public global pool of users.

- a single org owner, yes, but there will be the need for multiple org admins https://github.com/big-neon/docs/blob/master/overview.md

Timo614:
[...]

PR #96, opened 2018-08-09
Lock down event update parameters

Timo614:
Fixes: #87 

We're forcing all event parameters at the moment with the update flow which also exposes the created_at field to being modified. This branch uses the pattern we've been using to update where we pass in a subset of the editable fields as optional parameters (organizations uses it as well).

I tested updating an event locally which was successful for the partial updates:

```
curl -H "Content-Type: application/json" -X POST -d '{"email":"test@domain.com", "password": "password"}' http://localhost:8088/auth/token
curl -X PUT -d '{"name": "test2", "ticket_sell_date": "2013-07-08T09:10:11"}' http://localhost:8088/events/ddd1bd6b-27db-4a7c-bd80-8ab4aca76be7 -H "Content-Type: application/json" -H "Authorization: Bearer token_here"
```

Note: not handling https://github.com/big-neon/bn-api/issues/58 in this issue (I may deal with that other issue in a subsequent patch but don't want to expand the scope beyond the initial issue as it's a blocker for Jason and the other issue is more of a security concern / won't block him in the meantime).

Also note: this is currently pointing to my branch for travis and the cargo file. I'll revert that after the review approval process.

In addition to that I have made some modifications to address some broken tests in master for a branch pending on the api side that had its db side merged. Once it's merged I'll rebase so we remove that just wanted to let the tests pass.


mikethetike:
I've merged this in because there are a lot of PR's pending and they will each have a different cargo.toml, so they'll converge on the correct version.

Issue #80, opened 2018-08-07
Split `name` into `first_name` and `last_name` in users table


Jasonvdb:

Timo614:
Going to grab this

Jasonvdb:
Thanks updated the frontend in https://github.com/big-neon/bn-web/pull/43

PR #97, opened 2018-08-09
Separate user name into first and last names

Timo614:
Fixes: #80

This branch fixes issue 80 by splitting the name field into two separate fields. 

You can see the behavior via the registration logic:

```
curl -X POST -d '{"email": "test@tari.com", "first_name": "John", "last_name": "Michael", "phone": "1-421-444-4443", "password": "secret"}' http://localhost:8088/users/register -H "Content-Type: application/json"
```

Note: this branch currently is pointing to my bn-db fork on a revision so that travis tests pass. It should be reverted after approval before we merge.

The branch has some changes in place to let the tests pass that should be removed when I rebase later once the bn-api changes pending for it are merged in. Just the nature of the moving target for the bn-db as that code was merged but the bn-api is still pending so just a stopgap so I could build the repo and get the tests the pass until that's in.

https://github.com/big-neon/bn-db/pull/54


Timo614:
One thing I was thinking about while modifying the records here for the user we use in the tests... Would it make sense for us to move the builders into some sort of dev namespace that we can reference both here in our tests and in the tests for the bn-db?

We probably don't want to duplicate the builder logic here so doing that at least would allow us to use it in both. The actix-web folks have something similar setup: https://github.com/actix/actix-web/blob/2e8d67e2aecfb850c341146bce5ccf60ff04f73b/src/lib.rs#L239

If we think that's a good direction I'll switch it in my bn-db branch to do something similar. I'm inclined to go that direction so this is still a WIP at the moment until I finish that sometime tomorrow.

mikethetike:
> We probably don't want to duplicate the builder logic here so doing that at least would allow us to use it in both. The actix-web folks have something similar setup: https://github.com/actix/actix-web/blob/2e8d67e2aecfb850c341146bce5ccf60ff04f73b/src/lib.rs#L239

Sounds good to me

SWvheerden:
Should we keep of merging this branch till we did this or just merge and refactor later?

mikethetike:
Tried to resolve the conflicts

mikethetike:
> Tried to resolve the conflicts

Ok well that didn't work....

Timo614:
I'll quickly rebase this and fix any of the conflicts.

Timo614:
Should be set once the tests run through travis :+1:

Issue #75, opened 2018-08-03
When admin adds a new venue/organization it's not saving all details


Jasonvdb:
Passing below params to `/venues` but it only seems to be saving the `name` field

`
{
"name":"Jazz Hands Bar",
"organization_id":"14484360-f049-44d9-bca1-9824f70eb088",
"phone":"0897373833",
"address":"10 2nd Avenue, New York, NY, USA",
"city":"New York",
"state":"New York",
"country":"United States",
"zip":"10003",
"place_id":"ChIJgdinUYRZwokR06E7yqMzm5Y"
}
`

Not sure if I'm maybe calling it incorrectly.

Jasonvdb:
I'm creating the entry and then updating it right away from the frontend with the rest of the details. This works but it would be better for it to be able to save all the details in one API call in the future.

Jasonvdb:
I'm creating the entry and then updating it right away from the frontend with the rest of the details. This works but it would be better for it to be able to save all the details in one API call in the future.

Jasonvdb:
Same happening when creating an hitting `/organizations`

philipr-za:
This was fixed in https://github.com/big-neon/bn-db/pull/55

PR #102, opened 2018-08-10
Fix tests for new venue and org structs that have been updated in bn-db

philipr-za:
The fix for issue https://github.com/big-neon/bn-api/issues/75 in bn-db broke these two tests due to adding new fields to the structs. Added those fields to the tests.

The bn-api tests will fail until https://github.com/big-neon/bn-db/pull/55 is merged in.


Timo614:
Ended up making the same modification when I rebased #97 to get the tests passing via https://github.com/big-neon/bn-api/pull/97/commits/87bb3b41e6f067241e4cf3906f5f56a7f67e869e so should be set :+1:

philipr-za:
This issue was resolve in https://github.com/big-neon/bn-api/commit/87bb3b41e6f067241e4cf3906f5f56a7f67e869e

PR #106, opened 2018-08-14
Fix organization invites

SWvheerden:
Added in refactoring of a word from bn-db side
Fixed routing issues with the api calls not correctly mapping to functions
Fixed users  function returning Ok(None) on user not found and not error like previously.


Cargo.toml
```diff
@@ -13,9 +13,9 @@ actix-web = "0.7"
 diesel = {version = "1.3",features = ["r2d2"]}
 dotenv = "0.13"
 # use this one for testing with master branch
-bigneon_db = { git="https://github.com/big-neon/bn-db.git"}
+# bigneon_db = { git="https://github.com/big-neon/bn-db.git"}
 # use this one when testing locally.
-# bigneon_db = { path = "../bn-db" }
+bigneon_db = { path = "../bn-db" }
```
mikethetike: Wrong path

src/controllers/organization_invites.rs
```diff
@@ -18,24 +18,45 @@ pub struct Info {
     pub user_id: Uuid,
 }
 
-pub fn create(data: (State<AppState>, Json<NewOrganizationInvite>, AuthUser)) -> HttpResponse {
+#[derive(Deserialize)]
+pub struct NewOrgInviteCall {
```
mikethetike: Rename to `NewOrgInviteRequest`

src/controllers/organization_invites.rs
```diff
@@ -18,24 +18,45 @@ pub struct Info {
...
+        organization_id: invite_args.organization_id,
+        inviter_id: user.user.id,
+        user_email: invite_args.user_email,
+        security_token: None,
+        user_id: invite_args.user_id,
+    };
     let email = actual_new_invite.user_email.clone();
-    let org_invite = match NewOrganizationInvite::commit(&mut actual_new_invite, &*connection) {
+
+    let mut org_invite = match NewOrganizationInvite::commit(&mut actual_new_invite, &*connection) {
```
mikethetike: Do `actual_ new_invite` and `org_invite` both need to be `mut`?

Issue #104, opened 2018-08-13
Route `/events` shouldn't require a token to call


Jasonvdb:
Users need to be able search and view all events without creating an account yet.

Timo614:
I'll quickly grab this one.

Timo614:
One difficulty here is the way the middleware is being handled. We're associating it with a given resource but the only way to define two controller actions using the same path is to keep them in the same resource as well. 

What we have now, for example:

```
        .resource("/events", |r| {
            r.middleware(AuthMiddleware::new());
            r.method(Method::POST).with(events::create);
            r.method(Method::GET).with(events::index);
        })
```

The following won't work though as the index route is then ignored and it returns 404:

```
        .resource("/events", |r| {
            r.middleware(AuthMiddleware::new());
            r.method(Method::POST).with(events::create);
        })
        .resource("/events", |r| {
            r.method(Method::GET).with(events::index);
        })
```

I'm checking the docs if there's a way I can get around this otherwise may have to figure out a better way to associate the middleware with the given routes (not sure offhand how though but will look into it).

Timo614:
Should the venues index and show actions require authentication (right now they do)?

Also in the associated PR I set the following routes so they didn't require it (venues has similar routes as well) but let me know if these should require authorization:

```
        .resource("/events/venues/{id}", |r| {
            r.method(Method::GET).with(events::show_from_venues);
        })
        .resource("/events/organizations/{id}", |r| {
            r.method(Method::GET).with(events::show_from_organizations);
        })
```

PR #111, opened 2018-08-14
Modify auth for public actions in same scope as authed

Timo614:
Fixes: #104

This branch removes the auth middleware logic so that it can be applied on an individual controller action basis. Previously we assigned it on a per resource basis but that doesn't work for the RESTful nature of the routing as the index and create routes for events use the same scope `/events`. 

To accomplish this the following was done:

- Auth middleware was removed and logic for loading the auth user is entirely in the extractor
- The extractor for the User logic returns a result with an unauthorized error for any of the same conditions previously triggered via the auth middleware
- Removed the `events::read` scope as it's always public re: #104
- Modified the routes for the organization events and venue events to be more restful in terms of the scoping (`organization/1/events` now in place of `events/organization/1` for example)

Originally I had this PR continue to set the user id via the middleware which was then extracted (as you can see from the commit history) but then opted to move it all into the extractor as it's a bit cleaner and it removes the need for the logic to run at all for controller actions not requiring auth (versus having to parse the token in the event it was provided for an action not requiring it).


Timo614:
Thinking about:

> Removing the errors for an invalid parsed token (in favor of just returning a generic invalid auth token response if the user has not been set by the middleware)

As mentioned above I had to move the middleware to the app level and am using the extractor to handle the 401 case now. This was done to allow routes matching the same path to have different authentication requirements (which wouldn't be possible with the older approach).

Right now the middleware code still handles parsing and returning an error if the token is invalid or expired. This means that if a method not requiring the middleware passes in the auth token it will still attempt to parse it and return an error if it's invalid. I could remove those errors and just treat the missing id from the request extensions (normally set by a valid token) as a 401 condition during extraction but we'd lose out on some additional debugging info.

With the change to move the user load into the extractor the middleware's only purpose is to handle parsing the auth token when provided and setting the value. If we don't pass in the auth token then it won't do any additional overhead when handling those controller actions besides checking for the presence of the header.

I'm thinking perhaps the best call here is to nuke the middleware completely and just have the extractor handle parsing out the auth token. That way we could still have the relevant errors returned, ignore auth tokens when they are not required, and remove the need to do the user load as the struct could require the db user record again (as opposed to just setting the id and then loading it after extraction). Going to quickly see how feasible of a change that is.

Timo614:
In the end opted for what I described in the last comment. I removed the auth middleware in favor of having the extractor handle it. Can see the behavior in the branch via:

Venues (still requires auth at this time, may change):

Success: 

```
curl -X GET http://localhost:8088/venues -H "Content-Type: application/json" -H "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiJkMTlhNTYxNy0yOTA4LTQxZjktOTIwZC1hZWNkM2IyYWUwOGIiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTM0MzQ3MDg0fQ.AOxph6YJvl8irn0zaRusLqzDEu+4ZaWYDrUCJ/zVmOE"  -H "Accept: application/json" -I

HTTP/1.1 200 OK
content-length: 2
access-control-allow-origin: *
vary: Origin
date: Wed, 15 Aug 2018 15:26:25 GMT
```

Expired token:

```
curl -X GET http://localhost:8088/venues -H "Content-Type: application/json" -H "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiJkMTlhNTYxNy0yOTA4LTQxZjktOTIwZC1hZWNkM2IyYWUwOGIiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTM0MzQ3MDg0fQ.AOxph6YJvl8irn0zaRusLqzDEu+4ZaWYDrUCJ/zVmOE"  -H "Accept: application/json" -I
HTTP/1.1 401 Unauthorized
content-length: 0
access-control-allow-origin: *
vary: Origin
date: Wed, 15 Aug 2018 15:33:07 GMT
```

Missing token:

```
curl -X GET http://localhost:8088/venues -H "Content-Type: application/json"  -H "Accept: application/json" -I
HTTP/1.1 401 Unauthorized
content-length: 0
access-control-allow-origin: *
vary: Origin
date: Wed, 15 Aug 2018 15:32:54 GMT
```

Invalid token:
```
curl -X GET http://localhost:8088/venues -H "Content-Type: application/json" -H "Authorization: Bearer invalid"  -H "Accept: application/json" -I
HTTP/1.1 401 Unauthorized
content-length: 0
access-control-allow-origin: *
vary: Origin
date: Wed, 15 Aug 2018 15:32:17 GMT
```

Events (ignores auth header):

```
curl -X GET http://localhost:8088/events -H "Content-Type: application/json" -H "Authorization: Bearer invalid"  -H "Accept: application/json" -I
HTTP/1.1 200 OK
content-length: 2
access-control-allow-origin: *
vary: Origin
date: Wed, 15 Aug 2018 15:33:54 GMT
```

Events, no header provided:

```
curl -X GET http://localhost:8088/events -H "Content-Type: application/json"  -H "Accept: application/json" -I
HTTP/1.1 200 OK
content-length: 2
access-control-allow-origin: *
vary: Origin
date: Wed, 15 Aug 2018 15:34:25 GMT
```

Post request to events (invalid data just using to show the auth logic):

```
curl -X POST -d '{}' http://localhost:8088/events -H "Content-Type: application/json"  -H "Accept: application/json" -v
...
< HTTP/1.1 401 Unauthorized
< content-length: 0
< access-control-allow-origin: *
< vary: Origin
< date: Wed, 15 Aug 2018 15:35:52 GMT
```

Posting with valid token (invalid data just using to show the auth logic): 

```
curl -X POST -d '{}' http://localhost:8088/events -H "Content-Type: application/json" -H "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiJkMTlhNTYxNy0yOTA4LTQxZjktOTIwZC1hZWNkM2IyYWUwOGIiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTM0MzQ4MzAzfQ.kh7dZr2m1q+qAT+Wcp+Pflu/W5TnzqBWtxiYDpUKbY8"  -H "Accept: application/json" -v
...
< HTTP/1.1 400 Bad Request
< content-length: 0
< access-control-allow-origin: *
< vary: Origin
< date: Wed, 15 Aug 2018 15:37:16 GMT
```

Timo614:
The code here should be set just waiting on some feedback from questions asked in #104 re the route behavior for the venues routes as well and if they should be made public (and confirming the events via their association routes should be open -- something I did in this branch but could revert once I hear back).

Issue #114, opened 2018-08-16
A user should be able to easily indicate that they are interested in an event


philipr-za:
Milestone 2 feature

philipr-za:
completed in PR https://github.com/big-neon/bn-api/pull/115

PR #115, opened 2018-08-16
A user should be able to easily indicate that they are interested in an event

philipr-za:
Milestone 2 feature. Issue: https://github.com/big-neon/bn-api/issues/114

Added the event_like model and endpoints to:
1. Allow a logged in user to like an event
2. Allow any user to check how many total likes an event has
3. Allow a logged in user to unlike an event.


api/src/auth/user.rs
```diff
@@ -125,12 +129,16 @@ fn get_scopes(roles: Vec<String>) -> Vec<String> {
 
 fn get_scopes_for_role(role: &str) -> Vec<Scopes> {
     match role {
-        "Guest" => vec![Scopes::ArtistRead, Scopes::VenueRead],
+        "Guest" => vec![Scopes::ArtistRead, Scopes::VenueRead, Scopes::EventRead],
```
mikethetike: I think `Scope::EventRead` has been removed because unauth'ed guests can view events

api/src/auth/user.rs
```diff
@@ -150,16 +158,21 @@ fn get_scopes_for_role(role: &str) -> Vec<Scopes> {
...
+        res
+    );
     let res = get_scopes_for_role("OrgOwner");
     assert_eq!(
         vec![
             Scopes::OrgWrite,
             Scopes::UserRead,
             Scopes::EventWrite,
             Scopes::OrgRead,
+            Scopes::EventLikeWrite,
```
mikethetike: I think let's rename this to `Scopes::EventLike`

api/src/auth/user.rs
```diff
@@ -174,13 +187,15 @@ fn scopes_to_string() {
 #[test]
 fn get_scopes_test() {
     let res = get_scopes(vec!["Guest".to_string()]);
-    assert_eq!(vec!["artist:read", "venue:read"], res);
+    assert_eq!(vec!["artist:read", "event:read", "venue:read"], res);
```
mikethetike: See above

api/src/controllers/events.rs
```diff
@@ -109,3 +114,48 @@ pub fn update(
...
+    let connection = state.database.get_connection();
+    let event_like_response = EventLike::create(parameters.id, user.id()).commit(&*connection);
+    match event_like_response {
+        Ok(event_like) => HttpResponse::Created().json(&event_like),
+        Err(e) => HttpResponse::from_error(ConvertToWebError::create_http_error(&e)),
+    }
+}
+
+pub fn total_likes(
+    (state, parameters, user): (State<AppState>, Path<PathParameters>, User),
```
mikethetike: I think that the summary of likes (just the total in this case) should be returned with `/events` list and `/events/{id}`. Then this method can return a paged list of users who have liked this event. (I think this can be done later)

api/src/auth/user.rs
```diff
@@ -28,7 +30,9 @@ impl fmt::Display for Scopes {
         let s = match self {
             Scopes::ArtistRead => "artist:read",
             Scopes::ArtistWrite => "artist:write",
+            Scopes::EventRead => "event:read",
             Scopes::EventWrite => "event:write",
+            Scopes::EventLikeWrite => "event_like:write",
```
mikethetike: rename to `"event:like"`

api/src/routing.rs
```diff
@@ -67,6 +67,12 @@ pub fn routes(app: &mut CorsBuilder<AppState>) -> App<AppState> {
             r.method(Method::GET).with(events::show);
             r.method(Method::PUT).with(events::update);
         })
+        .resource("/events/{id}/interest", |r| {
+            r.method(Method::POST).with(events::add_interest);
+        })
+        .resource("/events/{id}/disinterest", |r| {
```
mikethetike: I think we can use `/events/{id}/interest` for this one as well, to keep the paths simple. The DELETE verb already shows that it is removing the interest

PR #120, opened 2018-08-17
Add ticket allocations

mikethetike:
Added the beginnings for creating an allocation of tickets for an event. I've created a table in postgres to store the ticket allocations and an initial client that will talk to the Tari network.

I imagine that the Tari client will move into it's own repo at some point. 

Currently the data is stored in the database and then immediately sent to Tari, but later I think that we will need to have an async processor that tries to send it to the network with retries.


db/src/models/ticket_allocations.rs
```diff
@@ -0,0 +1,78 @@
...
+#[belongs_to(Event, foreign_key = "event_id")]
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+#[table_name = "ticket_allocations"]
+pub struct TicketAllocation {
+    pub id: Uuid,
+    pub event_id: Uuid,
+    tari_asset_id: Option<String>,
+    pub created_on: NaiveDateTime,
+    synced_on: Option<NaiveDateTime>,
+    ticket_delta: i64,
```
SWvheerden: table is i8? struct is i64?

mikethetike: Int8 is correct. Int4 in diesel is actually a standard i32

Issue #64, opened 2018-08-01
Add OrgOwner role when a user is set as the owner of new organization


Jasonvdb:
When creating an organization on the frontend and assigning an user as the owner, that user user needs to have the OrgOwner role added.

Timo614:
I wonder if it's worth just doing a SQL query for some of these roles like org owner and org member at the time we're fetching them from the DB as they are dynamic depending upon the state of the other tables. For admin I think it makes sense to store that but I'm wondering if we're just going to create a headache for ourselves by trying to keep this cache (since it's determined based on their state in their other tables) always up to date.

Otherwise when transferring ownership we have to be careful to always keep the cache across users up to date.

We already have to hit the db to determine the roles as otherwise the access token may not have the most up to date information so it's just another query at that time and it's looking up by the user_id so it should be pretty quick given how few organizations the system will have and the index.

It's not the end of the world if we want to cache it but I wonder if by doing so we're just headaches for ourselves by prematurely optimizing this. The alternative is when we update an organization to have a new owner we need to make sure to check if the existing owner had any other organizations (so they should keep the role) or if we should remove it from them and if the new owner already has the role or if we should add it for them. I can't imagine we'll have too many paths to set an organization owner. Organization member may require a few locations though as setting the org owner likely should give org member status and remove it from the other and so should the path to add a member to an organization.

I'm just wondering if we're prematurely optimizing a bit by trying to cache these dynamic roles. We could do something with a view that determines the roles in addition and appends them that should be added though if we want middle ground with a cache still. Still may be a bit of premature optimization so not sure if that's worth it.

Not sure, what do you all think?

Jasonvdb:
At the moment in bn-web the roles are not being cached at all. They were when they were stored in the token.

Right now I'm loading in user details & roles from `/users/me` every time the web app gets loaded and then again every 5min if the tab is still open. So if a user's roles gets updated, they could just refresh the page to get the latest roles right away.

Guess this will change if we're going ahead and doing the scopes thing but I needed a way to use roles in the meantime so that's why @mikethetike added it to the api for me.

Timo614:
I meant we're technically caching the OrgOwner and OrgMember by storing them in the users table role field since they are dynamically determined. Was thinking one option could be to use a db view to append any dynamic role to the user's existing static roles list (like Admin which isn't determined based on any other table data). 

[...]

PR #127, opened 2018-08-21
Add OrgOwner roles to users when they are made the owner of a newly created Organization

philipr-za:
Addresses issue https://github.com/big-neon/bn-api/issues/64

- Check that a role is not a duplicate when adding it to a user
- Add the OrgOwner role when an Org is created with that user as the owner and when the org is updated.


api/src/controllers/organizations.rs
```diff
@@ -107,7 +115,12 @@ pub fn update_owner(
     match Organization::find(parameters.id, &*connection) {
         Ok(organization) => {
             match organization.set_owner(json.into_inner().owner_user_id, &*connection) {
-                Ok(updated_organization) => HttpResponse::Ok().json(&updated_organization),
+                Ok(updated_organization) => {
+                    let _ = DbUser::find(&updated_organization.owner_user_id, &*connection)
+                        .unwrap()
+                        .add_role(Roles::OrgOwner, &*connection);
```
Timo614: We should probably check to see if the old organization owner is still the owner of any other organizations after the update here and then remove the OrgOwner role if not.

May be worth handling the result from the `DbUser#find` method above -- although there should never be a case where the uuid doesn't exist given the foreign key constraints on the db so probably not super critical and none of the other errors it can return seem that likely given its just a find (https://docs.diesel.rs/diesel/result/enum.Error.html https://docs.diesel.rs/diesel/result/enum.DatabaseErrorKind.html).

philipr-za:
So what I ended up doing was moving the OrgOwner logic into the DB layer. 
- When a Org is created the user who is the owner will get the OrgOwner role.
- Added a remove_role method on User
- The set_owner method will now give the new owner the OrgOwner role and check if the old owner owns any other Orgs and if not remove their OrgOwner role.
- Updated the tests for User and Organization
- Added the change_owner test

db/src/models/users.rs
```diff
@@ -114,6 +116,20 @@ impl User {
...
 
+    pub fn remove_role(&self, r: Roles, conn: &Connectable) -> Result<User, DatabaseError> {
+        let mut current_roles = self.role.clone();
+
+        current_roles.retain(|x| x.as_str() != &r.to_string());
+
+        DatabaseError::wrap(
+            ErrorCode::UpdateError,
+            "Could not remove role from user",
+            diesel::update(self)
```
mikethetike: This `diesel::update` should be refactored into a common function to reduce duplication between `add_role` and `remove_role`

philipr-za:
Thanks for the input, the Rustacean way is becoming clearer every day. Addressed all the comments.

Issue #122, opened 2018-08-20
Get Email from Facebook Login


mikethetike:
Currently we accept the facebook login token via the web, but we do not receive the email or the name and last name. We should retrieve these values from Facebook by making a call on the `bn-api` server and then populating the fields in the database.

Timo614:
I'll grab this one.

mikethetike:
@Timo614 I have a branch `mikethetike-fb-refactor` that fixes some of the server calls to FB, and replaces the `hyper` low level http client with `reqwest`. Feel free to use it or not as you see fit. Hopefully it helps, but I'm not too fussed if it isn't used. 
I didn't continue it because of the fact that we are using the web login for facebook, so server code wouldn't be used. Let me know what you think though

Timo614:
Sounds good, thanks!

PR #129, opened 2018-08-21
WIP: DRY up DB error handling with question mark operator

Timo614:
We tend to have a lot of boilerplate code for handling the DB errors as we need to match on the results and then return a HttpResponse converted from the database error itself.

This branch cleans up some of that boilerplate by allowing us instead to return the error which is the converted into the HttpResponse.

As we keep the `DatabaseError` bn-db logic in a separate crate I unfortunately couldn't just implement actix-web's `ResponseError` trait for it which would have skipped a step in the flow. Instead the logic implements the `From` trait to convert `DatabaseError` into a new struct `WebError` which then implements `ResponseError`.

I merged in the bn-db change from my artists branch (will nuke it when we release that and I rebase as they'll conflict) so that I could update some of the 404 controller responses to rely directly on this logic as well. I adjusted the logic that was expecting an Option for the User instead to check to see if the `NotFound` associated code was returned for that instance where we needed to branch off it.

Also took care of updating the venues update to match the pattern we've been using in other branches to pass in the whitelisted attributes for editing.

The end result is the code is a lot cleaner for most things (example of before and after):

```
    let connection = state.database.get_connection();
    match Event::find(parameters.id, &*connection) {
        Ok(event) => match event.update(event_parameters.into_inner(), &*connection) {
            Ok(updated_event) => HttpResponse::Ok().json(&updated_event),
            Err(e) => HttpResponse::from_error(ConvertToWebError::create_http_error(&e)),
        },
        Err(e) => HttpResponse::from_error(ConvertToWebError::create_http_error(&e)),
    }
```

https://github.com/Timo614/bn-api/blob/94df788c97bdb423b01afbfb7c618941a486996a/api/src/controllers/events.rs#L100-L107

```
    let connection = state.database.get_connection();
    let event = Event::find(parameters.id, &*connection)?;
    let updated_event = event.update(event_parameters.into_inner(), &*connection)?;
    Ok(HttpResponse::Ok().json(&updated_event))
```

https://github.com/Timo614/bn-api/blob/081022e9d653b85662d42c8143d54ceb04637fac/api/src/controllers/events.rs#L88-L91


Timo614:
The branch should be set (passes FMT and tests for me). I put it in progress though so I could confirm that nothing in this branch causes a regression with some additional manually testing. Will try to take care of that first thing tomorrow.

mikethetike:
I was literally doing the same thing, so I'm happy with this. The only thing I was doing different was returning an actual struct instead of `HttpResponse` in the `Result`, but that can be opt in.

api/src/errors/web_error.rs
```diff
@@ -0,0 +1,32 @@
+use actix_web::HttpResponse;
+use actix_web::ResponseError;
+use bigneon_db::utils::errors::DatabaseError;
+use errors::*;
+use std::error::Error as StdError;
+use std::fmt;
+
+#[derive(Debug)]
+pub struct WebError(DatabaseError);
```
mikethetike: In some libraries I've seen (e.g. Crates.io) they have a single error named after the library (`CargoError`). What do you think about calling this `BigNeonError`?

Timo614: Sounds good, I'll rename name it to that so it's consistent with that name pattern. :+1:

Timo614:
I just pushed up another commit which refactors the organization invites logic a bit:
https://github.com/big-neon/bn-api/pull/129/commits/9290beb5092c1afe42bce447963ea0c69ca81147

I removed the test that was causing the deadlock (although unsure if it's some underlying issue and we may see it again in some form) as the method it was testing no longer exists after the refactoring. I haven't seen the error again since I did that locally and I was seeing it pretty consistently before so perhaps it was something specific to that test and a race condition of some sort.

I also adjusted the create tests to check for the mail going out as we use a test transport in that environment which stores the emails in memory for inspection.

api/src/controllers/organization_invites.rs
```diff
@@ -23,134 +22,97 @@ pub struct PathParameters {
 #[derive(Deserialize)]
 pub struct NewOrgInviteRequest {
     pub user_email: String,
-    pub user_id: Option<Uuid>,
```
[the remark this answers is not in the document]
Timo614: I added the field back via: https://github.com/big-neon/bn-api/pull/129/commits/3a5a44cdd1135351c2d7f9b78b237636af0d9434

They're both optional now and in the event the user id is not provided it falls back on the email. It ignores the passed in email if the user id is valid, however.

If the given user doesn't have an email on file it panics:
https://github.com/big-neon/bn-api/pull/129/commits/3a5a44cdd1135351c2d7f9b78b237636af0d9434#diff-1063e961ca02c2bb2c80410e99596bb3R57

We currently allow users to exist without an email but that may change with #122 so can re-evaluate that in that branch. Just wanted to add back this functionality here that was refactored in the branch so that we're not potentially losing any capabilities from the refactor.

PR #131, opened 2018-08-23
Include builders in dev namespace for use in the API app

Timo614:
One of the biggest issues with the API tests we've had so far has been the amount of code devoted to prepping the environment prior to the controller actions. In most cases we also end up having to build out the associations as well as they are required for the models being tested so it ends up creating a lot of duplicate code.

To deal with that issue we added builders to the DB side but previously they were inaccessible on the API side. The change in this branch mimics actix-web's dev namespace setup (https://github.com/actix/actix-web/blob/master/src/lib.rs#L243) to provide these builders to the API as well.

The nice thing about the builder approach is adding new fields no longer requires us to update a given model create call in several files but can rely on the builders themselves. I tried to cover all of the current cases and in most situations it's a pretty big save code line / clarity wise.

An example of the change:

https://github.com/big-neon/bn-api/compare/master...Timo614:shared-builders?expand=1#diff-da64d485a3356584f2dc96470fcbc166L264

```
    let user = User::create(
        "Jeff",
        "Roen",
        "jeff@tari.com",
        "555-555-5555",
        "examplePassword",
    ).commit(&*connection)
        .unwrap();
    let user2 = User::create(
        "Jeff2",
        "Roen",
        "jeff2@tari.com",
        "555-555-5555",
        "examplePassword",
    ).commit(&*connection)
        .unwrap();
    let user3 = User::create(
        "Jeff3",
        "Roen",
        "jeff3@tari.com",
        "555-555-5555",
        "examplePassword",
    ).commit(&*connection)
        .unwrap();
    let organization = Organization::create(user.id, &"OrgName")
        .commit(&*connection)
        .unwrap();
    //create links from org to users
    let _orguser = OrganizationUser::create(organization.id, user2.id)
        .commit(&*connection)
        .unwrap();
    let _orguser2 = OrganizationUser::create(organization.id, user3.id)
        .commit(&*connection)
        .unwrap();
```

With the builders becomes:

```
    let user = database.create_user().finish();
    let user2 = database.create_user().finish();
    let user3 = database.create_user().finish();
    let organization = database
        .create_organization()
        .with_owner(&user)
        .with_user(&user2)
        .with_user(&user3)
        .finish();
```


Timo614:
I'll rebase this today -- one other option over the dev namespace is to potentially put the builders in their own crate in a similar setup as we have the db one now but I think that may be overkill / could always change it easily in the future if there was some need. Could be convinced either way.

PR #135, opened 2018-08-23
Add an API call to add tickets to a cart

mikethetike:


db/src/models/carts.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn find_for_user(user_id: Uuid, conn: &Connectable) -> Result<Cart, DatabaseError> {
+        let cart = carts::table
+            .filter(carts::user_id.eq(user_id))
+            .first(conn.get_connection())
+            .to_db_error(
+                errors::ErrorCode::QueryError,
+                "Could not load cart for user",
+            );
+        cart
```
philipr-za: Is this assignment to the cart variable needed?

db/src/models/carts.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    pub fn add_item(
+        &self,
+        ticket_allocation_id: Uuid,
+        quantity: i64,
+        conn: &Connectable,
+    ) -> Result<(), DatabaseError> {
```
philipr-za: Not returning the created cart item? Is it not something that might be required by the front-end to display a pop-up of adding this item?

db/src/models/carts.rs
```diff
@@ -0,0 +1,178 @@
...
+                errors::ErrorCode::QueryError,
+                "Could not retrieve cart item",
+            )
+    }
+
+    fn update(&self, conn: &Connectable) -> Result<(), DatabaseError> {
+        diesel::update(self)
+            .set(cart_items::quantity.eq(self.quantity))
+            .execute(conn.get_connection())
+            .map(|_| ())
```
philipr-za: Curious what this line is for? Taking in the result of the update and nulling it?

db/src/models/carts.rs
mikethetike: No, it isn't. I'll remove it

db/src/models/carts.rs
mikethetike: Personally I don't like doing that unless the method `moves` self instead of `borrowing` it. If it is borrowing, that means that there is a different instance of cart that the caller is already using, and then does not need a second instance.

mikethetike: The intention here was that the cart items should not be used as entities by themselves, but rather as part of the aggregate root of the Cart

db/src/models/carts.rs
mikethetike: The reason is that the value returned from the `diesel::update` statement is the number of rows affected, which is not meaningful to the caller of `cartItem.update(...)`.  The signature of returning an empty tuple `()` reflects that this method returns `void` or an error.

Issue #118, opened 2018-08-17
Add more fields to `/events/{id}`


Jasonvdb:
Right now this just returns: 
`{"id":"713f28fa-20cb-4d02-adf4-5c82e0c5bc56","name":"Some festival thing","organization_id":"2843e9cb-a2c5-4f59-af28-97c13519ec07","venue_id":"28c4a786-04af-455d-af42-c9e8accaf016","created_at":"2018-08-14T06:24:51.726977","ticket_sell_date":"2018-08-14T06:24:51.726977","event_start":"2018-08-30T10:00:00"}`

For a use to view all the details on an events page (https://sketch.cloud/s/dwaP2/all/fan_event-detail-page-edp/edp_onsale).

We need:
Organization name
All public artist fields (relies on artists being stored in events https://github.com/big-neon/bn-api/issues/109)
All venue fields

mikethetike:
Add number of likes to this as well

Issue #109, opened 2018-08-14
Add new fields to /events


Jasonvdb:
We need some extra fields in the model when creating and updating events.

Frontend design: https://sketch.cloud/s/dwaP2/all/promoter_events/events_create-event_add-support-artist

Basing the missing fields off the layout design above:
- Artist array [Artist ID, Set time]
- Door time
- Age limit
- Additional info
- Event status
- Event promo image URL
- Ticketing array [Name, Capacity, On sale time, Off sale time, Price in advanced, Price at the door, maximum allowed]
- Publish date (Null for immediately?)

philipr-za:
Ticketing details will be available via Ticket Allocation api calls.

PR #198, opened 2018-08-27
Add new fields to create, update and show for events

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/109
Closes https://github.com/big-neon/bn-api/issues/118

Added the fields requested in https://github.com/big-neon/bn-api/issues/109 to the events model. Only left out the ticketing array fields as those will be handled by the ticket allocations.
Added a new model called event_artist as a linking table to store artists linked to an event.

Added the logic for creating and adding new events and updated the event-builder to support them.

Modified the event show method to construct the fields required by https://github.com/big-neon/bn-api/issues/118.


api/src/controllers/events.rs
```diff
@@ -22,6 +23,27 @@ pub struct SearchParameters {
...
+#[derive(Deserialize)]
+pub struct EventArtistParameters {
+    pub artist_id: Uuid,
+    pub rank: i32,
+    pub set_time: Option<NaiveDateTime>,
+}
+
+#[derive(Deserialize, Validate)]
+pub struct CreateEventParameters {
+    pub name: String,
```
mikethetike: Rename to `CreateEventRequest`

api/src/controllers/events.rs
```diff
@@ -37,11 +59,44 @@ pub fn index(
 }
 
 pub fn show(
-    (state, parameters): (State<AppState>, Path<PathParameters>),
+    (state, parameters, user): (State<AppState>, Path<PathParameters>, User),
```
mikethetike: This path needs to be unauthed..... I think you'll need to try explicitly extract the user later in the method

api/src/controllers/events.rs
```diff
@@ -115,6 +190,28 @@ pub fn remove_interest(
     Ok(HttpResponse::Ok().json(&event_interest))
 }
 
+pub fn add_artist(
+    (state, parameters, event_artist, user): (
+        State<AppState>,
+        Path<PathParameters>,
+        Json<EventArtistParameters>,
```
mikethetike: Rename to `AddArtistRequest`

api/tests/functional/events.rs
```diff
@@ -70,21 +73,64 @@ pub fn index_search_returns_only_one_event() {
...
+    struct R {
+        event: Event,
+        organization: ShortOrganization,
+        venue: Venue,
+        artists: Vec<EventArtist>,
+        total_interest: u32,
+        user_is_interested: bool,
+    }
+
+    let event_expected_json = serde_json::to_string(&R {
```
mikethetike: It might be easier (less code) if you construct the json manually here for the test.
E.g.:
```rust

json!({
  event: {}
})

```

mikethetike: But I don't know for sure...

PR #202, opened 2018-08-27
Front end url config file checking

SWvheerden:
added front end url checking in config file so api can panic at startup if this is missing


SWvheerden:
This also fixes the seed file

api/src/config.rs
```diff
@@ -105,6 +106,9 @@ impl Config {
 
         let facebook_app_secret = env::var(&FACEBOOK_APP_SECRET).ok();
 
+        let front_end_url =
+            env::var(&FRONT_END_URL).expect(&format!("Front end url must be defined"));
+
         Config {
             allowed_origins,
```
mikethetike: Add `front_end_url`

Issue #139, opened 2018-08-23
An organization member should be able to sell a ticket to an event at the venue using their own payment system or cash


mikethetike:
Requires #289
Requires #290

PR #203, opened 2018-08-27
Beginnings of Cart Checkout

mikethetike:
I didn't want to go too far down the road with this and get a huge PR, so I've stopped here and will wait for some feedback.

Basically at this point, you can create a cart, add some tickets to it, and then check it out and convert it to an order. We'll still need to create the order line items and also at some point do the reserving and writing back to the block chain.

I re-ordered a couple of the create table sql migrations to have the foreign keys work.

Referencing issues #126 and #139


mikethetike:
Referencing issues #126 and #139

db/migrations/00000000000016_orders/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX IF EXISTS index_orders_cart_id;
```
Timo614: I think this one may be from an earlier revision?

I think we may need one for `index_orders_user_id` as well.

db/migrations/00000000000016_orders/up.sql
```diff
@@ -0,0 +1,11 @@
+
+-- Define the orders table
+CREATE TABLE orders (
+  id uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+  user_id uuid NOT NULL REFERENCES users (id),
+  status TEXT NOT NULL,
+  create_at TIMESTAMP NOT NULL DEFAULT now()
```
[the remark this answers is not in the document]
mikethetike: Correct

db/migrations/00000000000016_orders/down.sql
mikethetike: Thanks. Will add

Issue #199, opened 2018-08-27
Add org members to endpoint `/organizations/{id}`


Jasonvdb:
Need a list of users belonging to an organization

PR #210, opened 2018-08-28
Added endpoint to get list of organization members

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/199

Added the /organizations/{id}/users GET endpoint which will return the organization owner and an array of members.


api/src/controllers/organizations.rs
```diff
@@ -108,7 +116,56 @@ pub fn remove_user(
     Ok(HttpResponse::Ok().json(&organization))
 }
 
-#[derive(Serialize, Deserialize)]
-pub struct UpdateOwnerRequest {
-    pub owner_user_id: Uuid,
+pub fn show_organization_members(
```
mikethetike: In general, `list` is used instead of `show` if the REST method returns more than one item

api/src/controllers/organizations.rs
```diff
@@ -108,7 +116,56 @@ pub fn remove_user(
...
+
+    let organization = Organization::find(parameters.id, &*connection)?;
+
+    let owner = DbUser::find(&organization.owner_user_id, &*connection)?;
+
+    let organization_users =
+        OrganizationUser::find_users_by_organization(parameters.id, &*connection)?;
+
+    #[derive(Serialize)]
+    struct UserDetails {
```
mikethetike: Can we not use `DisplayUser` here?

db/src/models/organization_users.rs
```diff
@@ -43,4 +44,14 @@ impl OrganizationUser {
             user_id,
         }
     }
+
+    pub fn find_users_by_organization(
```
mikethetike: This is a bit of duplication of the `Organization::users(&self)` method. It might be worth combining them

api/src/controllers/organizations.rs
philipr-za: It has a non-optional "Created At" field that is not relevant.

api/src/controllers/organizations.rs
philipr-za: Done

api/src/controllers/organizations.rs
philipr-za: after discussion removed the Created At field from DisplayUser

db/src/models/organization_users.rs
philipr-za: Used the suggested method

PR #215, opened 2018-08-28
Add `add_user` API method

mikethetike:
This method exposes functionality to add a user directly to an organization as opposed to the invite method. This is available to OrgOwners and Admins. The invite method is still available to be used by the front end, but this allows API's and other external services to add a user.

I've also tried to fix some warnings in this PR.

And also some spelling....


db/src/models/organizations.rs
```diff
@@ -213,6 +213,14 @@ impl Organization {
...
     }
 
+    pub fn add_user(
+        &self,
+        user_id: Uuid,
+        conn: &Connectable,
+    ) -> Result<OrganizationUser, DatabaseError> {
+        OrganizationUser::create(self.id, user_id).commit(conn)
+    }
+
```
philipr-za: I think you also need to update the user's role to include OrgMember.

PR #217, opened 2018-08-28
Add Ticket Types and Price Points Entities

mikethetike:
In this PR, I've created a new entity called PricePoints to match the UI designs in bn-web. I've also pointed the CartItems to this entity instead of the TicketAllocations. I think that TicketAllocations is not the correct term, so I've replaced it with TicketTypes. 

In a future PR, I think we should remove TicketAllocations and remove Cart and CartItems.

The main reason for this PR was to solve #204, but there is still some work to be done on that side, but will remove the WIP when it is ready


philipr-za:
The only issue I see with changing TicketAllocations to PricePoint is that what if you have two different sets of tickets that have the same price but differ in other ways?

db/src/models/enums.rs
```diff
@@ -0,0 +1,33 @@
...
+                }
+            }
+
+            impl $name {
+                pub fn parse(s: &str) -> Result<$name, &'static str> {
+                    serde_json::from_str(s).map_err(|_| "Could not parse value")
+                }
+            }
+        }
+}
```
philipr-za: LGTM :P

api/src/models/encoded_ticket_type.rs
```diff
@@ -0,0 +1,38 @@
+use bigneon_db::db::Connectable;
+use bigneon_db::models::TicketType;
+use bigneon_db::utils::errors::DatabaseError;
+use models::EncodedPricePoint;
+use uuid::Uuid;
+
+#[derive(Serialize)]
+pub struct EncodedTicketType {
```
philipr-za: Maybe DisplayTicketType rather than Encoded, we have used this convention for a few other models already.

PR #232, opened 2018-08-31
Remove Cart Entity and replace with Draft Orders

mikethetike:


db/src/models/orders.rs
```diff
@@ -39,13 +43,158 @@ impl NewOrder {
...
+            )
+    }
+
+    pub fn add_tickets(
+        &self,
+        ticket_type_id: Uuid,
+        quantity: i64,
+        conn: &Connectable,
+    ) -> Result<(), DatabaseError> {
+        let item = OrderItem::find(self.id, ticket_type_id, conn)?;
```
philipr-za: This might be handled in other logic but should we check the status of the order before allowing modifying the ticket values? Only allow this if the order is in the Draft state?

db/src/models/orders.rs
```diff
@@ -39,13 +43,158 @@ impl NewOrder {
...
+        //            .to_db_error(ErrorCode::UpdateError, "Could not update cart")?;
+        //        Ok(o)
+    }
+}
+
+#[derive(Identifiable, Associations, Queryable, AsChangeset)]
+#[belongs_to(TicketType)]
+#[belongs_to(Order)]
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+pub struct OrderItem {
```
philipr-za: I agree with putting this struct in this file as they are closely coupled but this is also the first time we have put two structs in a single file so wanted to confirm that this is a file organisation practice we are happy with?

db/src/models/orders.rs
mikethetike: Good point

Issue #66, opened 2018-08-02
Return a message when API responds with an error status


Jasonvdb:
E.g. If a user doesn't exist when trying to login we need a message specifying why they can't login.

Timo614:
The login logic currently returns a generic response if the user doesn't exist in the system:
https://github.com/big-neon/bn-api/blob/master/src/controllers/auth.rs#L45

https://github.com/big-neon/bn-api/blob/dad5ada8bc67f8cd3d0158928287217957009b75/src/helpers/application.rs#L10

```
curl -H "Content-Type: application/json" -X POST -d '{"email":"test@tari.com", "password": "wrong"}' http://localhost:8088/auth/token -i
HTTP/1.1 401 Unauthorized
content-length: 39
content-type: application/json
access-control-allow-origin: *
vary: Origin
date: Wed, 15 Aug 2018 16:23:58 GMT

{"error":"Email or password incorrect"
```

That said we don't do it for a lot of the other errors (like the db generated ones) so worth us adding a response there as well.

The reasoning behind a generic login message is just to avoid leaking membership information via the login form as it would be pretty trivial to quickly test if given emails were members of the site otherwise. Membership could still be determined by creating an account and seeing if the email already existed so it's not bulletproof. 

That said, I'd imagine most hackers attempting to use our login form though don't care about individual emails but rather about a set of leaked credentials they want to attempt to login with (so we'd need a captcha and throttling/blocking of user requests based on IP & email some day likely). There's nothing embarrassing about being part of a ticket system too (versus adult entertainment membership) so just finding out a given set of emails belongs to the system isn't going to give a person much power that they didn't already have if they lack their valid credentials. I'm indifferent -- I think from a usability side it may make sense for us to drop the generic messaging I had originally gone with in favor of the email not being part of the system.

Probably worth us doing in this issue as we standardize the errors.

Jasonvdb:
Yeah I'm in favour of a more generic message for when password is incorrect or we don't even have that email. Sorry I didn't notice this had been added for auth, will implement in bn-web.

Besides the generic one I guess we'll eventually need other messages like if a user's account is locked by an admin or maybe they haven't verified their email yet. In those scenarios we would need more descriptive messages for the users so they know what to do.

If the error message format stays the same I'll implement across bn-web and just check if there's an error to display. Then as they're added to the api it'll just work.

PR #237, opened 2018-09-03
WIP Add an error message in the JSON body when a Bad Request http response is returned

philipr-za:
Addressing https://github.com/big-neon/bn-api/issues/66

This includes the API internal error message as part of the JSON body on the response. I have implemented so far for just one endpoint to get comments on if I should implement this on all the relevant endpoints.

Can you guys think of a neater way to do this?


philipr-za:
NVM found a much better way of doing this

Issue #233, opened 2018-09-03
Add the ability to delete (cancel) an event


Krakaw:
This end point should be more of a cancellation than deleting the entry out of the database.

Child of big-neon/docs#15

PR #240, opened 2018-09-03
Add ability to cancel and event

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/233

I added a new column to the events table called cancelled_at which is null to start off with and if the event is cancelled becomes set to a timestamp.

I could have just used the existing Update method in the DB layer for this but I decided to create a dedicated function so I could use the DB's now() which seemed safer than using NaiveDate.now().


api/src/controllers/organization_invites.rs
```diff
@@ -93,7 +93,8 @@ pub fn create(
     }
     //If an active invite exists for this email then first expire it before issuing the new invite.
     if let Ok(i) = OrganizationInvite::find_active_invite_by_email(&email, &*connection) {
-        i.change_invite_status(0, &*connection);
+        //This assignment squashes a warning
+        let _ = i.change_invite_status(0, &*connection);
```
mikethetike: Does this not need to be propogated with `?`

api/src/controllers/organization_invites.rs
```diff
@@ -93,7 +93,8 @@ pub fn create(
     }
     //If an active invite exists for this email then first expire it before issuing the new invite.
     if let Ok(i) = OrganizationInvite::find_active_invite_by_email(&email, &*connection) {
```
mikethetike: Are you not swallowing an error here?

db/src/models/events.rs
```diff
@@ -116,6 +119,14 @@ impl Event {
         )
     }
 
+    pub fn cancel(&self, id: Uuid, conn: &Connectable) -> Result<Event, DatabaseError> {
+        diesel::update(self)
+            .filter(events::id.eq(id))
```
mikethetike: I think this is unnecessary because you used `update(self)` and `Event` is `Identifiable`

db/src/models/events.rs
```diff
@@ -116,6 +119,14 @@ impl Event {
         )
     }
 
+    pub fn cancel(&self, id: Uuid, conn: &Connectable) -> Result<Event, DatabaseError> {
```
mikethetike: Because you are returning a new event, maybe you should take ownership of the event passed into self so that it can't be used again later?

api/src/controllers/organization_invites.rs
philipr-za: Ja

api/src/controllers/organization_invites.rs
philipr-za: Yes I am because it is equally valid that an email exists or doesn't exist depending on if this is the first invite or not. I only want to expire the previous invite IF there is an active invite and if there is none then I want to do nothing but it is not cause for throwing an error.

db/src/models/events.rs
philipr-za: So it is.

db/src/models/events.rs
philipr-za: Makes sense.

Issue #134, opened 2018-08-23
API and DB functions need transactions


mikethetike:
If an API call or DB method calls multiple methods to the database where data is changed, then they should all fail or succeed. This should be done using a transaction.

For example:
```rust
  pub fn create_from_external_login(
        external_user_id: String,
        site: String,
        access_token: String,
        conn: &Connectable,
    ) -> Result<User, DatabaseError> {
        let hash = PasswordHash::generate("random", None);
        let new_user = NewUser {
            first_name: String::from("Unknown"),
            last_name: String::from("Unknown"),
            email: NTone,
            phone: None,
            hashed_pw: hash.to_string(),
            role: vec![Roles::Guest.to_string()],
        };
        new_user.commit(&*conn).and_then(|user| {
            user.add_external_login(external_user_id, site, access_token, conn)?;
            Ok(user)
        })
    }
```
If the new user is created, then the `add_external_login call` fails, then the database is in an inconsistent state. 

This should be wrapped in a transaction.

mikethetike:
Is it worth adding an actix middleware that begins and ends the transaction?

mikethetike:
It might be worth checking out how this is done in crates.io https://github.com/rust-lang/crates.io/blob/master/src/models/krate.rs

Timo614:
Grabbing -- will take a look at crates.io as a reference.

mikethetike:
I also wasn't sure whether the closure would be possible in the API. I agree with your approach. I assume that the `response` logic in point 2 is the actix middleware response logic? You might be able to check for an error response at that point and then commit or rollback then, but see whatever is easiest to implement.

Timo614:
Not sure if the Connection trait is going to work out it has two associated types with it so they would need to be specified whenever you use the trait object. But we can't even make a trait object with it anyhow as it requires Sized https://github.com/diesel-rs/diesel/blob/765b782f7f94932cf129bc603dc5909d7155e9fe/diesel/src/connection/mod.rs#L30

We could hardcode PgConnection in place of that abstraction if we'd like to go that route. Will just hold off on changing anything for it in this branch though I think for now and just get the transaction in place for the API per above.

Timo614:
Yeah I'll have to use the middleware for the rollback as well as we have no access to the HttpRequest in the logic that converts the Db Error over to a BigNeonError for the http response. 

https://github.com/actix/actix-web/search?q=from_error&unscoped_q=from_error

[...]

PR #248, opened 2018-09-05
Transaction middleware

Timo614:
Fixes: #134 

This branch adds db transactions to the requests. It does this via an extractor and middleware. 

- The middleware looks to see if a connection is set on the request and if it is will commit the transaction if the response is a normal one and will rollback if the response was handled via the from_error logic https://github.com/actix/actix-web/search?q=from_error&unscoped_q=from_error.
- The connection is only set on the request if the Connection extractor was used by it. I did this to avoid creating connections where it's not needed like for the status endpoint. 
- I opted to retire the ConnectionGranting and Connectable traits in favor of PgConnection (to remove the abstraction as per our discussion) which cleans up the interface a bit as well. 
- Previously we used the ConnectionGranting trait to pass in a test database which stored the PgConnection used by the frontend. This isn't needed here though as in the tests we pass in the Connection as a parameter in the controller action and are not relying on the extractors (as the TestRequest logic does not use extractors we need to manually handle it as parameters). 

Note:
- I was thinking of using Deref for the Connection object to allow us to drop down to a normal PgConnection without needing to call get explicitly but by just using the `*` operator but the [documentation for that trait](https://doc.rust-lang.org/std/ops/trait.Deref.html) advises against using outside of smart pointer implementations so I didn't want to abuse it despite the nice interface it would have.


mikethetike:
Happy with it so far. We might want in future to create a `ReadOnlyConnection` object that does not create transactions, but that is for another time. 


>I was thinking of using Deref for the Connection object to allow us to drop down to a normal PgConnection without needing to call get explicitly but by just using the * operator but the documentation for that trait advises against using outside of smart pointer implementations so I didn't want to abuse it despite the nice interface it would.

Agree. I am not sure why diesel does it this way, but let's not diverge from them for now

mikethetike:
Closes #134

Timo614:
I rebased and pushed up some modifications so I think this branch may be ready. Going to improve the pull request description a bit but unless we have any pending issues on the review side I think the code here should be set.

> Happy with it so far. We might want in future to create a ReadOnlyConnection object that does not create transactions, but that is for another time.

Agreed on the ReadOnlyConnection. May be worth having the option of either using the Connection or the ReadOnlyConnection extractor as I could see some value in allowing transactions where we want to have some underlying logic revert but still log into a table. Like a webhook processing job where we want to log it was attempted but failed (as many still retry). Not sure how we'd handle the ReadOnlyConnection as well as the Connection trait used by the diesel methods has access to the `transaction_manager()` method. May be just something we have to keep an eye out for in reviews. I think it'll be pretty rare for people to do the transaction though so probably safe for the time being.

PR #252, opened 2018-09-06
Creating the entities needed for checkout

mikethetike:
This PR creates all the entities that are needed for the checkout process so that future PRs can use them to implement the API methods.

Some of the `add_to_cart` kind of tests have been commented out because the logic will change with the new structure.


Krakaw:
Looks good to me, only thing I saw was in tariclient.rs your test data the expire_date is set to Sun Apr 26 19:46:39 SAST 1970

mikethetike:
Thanks, I'll fix that when we have actual data going through it

db/migrations/00000000000025_assets/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE assets (
```
SWvheerden: currently the assets have a uuid for the database, but they have unique string id as the assit_id which is used on the communication outward. 

Perhaps store both, or use the string as primary key?

db/migrations/00000000000029_ticket_instances/up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE ticket_instances (
+  id uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+  asset_id uuid NOT NULL REFERENCES assets(id),
```
SWvheerden: see comment on assits, but here it makes sense to link to the id of the asset table

db/src/models/assets.rs
```diff
@@ -0,0 +1,43 @@
...
+use schema::assets;
+use utils::errors::ConvertToDatabaseError;
+use utils::errors::DatabaseError;
+use utils::errors::ErrorCode;
+use uuid::Uuid;
+
+#[derive(Queryable, Identifiable)]
+pub struct Asset {
+    pub id: Uuid,
+    blockchain_name: String,
```
SWvheerden: assuming this is the actual asset_id

db/src/models/assets.rs
```diff
@@ -0,0 +1,43 @@
...
+    status: String,
+    created_at: NaiveDateTime,
+    updated_at: NaiveDateTime,
+}
+
+impl Asset {
+    pub fn create(blockchain_name: String) -> NewAsset {
+        NewAsset {
+            blockchain_name,
+            status: AssetStatus::Unsynced.to_string(),
```
SWvheerden: are we saving the data this was last sync'ed?

db/migrations/00000000000025_assets/up.sql
mikethetike: I'm going to store both.

Issue #189, opened 2018-08-24
A User should be able to update their contact information


zmelamed:
child of big-neon/docs#24

Timo614:
I'll grab this one.

PR #254, opened 2018-09-06
User can update their contact information

Timo614:
Fixes: #189

This branch adds the ability for users to update their contact information. It adds a new route for `/users/me` (the route we use for getting the current user) that allows for a `PUT` request with a limited set of fields (the contact related ones).

- Opted to have the User model take an `UserEditableAttributes` for the update call. The `UserContactAttributes` model in bn-api's models uses the From trait to convert to that prior to update -- was thinking this approach allows us to have admins update users via some endpoint and still rely on the same update method in the bn-db.
- Modified register to use Created 201 code to be more consistent with our normal creation responses
- Added validators for registration and updating for the email field
- Should we require the user's password to update these fields? I'm inclined to go that direction but not sure on how that'll affect things on the UX side. Otherwise people could just take over someone's account upon stealing their access token.

Notes:
- Thought about adding the phone field for validation but it's a bit strict so we should confirm what we want to do on the frontend with that: https://github.com/Keats/validator#phone
- Not sure if we have email confirmation planned for the future (where we force a user to confirm prior to their use of the site) but if we add that we'll probably want to prevent the email field from being updated and set some new field that's temporary until the new address has been confirmed. Right now someone could modify their email to an invalid address and lose access.
- We may want to add auditing in the future for certain fields such as email so support related folks can still look up based on the previous address.


api/src/models/user_contact_attributes.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+impl From<UserContactAttributes> for UserEditableAttributes {
+    fn from(attributes: UserContactAttributes) -> Self {
+        UserEditableAttributes {
+            first_name: attributes.first_name,
+            last_name: attributes.last_name,
+            email: attributes.email,
+            phone: attributes.phone,
+            ..Default::default()
```
mikethetike: hmm... didn't realise you could do that...

Issue #259, opened 2018-09-07
Captcha on Login


Krakaw:
If there is a GOOGLE_API_KEY .env then implement recaptcha, otherwise we can implement other options at a later date.

No rate limiting on login (low risk) - this ability to brute-force, chained with the weak password requirements, would leave user accounts open to the taking. The easiest fix is to lock an account after N number of incorrect attempts for N seconds and gradually increase the time. The slightly harder would be like a capture

Child of big-neon/docs#28
Blocked by big-neon/bigneon#20

Timo614:
Grabbing

PR #264, opened 2018-09-10
Add google recaptcha support

Timo614:
Fixes: #259 

This branch adds google recaptcha support for login. 

Functionality
- If the `GOOGLE_RECAPTCHA_SECRET_KEY` environmental variable is set the logic requires the `g-recaptcha-response` parameter which is provided by the google recaptcha form element.

Examples:

Already used captcha value:
```
curl --header "Content-Type: application/json" --request POST --data '{"email":"test@tari.com","password":"secret", "g-recaptcha-response":"03AL4dnxqRxRIAZFuxFlV5Z2RdR9cumKPN3y4UnxY90-JInATdq2lBUnIqYhDkugbwNbmHs9JhHx1oDJZIgibYVTW7lvTUUk1VupgfKKYLphxf6N0qAiNcPZ9lnNPWmCihDGYVO07IPas_aepH8hZPyD0kQMTgI-8fG5ga0X_8Bp6pU2TFvzw8vK9vqDsKDWTYJIeS821umEpTROiDQ0O1s_rRA5Dv2YOuAer8yqjdNjX_SKSB_1_2yE66jZSsCyk-IMO57gWf8cFYmCrUGD85huk5-ySADd9OMg"}' http://localhost:9000/auth/token -i
HTTP/1.1 401 Unauthorized
{"error":"Captcha value invalid"}
2018-09-10 16:12:17.107 [WARN]   - Google captcha error encountered: ["timeout-or-duplicate"]
2018-09-10 16:12:17.109 [WARN]   - Unauthorized: Captcha value invalid
```

Captcha missing:
```
curl --header "Content-Type: application/json" --request POST --data '{"email":"test@tari.com","password":"secret"}' http://localhost:9000/auth/token -i
HTTP/1.1 401 Unauthorized
{"error":"Captcha required"}
2018-09-10 16:12:59.355 [WARN]   - Unauthorized: Captcha required
```

Valid captcha:
```
curl --header "Content-Type: application/json" --request POST --data '{"email":"test@tari.com","password":"secret", "g-recaptcha-response": "03AL4dnxpAxB1UnN3N--OUB4VnKUeam65o4viazOySJuJJlpEFYc8qMVzuy_MJoDijX5NFgbxs88QZDvcaA1jeDpOVJUssmaAixFMH2The2M0VfOcvDLmpo4wfAdIBaz7vGe5FyksU5C8A9P8ErCclC2Ciun3Wyyxm9WtykYFAia-kgY7wHt03qQIcxu06hG7gYkyWpY5XuN2N1x86ZpIWhC-7iNiUKNb82azv67BNP1rLXJWCtW33zMjWqwts-HG7--9GsFYrkAiMddB9ddOnHEfaqaobERw60g"}' http://localhost:9000/auth/token -i
HTTP/1.1 200 OK
{"access_token":"eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI5YzhlZjQxNi1lMDZlLTRkNDItYjAwNi1iNWNhMGU0YzU4MmYiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTM2NTk3MDA5fQ.VIeYrVwg/F19QlFS9hktCOYuIXb1JR3zikMEGjt0w84","refresh_token":"eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI5YzhlZjQxNi1lMDZlLTRkNDItYjAwNi1iNWNhMGU0YzU4MmYiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiaXNzdWVkIjoxNTM2NTk2MTA5fQ.DqwdS9Mdy36Unclj6k1jyjQKIZiKkja4vNRTCCCoaso"}
```

Notes: 
- As this relies on a third party API it does not have any test support for this feature. I'd like to implement something and spent a little while looking for some rust VCR equivalent but was unable to find a good solution. Crates.io [does do something for that problem](https://github.com/rust-lang/crates.io/blob/53b8afe35398a07d703280d9708dffda7596f996/src/tests/record.rs) so we may want to look into something similar for the future but it felt out of scope.
- Card mentioned this may be the first of many captchas after the release candidate but opted not to over-engineer the solution with traits, etc. can refactor easily later if we add more.

If useful for review, I used this simple sinatra app to generate the recaptcha codes while we're waiting on the bn-web side for adding it to the form:

```rb
require 'sinatra'

get('/') {
%q{
<html>
<head>
<script src='https://www.google.com/recaptcha/api.js'></script>
</head>
<body>
  <form method="post" action="?">
    <div class="g-recaptcha" data-sitekey="SITE_KEY_HERE"></div>
    <input type="submit">
  </form>
</body>
</html>
}
}
post('/') {
  params["g-recaptcha-response"]
}
```


mikethetike:
I'm 100% fine with this PR. If the user does not want to use the Google service, they can leave the setting blank and have no captcha (as you have implemented). IMHO, there isn't much value in adding a Rust specific version.

api/src/controllers/auth.rs
```diff
@@ -37,8 +54,35 @@ impl RefreshRequest {
...
+        HttpRequest<AppState>,
+        Connection,
+        Json<LoginRequest>,
+    ),
 ) -> Result<HttpResponse, BigNeonError> {
+    let state = http_request.state();
+    let connection_info = http_request.connection_info();
+    let remote_ip = connection_info.remote();
+
+    if state.config.environment != Environment::Test {
```
mikethetike: I don't think checking for the Environment is necessary. If the config setting is blank, then skip the check, regardless of the environment

api/src/controllers/auth.rs
```diff
@@ -83,3 +127,29 @@ pub fn token_refresh(
...
+    google_recapatcha_secret_key: &String,
+    captcha_response: &String,
+    remote_ip: Option<String>,
+) -> Result<bool, BigNeonError> {
+    let client = reqwest::Client::new();
+    let mut params = HashMap::new();
+    params.insert("secret", google_recapatcha_secret_key);
+    params.insert("response", captcha_response);
+
+    if let Some(ref remote_ip) = remote_ip {
```
mikethetike: is the `ref` necessary here?

Timo614: Yeah, I could use `&remote_ip` below but the value doesn't live long enough then. With the `ref remote_ip` here the lifetime works though.

api/src/controllers/auth.rs
Timo614: The one downside of removing the env check would be for dev machines using bn-web and bn-api who have the captcha in place for testing the full experience. My machine has the tests fail as soon (on a new tab) as I remove the check as it's still sourcing my `.env` file with it in it. I do think most development machines will not have it set though so not the end of the world.

I could switch the config itself to only be set outside of the test environment as an alternative. May add confusion if someone was to use it after seeing it in the sample env (or copy it over) and to receive errors though.

Timo614: I removed the env check for now -- I don't think most developers will locally want a captcha anyhow as it makes testing endpoints from curl, etc. much harder given you need to first generate the captcha response so probably fine.

mikethetike:
Closes #259

Issue #245, opened 2018-09-04
401 when OrgOwner tries to create/update artists


Jasonvdb:
`2018-09-04 08:45:39.336 [WARN]   - Unauthorized: Unauthorized`

Jasonvdb:
Linked to https://github.com/big-neon/bn-api/issues/222

An artist does need to be linked to an org so other orgs can't mess with it

PR #267, opened 2018-09-11
Allow OrgOwners to create and update Artists linked to their Org

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/245

This PR adds the logic for OrgOwners to add new artists via the /organizations/{id}/artists POST endpoint. It also allows those artists to be updated by OrgOwners of the relevant Org via the usual artists/{id} PUT endpoint.
Indexing and show logic was updated to cater for private artists.

I extended the test coverage on the Artist and Venue endpoints.


api/src/controllers/artists.rs
```diff
@@ -41,6 +44,18 @@ pub fn create(
     }
 }
 
+pub fn show_from_organizations(
```
mikethetike: What scenarios is this called without a user?

api/src/controllers/artists.rs
```diff
@@ -49,11 +64,17 @@ pub fn update(
...
     ),
 ) -> Result<HttpResponse, BigNeonError> {
-    if !user.has_scope(Scopes::ArtistWrite) {
-        return application::unauthorized();
-    }
     let connection = connection.get();
     let artist = Artist::find(&parameters.id, connection)?;
+
+    if !user.has_scope(Scopes::ArtistWrite)
+        && !(user.has_scope(Scopes::OrgWrite) && artist.organization_id.is_some()
```
mikethetike: It might be worth splitting this statement up or simplifying it somehow

api/src/controllers/organizations.rs
```diff
@@ -120,6 +121,29 @@ pub fn add_venue(
...
 
+pub fn add_artist(
+    (connection, parameters, new_artist, user): (
+        Connection,
+        Path<PathParameters>,
+        Json<NewArtist>,
+        User,
+    ),
+) -> Result<HttpResponse, BigNeonError> {
+    if !user.has_scope(Scopes::OrgWrite) {
```
mikethetike: We also need to check if the user is a member of this organization

api/src/controllers/venues.rs
```diff
@@ -33,12 +33,14 @@ pub fn show(
 }
 
 pub fn show_from_organizations(
-    (connection, organization_id, user): (Connection, Path<PathParameters>, User),
+    (connection, organization_id, user): (Connection, Path<PathParameters>, Option<User>),
```
mikethetike: I'm also wondering about when this would be called without a user (from the front end)

db/src/models/artists.rs
```diff
@@ -68,12 +81,30 @@ impl Artist {
...
-            ErrorCode::QueryError,
-            "Unable to load artists",
-            artists::table.load(conn),
-        )
+    pub fn all(user_id: Option<Uuid>, conn: &PgConnection) -> Result<Vec<Artist>, DatabaseError> {
+        let query = match user_id {
+            Some(u) => artists::table
+                .left_join(
+                    organization_users::table.on(artists::organization_id
+                        .eq(organization_users::organization_id.nullable())
```
mikethetike: Is this `nullable()` necessary?

db/src/models/artists.rs
```diff
@@ -84,6 +115,38 @@ impl Artist {
...
+    pub fn find_for_organization(
+        user_id: Option<Uuid>,
+        organization_id: Uuid,
+        conn: &PgConnection,
+    ) -> Result<Vec<Artist>, DatabaseError> {
+        let query = match user_id {
+            Some(u) => artists::table
+                .left_join(
+                    organization_users::table.on(artists::organization_id
+                        .eq(organization_users::organization_id.nullable())
```
mikethetike: Same question for this nullable

api/src/controllers/artists.rs
philipr-za: If someone is just browsing the website without having logged in there wouldn't be a user? Do we expect them to register before browsing artists, venues or events?

api/src/controllers/venues.rs
philipr-za: So I am not super familiar with the architecture of the frontend. As I said above it felt like we would want to allow visitors that are not logged in to browse venues, events and artists but if the frontend will never call these without a user then it makes life significantly easier on this end.

api/src/controllers/artists.rs
philipr-za: I will make the two cases separate for readability.

philipr-za: So couldn't think of a way to make it any simpler really but if you think its worth it to make it more readable I can assign the second clause to a variable called has_org_write_and_is_org_member?

api/src/controllers/organizations.rs
philipr-za: Done

db/src/models/artists.rs
philipr-za: Yes, the field is nullable in the artists table so diesel insists.

db/src/models/artists.rs
philipr-za: as above

PR #287, opened 2018-09-14
List of people interested in attending event

### LINKED ISSUE #161: A user should be able to see a list of the people who have shared that they are going to the event, or are interested in the event
child of big-neon/docs#23

A user should be able to see a list of the people who have shared that they are going to the event, or are interested in the event prioritized by distance from them in the social graph (1st degree, 2nd degree, ..., n degree)

neonknight64:
Overview:
Added the functionality to make pageable queries for retrieving a list of users that are interested in a specific event.

API Changes:
+ Added GET Route for "/events/{id}/interest"
+ Added PagingSearchParameters structure for allowing pageable search queries
+ Added "list_interested_users" Controller for Events
+ Added functional role testing for "list_interested_users"
+ Added functional base test for new even function

DB Changes:
+ Added num crate to Cargo.toml
+ Add two display structures "DisplayEventInterestedUser" and "DisplayEventInterestedUserList"
+ Added "list_interested_users" model function for event_interest
+ Added unit test for new functionality

Closes #161


db/src/models/event_interest.rs
```diff
@@ -85,4 +102,60 @@ impl EventInterest {
...
+        conn: &PgConnection,
+    ) -> Result<DisplayEventInterestedUserList, DatabaseError> {
+        //Request list of users with an interest for a specific event
+        let event_interest_list = event_interest::table
+            .filter(event_interest::event_id.eq(event_id))
+            .filter(event_interest::user_id.ne(user_id)) //Remove primary user from results
+            .inner_join(users::table)
+            .select(users::all_columns)
+            .load::<User>(conn)
+            .to_db_error(ErrorCode::QueryError, "Error loading event interest")?;
```
Timo614: Are we just using the min and max index to retrieve a subset of the results between the two indexes (I think from below)?

May make sense to just use a postgres limit call with an offset then to avoid needing to pull in all the additional user data and needing to filter it.

neonknight64:
I have made the requested changes and merged with the latest master

Issue #272, opened 2018-09-12
Calculate cart total


philipr-za:
Calculate the total of a cart/order including all the various OrderItems i.e. Ticket price, fees and potentially delivery, merch etc.

PR #295, opened 2018-09-17
Calculate the cart total

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/272

Added a method to the Order DB entity to calculate the total price of an order.
To do this I had to flesh out adding tickets to the order by determining the current ticket price, current fee schedule and creating a ticket order item and adding a fee order item.


db/src/models/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(proc_macro_derive_resolution_fallback)]
```
mikethetike: Is this needed? Perhaps put a comment saying why it is here

db/src/models/orders.rs
```diff
@@ -78,7 +81,56 @@ impl Order {
...
+        let ticket_pricing = TicketPricing::get_current_ticket_pricing(ticket_type_id, conn)?;
+
+        let organization = Organization::find(
+            Event::find(TicketType::find(ticket_type_id, conn)?.event_id, conn)?.organization_id,
+            conn,
+        )?;
+
+        if organization.fee_schedule_id.is_none() {
+            return Err(DatabaseError::new(
+                ErrorCode::NoResults,
```
mikethetike: Use a different error type here, otherwise it might become a 404 on the API

db/src/models/orders.rs
```diff
@@ -78,7 +81,56 @@ impl Order {
...
+            ));
+        }
+
+        let fee_schedule_range = FeeSchedule::find(organization.fee_schedule_id.unwrap(), conn)?
+            .get_range(ticket_pricing.price_in_cents, conn)?;
+
+        //I would like to have used an .unwrap_or_else but I dont think I can return this error value from inside the closure
+        if fee_schedule_range.is_none() {
+            return Err(DatabaseError::new(
+                ErrorCode::NoResults,
```
mikethetike: Again, use a different ErrorCode. Preferrably one that gets mapped to HTTP 500.

db/src/models/organizations.rs
```diff
@@ -286,14 +286,17 @@ impl Organization {
         &self,
         fee_schedule: &FeeSchedule,
         conn: &PgConnection,
-    ) -> Result<(), DatabaseError> {
+    ) -> Result<Organization, DatabaseError> {
+        let attributes = OrganizationEditableAttributes {
+            fee_schedule_id: Some(fee_schedule.id),
```
mikethetike: Thinking about it, the fee schedule should always be set on an organization. We should set it on creation. If needs be, we could create a Zero Fees schedule if there need to be organizations with no fees. That would make the checking code a lot simpler. Perhaps create a seperate ticket to do this.

db/src/models/orders.rs
philipr-za: I will do this, in the mean time I wanted to ask if you could think of a more Rustacean method for this check and the subsequent unwrap? I felt like I was missing an easy solution here.

db/src/models/mod.rs
philipr-za: Will do, this is what suppresses those new warnings in our code.

db/src/models/orders.rs
philipr-za: Changed to Business process error

db/src/models/orders.rs
philipr-za: Changing to BusinessProcess Error

db/src/models/organizations.rs
Jasonvdb: I'm only allowing the fee schedule to be set once, then I need to allow it to be updated?

philipr-za: So we have just made a change here that means you have to provide a fee_schedule_id when you create the Org. We dont want to allow an org to be created without a fee_schedule.

Jasonvdb: How would I create a fee schedule without having an org first?

philipr-za: Hmm that is a good point. I suppose I better include that as part of the Organzation create endpoint.

philipr-za: Just for the record we are going to open another issue to discuss how we are handling the fee schedules.

To summarise the current discussion:
- We don't think that the fee_schedule_id field of the Org table should be nullable. There should ALWAYS be a fee schedule associated with an Org
- This means we need to create a fee schedule before creating an Org
- Current work-around is to create a default zero fee schedule as the Org is created which will then be over-written in the front-end.
- Will make an urgent issue to refactor the fee schedule database entities so this process flows better.

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/299

PR #306, opened 2018-09-18
Combing endpoints for ticket pricing and ticket creation

neonknight64:
Overview: combined the endpoints for /ticket-pricing and /events/{id}/tickets, the ticket pricing endpoint was discarded. When creating a new ticket type, an array if ticket pricings are included in the request.

API Changes:
+ Changed CreateTicketTypeRequest structure to include a list of ticket pricing requests
+ Added CreateTicketPricingRequest and DisplayCreatedTicketAndPricing structures
+ Modified create_tickets() function from event controllers to perform ticket pricing aswell
+ Commented out unused functions from the ticket_pricing controller
+ Improved the create_tickets functional test to include testing of new ticket pricing system


api/src/controllers/events.rs
```diff
@@ -394,12 +394,27 @@ pub fn update_artists(
...
+    pub end_date: NaiveDateTime,
+}
+
 #[derive(Deserialize)]
 pub struct CreateTicketTypeRequest {
     pub name: String,
     pub capacity: u32,
     pub start_date: NaiveDateTime,
     pub end_date: NaiveDateTime,
+    pub ticket_pricing_list: Vec<CreateTicketPricingRequest>,
```
mikethetike: This can just be `ticket_pricing`

api/src/controllers/events.rs
```diff
@@ -459,7 +485,12 @@ pub fn create_tickets(
     //    allocation.set_asset_id(asset_id);
     //
     //    let updated_allocation = allocation.update(connection)?;
-    Ok(HttpResponse::Created().json(json!({"ticket_type_id": ticket_type.id})))
+    Ok(
+        HttpResponse::Created().json(DisplayCreatedTicketAndPricing {
+            ticket_type_id: ticket_type.id,
+            ticket_pricing_ids,
```
mikethetike: I don't know if this is necessary to return. @Jasonvdb what would you prefer to be returned here?

Jasonvdb: If I'm posting ticket pricing to the same `/events/${eventId}/tickets` then don't need anything in the request result

neonknight64: I spoke to Jason and he doesn't need any of output responses he will request them when he needs them, but how do we then test this function if it only returns a created 201 status?

neonknight64: I removed the output response for displaying the list of ticket pricing ids. Now only the primary ticket creation id will be returned as "id".

PR #310, opened 2018-09-19
Corrected list ticket types api call by including missing variables

### LINKED ISSUE #308: Update list event tickets to include missing variables

neonknight64:
Overview: Added functionality to count the capacity of an event ticket type from the issued tickets and added missing variables to the list ticket types api call.

Api changes:
+ Change DisplayTicketType struct to include a capacity variable
+ Updated the creation of a DisplayTicketType to call the db request to calculate the capacity and include it in the returned data

Db changes:
+ Added the TicketType member function "ticket_capacity" that queries the db to retrieve the asset_id and then use it to get the number of tickets issued for the event
+ Added the "get_current_ticket_capacity" unit test

Closes #308


db/src/models/ticket_types.rs
```diff
@@ -71,6 +72,22 @@ impl TicketType {
             )
     }
 
+    pub fn ticket_capacity(&self, conn: &PgConnection) -> Result<u32, DatabaseError> {
+        //Retrieve asset_id from ticket_type id in asset table
+        let asset_id: Uuid = assets::table
```
mikethetike: This can be done in one query using a join

neonknight64: Perfect, I changed it to one db query

Issue #309, opened 2018-09-19
Add current ticket pricing to `/events/{id}`


Jasonvdb:
Need to know which price points are active for which ticket types to display on the events page.

Right now using `/events/{id}/tickets/` but that doesn't tell me which ticket price is the current one.

Jasonvdb:
Also. Add status of ticket type if no ticket prices are in range or are sold out.

Timo614:
Does it make sense to also return the number of tickets remaining? Since even if it's not sold out they may only have a few remaining so you're likely to run into an error if you try to reserve greater than that.

The remaining tickets value will likely change as reserved tickets are released and people edit their carts removing tickets freeing them up so thinking maybe the best call here is to fetch the number of remaining tickets for that type from the db and return that but if it's at 0 override the normal status of `TicketTypeStatus::Published` to return `TicketTypeStatus::SoldOut`. Then as soon as more open up it'll just go back to the published state. Does that make sense here?

It would force us to do an extra query against the database each time we want to call this but not sure of a better way to figure out that x number of tickets are left. To optimize it a bit when we go to sell the tickets we could have the logic check if there are any remaining unsold tickets (reserved or not) and update accordingly to the `TicketTypeStatus::SoldOut` to avoid the extra query for ones that have none remaining.

Was thinking of something similar for the ticket prices -- for the range you're saying if they'd be active at the time the user was calling the API right? So if they don't currently have an active ticket price then returning `TicketTypeStatus::NoActivePricing` or something along those lines?

Handling this one along with the changes to #293 mentioned by Mike to lockdown that endpoint to just the organization in one pull request.

mikethetike:
> Does it make sense to also return the number of tickets remaining?  

Yes, although this might need to be obsfucated at a later stage (e.g. "<1000")

> Just confirming by sold out we mean there are currently no unreserved tickets remaining

Sold out means there are no tickets available for you to buy right now. It may be that it becomes unsold out later. So yes.

>extra query against the database each time we want to call this

Unfortunately, yes. We could risk caching this maybe next year some time, but that could lead to a very bad user experience when we get it wrong. I like the idea of updating the status though because for sold out events, they are total sold out. _However_, if the ticket status is "SoldOut", then is there ever going to be a check to see if reservations on tickets have expired?

> TicketTypeStatus::NoActivePricing

This is fine, it should show "Not available (yet?)" on the front end.

Issue #293, opened 2018-09-17
Create API to view Ticket Types and the current price point


mikethetike:

Timo614:
@mikethetike Just noticed we may already have this via #217: 
https://github.com/big-neon/bn-api/blob/cd11cfe79ef2a9c5dc0436874eed8700cd1e8859/api/src/routing.rs#L35
https://github.com/big-neon/bn-api/blob/cd11cfe79ef2a9c5dc0436874eed8700cd1e8859/api/src/controllers/events.rs#L401

Looks like it uses the ticket_pricing call to fetch the current price point:
https://github.com/big-neon/bn-api/blob/cd11cfe79ef2a9c5dc0436874eed8700cd1e8859/api/src/models/display_ticket_type.rs#L21
https://github.com/big-neon/bn-api/blob/cd11cfe79ef2a9c5dc0436874eed8700cd1e8859/db/src/models/ticket_types.rs#L55

Or is this card for something else? Just confirming since I think we have what the subject covers via 217's logic there.

Timo614:
Or is this to add a `GET` request and `show` action for the ticket types for the purposes of hitting an endpoint for an individual ticket type for its current values by id? Can add that if so just confirming.

Jasonvdb:
This endpoint can work but bn-web needs to know which current price should be used to display to the user. So this can probably be solved by just adding the `current_price` or something to each ticket type?

mikethetike:
I think there may need to be two endpoints here. The one is for the user to see, the other is for the org members to see when editing it. 
Currently, `/events/{id}/tickets` returns what is needed for the org members. It should actually be restricted as such.

I think we can perhaps add the tickets_types to the `/events/{id}` public method. In that case I suggest we show the ticket types with a single price

Jasonvdb:
Cool agreed

PR #318, opened 2018-09-20
Add ticket type data to event#show, lockdown event#list_ticket_types

Timo614:
Fixes: #293
Fixes: #309 

This branch addresses the two issues mentioned above. For #293 in the comments it was decided instead to lockdown the API to just admins and to modify the events endpoint (#309) to accomplish the goal of finding the current price point. This card does them both as the frontend would lose access to the ticket types if the changes for #293 were merged prior to the changes in #309 going out.

- Currently the logic to determine the remaining tickets is only checking against `reserved_until` as we don't have a concept of a paid ticket yet (this is true with the reservation logic too so once we merge that logic in the remaining quantity logic will just need to be adjusted to ignore paid for tickets)
- Removed the event show tests that ran for each role in favor of a single test as there isn't a difference between the roles and the behavior of the endpoint. We may want to make the user optional in the future and just not return event interest logic there as well but not sure entirely on how it'll be used and it's out of scope here so I left it requiring an authed user for now.
- In this branch I duplicated the `DisplayTicketType` model so we have an `AdminDisplayTicketType` and a `UserDisplayTicketType` as there are some fields we likely wouldn't want in the output for normal users.
- For the `AdminDisplayTicketType` I'm not currently overriding the status returned as I did for the `UserDisplayTicketType` but could add that if we think it's worth it.
- For the `UserDisplayTicketType` I'm also not showing the full capacity but could add that if we want to display it on the frontend (may be leaking too much info I was thinking when I split them but let me know if we're okay with sending it to normal users). I did add the quantity of remaining tickets to both though.


db/src/models/ticket_types.rs
```diff
@@ -60,6 +60,30 @@ impl TicketType {
             )
     }
 
+    pub fn remaining_ticket_count(&self, conn: &PgConnection) -> Result<u32, DatabaseError> {
+        let remaining_ticket_count: i64 = ticket_instances::table
+            .inner_join(assets::table)
+            .filter(assets::ticket_type_id.eq(self.id))
+            .filter(
```
mikethetike: We might need to check the status here as well when we get more statuses

PR #319, opened 2018-09-21
Updating of tickets and ticket pricing

### LINKED ISSUE #282: Allow for updating tickets and ticket prices
Right now we can only create tickets and then create price points for those tickets.

Not sure what's involved with updating ticket quantities though after already issuing assets. We can discuss here possibly.

But we'll definitely need to be able to update names and dates of these tickets at least.

Until then the bn-web form for adding tickets is going to act a little weird when a owner tries to update tickets at all.

neonknight64:
Overview: Added the ability to modify a previously created ticket type and the associated ticket pricings. Unwanted ticket pricings can be removed, new pricings can  be added and existing pricings can be modified. This works for all attributes except capacity. Partial updates are allowed and unused ticket pricings are deleted immediately, while used ticket_pricings are only marked for deletion.

API changes:
+ Added Patch route for "/events/{id}/tickets"
+ Added "update_tickets" controller with two structs "UpdateTicketTypeRequest" and "UpdateTicketPricingRequest"
+ Added the "update_tickets" functional test that creates a ticket type with ticket pricings and then modifies it using the api 
+ Changed "update_ticket" route to include venue_id and ticket_type_id in path
+ Added EventTicketPathParameters struct used for including the venue_id and ticket_type_id in path
+ Made the variables of the `UpdateTicketTypeRequest" and "UpdateTicketPricingRequest"  options
 and changed the update_ticket function to enable partial updates
+ Improve update_ticket functional test to include partial updating

DB changes:
+ Added the "TicketPricingEditableAttributes" structure
+ Added two member functions for "TicketPricing" -> "update" and "destroy"
+ Added the "TicketTypeEditableAttributes" structure
+ Added a member function for "TicketType" -> "update"
+ Added "update" unit test for "ticket_pricing"
+ Added "update" unit test for "ticket_types"
+ Added "Deleted" TicketPricingStatus
+ Modified the ticket_pricing destroy member function to check if any order_items are using the current ticket pricing. Unused ticket_pricings are deleted and used ticket_pricings are marked for deletion.
+ Added "valid_ticket_pricings" member function, this functions only returns ticket_pricings that have not been marked for deletion.

Additional changes:
+ Made small changes to order_items and ticket_instances to fix some warnings

Closes #282


db/src/models/ticket_pricing.rs
```diff
@@ -42,6 +51,23 @@ impl TicketPricing {
...
+        attributes: TicketPricingEditableAttributes,
+        conn: &PgConnection,
+    ) -> Result<TicketPricing, DatabaseError> {
+        diesel::update(self)
+            .set((attributes, ticket_pricing::updated_at.eq(dsl::now)))
+            .get_result(conn)
+            .to_db_error(ErrorCode::UpdateError, "Could not update ticket_pricing")
+    }
+
+    pub fn destroy(&self, conn: &PgConnection) -> Result<usize, DatabaseError> {
```
philipr-za: I think you need to do some checking here because this can fail if the ticket_pricing_id is being referenced in the OrderItem table. Check to see if the foreign key link (set up in migrations) will make this call fail, if thats the case then you can rely on that to throw a DatabaseError if you can't delete this. If the foreign key doesn't protect these entries it can break checkout.

api/src/controllers/events.rs
```diff
@@ -411,6 +410,25 @@ pub struct CreateTicketTypeRequest {
     pub ticket_pricing: Vec<CreateTicketPricingRequest>,
 }
 
+#[derive(Deserialize, Serialize)]
+pub struct UpdateTicketTypeRequest {
+    pub id: Uuid,
+    pub name: String,
+    pub capacity: u32,
```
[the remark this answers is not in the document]
neonknight64: It is specified as capacity on all the UI mockups, so I think this one is correct

api/src/controllers/events.rs
```diff
@@ -411,6 +410,25 @@ pub struct CreateTicketTypeRequest {
     pub ticket_pricing: Vec<CreateTicketPricingRequest>,
 }
 
+#[derive(Deserialize, Serialize)]
+pub struct UpdateTicketTypeRequest {
+    pub id: Uuid,
+    pub name: String,
```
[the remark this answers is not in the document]
mikethetike: I think `id` should be removed, since it's in the Path already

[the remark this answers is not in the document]
mikethetike: I agree in principle with Philip, but I see why you did it this way. I don't mind if you leave it as it is

api/src/controllers/events.rs
```diff
@@ -504,3 +522,74 @@ pub fn list_ticket_types(
...
+        for request_ticket_pricing in &data.ticket_pricing {
+            if request_ticket_pricing.id.is_some() {
+                if curr_ticket_pricing.id == request_ticket_pricing.id.unwrap() {
+                    found_flag = true;
+                    break;
+                }
+            }
+        }
+        if !found_flag {
+            curr_ticket_pricing.destroy(connection)?;
```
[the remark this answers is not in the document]
mikethetike: The FK will prevent it, but you will need to check Cascading deletes are not going to clear out the DB. I'm personally not a fan of cascading deletes because of the unintended effects that can be introduced in future.

I think that the `destroy` method should check if it is referenced in a reserved or purchased order, and then if so, change the status to "Deleted", otherwise it can delete it.
We'll also need to hide Deleted pricings when we are sending ticket_type data to the frontend for display

api/src/controllers/events.rs
```diff
@@ -504,3 +522,74 @@ pub fn list_ticket_types(
...
+    for curr_ticket_pricing in &data.ticket_pricing {
+        if curr_ticket_pricing.id.is_some() {
+            //Update the ticket pricing
+            let update_parameters = TicketPricingEditableAttributes {
+                name: Some(curr_ticket_pricing.name.clone()),
+                price_in_cents: Some(curr_ticket_pricing.price_in_cents),
+                start_date: Some(curr_ticket_pricing.start_date),
+                end_date: Some(curr_ticket_pricing.end_date),
+            };
+            let ticket_pricing = TicketPricing::find(curr_ticket_pricing.id.unwrap(), connection)?;
```
mikethetike: Why not do this in the previous loop when you already have the TicketPricing from the database?

api/src/controllers/events.rs
```diff
@@ -504,3 +522,74 @@ pub fn list_ticket_types(
         ticket_types: encoded_ticket_types,
     }))
 }
+
+pub fn update_tickets(
```
mikethetike: I think rename this to `replace_tickets` because the entity is updated in it's entirety

api/src/routing.rs
```diff
@@ -41,6 +41,7 @@ pub fn routes(app: &mut CorsBuilder<AppState>) -> App<AppState> {
     }).resource("/events/{id}/tickets", |r| {
         r.method(Method::GET).with(events::list_ticket_types);
         r.method(Method::POST).with(events::create_tickets);
+        r.method(Method::PATCH).with(events::update_tickets);
```
mikethetike: This behaviour is more of a replace, so it should be a PUT

db/src/models/ticket_pricing.rs
mikethetike: See my other comment

api/src/routing.rs
neonknight64: As discussed IRL, I will keep this as a patch as I made the updates more adaptable

api/src/controllers/events.rs
neonknight64: It was easier this way as the input request ticket pricings might have been out of order, but I fixed it by doing a proper search and not performing another db query.

api/src/controllers/events.rs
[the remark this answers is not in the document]
neonknight64: I updated the list ticket type system to hide the deleted pricings and added the extra destroy checks

PR #330, opened 2018-09-26
Check order expiry before marking it as paid

mikethetike:
Checks if the order is expired before it marks it as paid. 

It's not practical to try and handle all the cases in one PR. There are many edge cases, and I've asked @zmelamed for some clarity around the rules. 

I think we will being spending quite a bit of time refining this process.

I made some changes to pick up rust warnings as errors.


mikethetike:
Ugh, looks like I need to do a tiny bit of rebasing

db/src/models/payments.rs
```diff
@@ -79,7 +97,7 @@ pub struct NewPayment {
 }
 
 impl NewPayment {
-    pub fn commit(self, conn: &PgConnection) -> Result<Payment, DatabaseError> {
+    pub(crate) fn commit(self, conn: &PgConnection) -> Result<Payment, DatabaseError> {
```
philipr-za: Huh, thats a new one for me. Interesting.

PR #337, opened 2018-09-28
Create an asset and tokens on the blockchain when a TicketType and TicketInstances are created

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/296

For this issue I moved the Tari Client into its own crate. The client communicates with the mockchain using JSON-RPC over Http. 

When a TicketType is created for an event a call is made to to the mockchain to create the asset and tokens for each of the requested quantity of tickets. The mockchain id is returned and recorded in the Assets table against the asset linked to the TicketType.

The owner of the mockchain asset and tokens is currently set to the Uuid of the user that made the TicketType and so a wallet is not required yet to show ownership of the mockchain asset and tokens. I will make another issue to update the wallet implementation to generate and maintain public/private key pairs for mockchain communication once the signing functionality on the mockchain is complete.


api/src/controllers/events.rs
```diff
@@ -467,37 +470,26 @@ pub fn create_tickets(
...
-    //    };
-    //
-    //    allocation.set_asset_id(asset_id);
-    //
-    //    let updated_allocation = allocation.update(connection)?;
+
+    // TODO: move this to an async processor...
+    //let tari_client = state.get_tari_client();
+
+    let tari_asset_id = state.config.tari_client.create_asset(TariNewAsset {
```
mikethetike: We should check to see if there is a name collision. (I think it's fine to create a new issue for that)

api/src/controllers/events.rs
```diff
@@ -467,37 +470,26 @@ pub fn create_tickets(
...
+        total_supply: data.capacity as i64,
+        authorised_signers: vec![user.id().hyphenated().to_string()],
+        issuer: user.id().hyphenated().to_string(),
+        valid: true,
+        rule_flags: 0,
+        rule_metadata: "".to_string(),
+        expiry_date: data.end_date.timestamp(),
+    })?;
+
+    let asset = Asset::find(&ticket_type.id, connection)?;
```
mikethetike: Where does this Asset get created? I think this method should be `find_by_ticket_type`

db/src/models/assets.rs
```diff
@@ -28,6 +30,29 @@ impl Asset {
             status: AssetStatus::Unsynced.to_string(),
         }
     }
+
+    pub fn find(ticket_type_id: &Uuid, conn: &PgConnection) -> Result<Asset, DatabaseError> {
+        assets::table
+            .filter(assets::ticket_type_id.eq(ticket_type_id))
+            .first::<Asset>(conn)
```
mikethetike: What happens if there is more than one?

tari-client/src/tari_client.rs
```diff
@@ -0,0 +1,65 @@
+use super::tari_messages::*;
+use tari_error::TariError;
+
+use reqwest;
+use serde_json;
+
+pub trait Tari {
```
mikethetike: Tari might be a bit generic for this trait and prevents us from using it as the name for a `mod`. Maybe this trait should be `TariClient` and `TariClient` should be renamed to `HttpTariClient` or `RpcTariClient`.

api/src/controllers/events.rs
philipr-za: I agree, I have made an issue on the mockchain to enforce a uniqueness on Asset Name which will enforce this.

api/src/controllers/events.rs
philipr-za: The asset gets created in db::models::event::add_ticket_type()
I will change the method name.

db/src/models/assets.rs
philipr-za: There should not be. Should I enforce a unique constraint on ticket_type_id?

tari-client/src/tari_client.rs
philipr-za: Will do

db/src/models/assets.rs
philipr-za: I see that both the Blockchain name and the ticket_type_id have unique contraints. I have also made an issue to enforce a unique constraint in the asset name on mockchain so cannot be more than 1.

PR #350, opened 2018-10-01
Add publish method for event

### LINKED ISSUE #349: Add a publish button on events

mikethetike:
Added a simple api end point to publish events
Closes #349


db/src/models/events.rs
```diff
@@ -114,6 +117,20 @@ impl Event {
...
     }
 
+    pub fn publish(self, conn: &PgConnection) -> Result<Event, DatabaseError> {
+        if self.status() == EventStatus::Published {
+            return Event::find(self.id, conn);
+        }
+        diesel::update(&self)
+            .set((
+                events::status.eq(EventStatus::Published.to_string()),
+                events::updated_at.eq(dsl::now),
```
Timo614: I think we need a `events::publish_date.eq(dsl::now),` here as well.

The test logic is testing if it's set but that appears to be set because of the builder sets it (we probably replace it to `None` or the time now if the status is published in the builder):
https://github.com/big-neon/bn-api/blob/master/db/src/models/events.rs#L89
https://github.com/big-neon/bn-api/blob/master/db/src/test/builders/event_builder.rs#L84

philipr-za: I agree with this and have added the line and changed the builder to initialize publish_date to None. 

I am going to push this in as we need it for the Demo today.

PR #351, opened 2018-10-01
Add GET single order

mikethetike:
Closes #348


api/src/controllers/orders.rs
```diff
@@ -1,11 +1,26 @@
...
     Ok(HttpResponse::Ok().json(json!(orders)))
 }
+
+pub fn show(
+    (conn, path, user): (Connection, Path<PathParameters>, User),
+) -> Result<HttpResponse, BigNeonError> {
+    user.requires_scope(Scopes::OrderRead)?;
+    let order = Order::find(path.id, conn.get())?;
+
+    if order.user_id != user.id() {
```
Timo614: Should this be accessible to admins and organizations that own the event associated as well?

Or just locked down to the purchaser? Wondering if there will be some value in looking up a given order from a support perspective.

Issue #290, opened 2018-09-14
Transfer tickets to User's wallet when an order is paid


mikethetike:
When an order is paid in full, move the tickets from the Organization's wallet to the user on the order's wallet

Requires #289
Requires #296

PR #357, opened 2018-10-02
Transfer tokens to user on payment completion

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/290

When a payment has been completed this PR will call the tari-client to execute the token_transfer RPC method on the mockchain to transfer the tokens to be owned by the user ID.


api/src/controllers/cart.rs
```diff
@@ -97,31 +98,68 @@ pub fn checkout(
...
+
+        for (asset_id, token_ids) in tokens_per_asset.iter() {
+            let asset = Asset::find(asset_id, connection.get())?;
+            match asset.blockchain_asset_id {
+                Some(a) => state.config.tari_client.transfer_tokens(
+                    a,
+                    token_ids.clone(),
+                    user.id().hyphenated().to_string(),
+                )?,
+                None => return application::unprocessable(
```
mikethetike: Not entirely sure about this response code... 
If this asset has not yet been created, we should not charge the credit card. Maybe do a check before we do the credit card charge and throw a 500 if it's not set up correctly

db/src/test/builders/event_builder.rs
```diff
@@ -121,6 +121,12 @@ impl<'a> EventBuilder<'a> {
                     ).unwrap();
                 }
             }
+
+            let asset = Asset::find_by_ticket_type(&ticket_type.id, self.connection).unwrap();
+            let _ = asset.update_blockchain_id(
+                format!("{}.{}", event.name, ticket_type.name).to_string(),
+                self.connection,
+            );
```
mikethetike: missing `.unwrap()'?

api/src/controllers/cart.rs
philipr-za: Easy enough to do this however we will have the converse problem of reversing the token transfer if the payment fails. We cannot make the token transfer atomic so it will require a reversal but that is doable. I don't think we will have any issues with clashes in the mockchain as the tokens are issued based on Ticket Instances and reserving those is managed by the Order logic. 

Tim mentions maybe dealing with refunds in other situations so I thought I would make sure we want to go the token transfer reversal route before I implement.

We currently have much more control over the mockchain process than Stripe so I think we should rather go that route than Stripe refunds when we can.

db/src/test/builders/event_builder.rs
philipr-za: Agreed, then it will panic if there is a problem with the builder.

api/src/controllers/cart.rs
philipr-za: After speaking to Mike I realized that this comment did not mean to do the whole transfer before payment but just make sure that the asset has been assigned on the blockchain so that everything is ready to do the transfer after the payment is completed.

Issue #342, opened 2018-09-28
Extend cart expiry time when user attempts to pay


Jasonvdb:
We don't want users getting cut off as they submit their credit card details if stripe takes a while to process. 

A possible solution would be an endpoint to be used to extend the time by a minute or 2 when bn-web is retrieving a stripe token.

A little dodgy to do though if a user figures it out and keeps extending their cart time. Maybe we only allow an expiry time to be extended once per cart?

mikethetike:
Perhaps we can set the status to PendingPayment (perhaps AwaitingPaymentConfirmation?) that only expires after a few days. It opens us up to certain other problems, but I think that we will avoid the main problem of selling a ticket twice, and then having to work out who has the ticket

Jasonvdb:
I'm happy. But we still need someway to empty the cart if it expires. Will get the expiry time now to display to a user so we can communicate that with them. But I think right now they just stuck with a cart they can't use.

philipr-za:
I actually made a bug issue about that problem https://github.com/big-neon/bn-api/issues/356

PR #360, opened 2018-10-03
Extend expiry time on order/cart when a user starts payment process

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/342

Added status to ticket_instances to mark them as Available, Reserved, Paid and I extend the the expiry time on the Order and Ticket instances when a payment process starts.


db/src/models/orders.rs
```diff
@@ -298,6 +297,17 @@ impl Order {
         self.updated_at = db_record.updated_at;
         self.status = db_record.status;
 
+        //Extend the reserved_until time for tickets associated with this order
+        let order_items = OrderItem::find_for_order(db_record.id, conn)?;
+
+        for item in &order_items {
+            TicketInstance::update_reserved_time(
+                item,
+                Utc::now().naive_utc() + Duration::days(1),
```
mikethetike: Rather grab this time at the start of this method so that all of the tickets and the order have the same expiry time

philipr-za: 👍 Done.

Issue #183, opened 2018-08-24
An Organization member should be able to scan a tickets QR code to check a User into an event


zmelamed:
child of big-neon/docs#20

PR #364, opened 2018-10-04
Endpoint to redeem a ticket using the ticket_id and redeem key

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/183

The PR adds a redeem_key to ticket_instances which are generated on completion of payment. An endpoint was created that allows a user with the TicketAdmin scope to redeem a ticket using the ticket_id and redeem_key which will be the data that is stored in the QR Code.


db/src/utils/errors.rs
```diff
@@ -50,6 +52,8 @@ pub fn get_error_message(code: ErrorCode) -> (i32, String) {
         // Logical/Business errors - 7000 range. These represent errors
         // that arise from an invalid setup in the database
         BusinessProcessError => (7000, "Business Process error"),
+        TicketAlreadyRedeemedError => (7001, "Ticket has already been redeemed"),
```
mikethetike: This seems very specific to be here. Can't we use a generic error instead?

db/src/models/ticket_instances.rs
```diff
@@ -107,6 +110,17 @@ impl TicketInstance {
         Ok(grouped_display_tickets)
     }
 
+    //Used only for functional tests
+    pub fn find_full_instance(
```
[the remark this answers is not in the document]
philipr-za: I renamed the existing find to find_for_display and renamed this function as suggested.

db/src/utils/errors.rs
philipr-za: We decided this was not actually an error case and decided to communicate these as results from the redeem() function.

Issue #184, opened 2018-08-24
An Organization member should be able to look up a users name to check a User into an event


zmelamed:
child of big-neon/docs#20

PR #367, opened 2018-10-05
WIP #184 search for users at an event

mikethetike:


api/src/controllers/events.rs
```diff
@@ -414,3 +414,27 @@ pub fn update_artists(
 
     Ok(HttpResponse::Ok().json(&added_artists))
 }
+
+#[derive(Deserialize)]
+pub struct GuestListQueryParameters {
+    pub q: String,
```
philipr-za: q? For future readability this should be a bit more descriptive.

mikethetike: q is often used as a query parameter for "Search by this"

philipr-za: If its a common thing then no problems.

Issue #275, opened 2018-09-13
Event search n+1 query


Timo614:
I think we have a N+1 query with the Venue lookup for the event search results display. With smaller sets of data it probably won't be too bad but could get pretty slow with large result sets. 

https://github.com/big-neon/bn-api/blob/e8f9b900e8739c8e5a5ecdc47913b8339da80bb5/api/src/controllers/events.rs#L89

Not sure on the most efficient way to handle that with diesel as it's pretty hands off with the associations / I doubt there's anything built in to eager load the data, etc.

PR #372, opened 2018-10-05
Fix event n+1 query issue

Timo614:
Fixes: #275 

This branch attempts to fix the associated issue by fetching all of the venues all at once and then using a hash map to populate the result information returned. It'll be most helpful in situations where we have many events happening at one venue returned (a venue lookup sort of situation) but also should improve performance even when all the venues are different as it only does one query for the venues used across all the events instead of one for each event returned. I didn't add any tests for the API as nothing in the response has changed.

I did check diesel and there's nothing build in to eager load associations like this. Makes sense since we're just getting back a regular struct. We could have it return the venue as part of the event search itself though if that were more ideal (either in some transitional struct which includes the event and venue information or as a tuple for each event). I think this approach is fine for now though but let me know and can modify as needed.

The sort and dedup may not be necessary as postgres won't return multiple records regardless of how many times you specify the id but I figured it was better to avoid sending a giant list of duplicated ids. Perhaps it's fine though and we can remove those lines.

Example (just events endpoint returning an event without a venue and one with one):
```
http://localhost:9000/events
[{"id":"60cd3d4b-7dbc-467f-8e99-8dd52d7b03df","name":"event two","organization_id":"cb98289d-03ad-4b20-abcd-c1b18ba56414","venue_id":null,"created_at":"2018-10-05T14:27:12.215608","event_start":"2018-10-05T13:56:28.872973","door_time":null,"status":"Published","publish_date":"2018-10-05T13:56:28.872973","promo_image_url":null,"additional_info":null,"age_limit":null,"cancelled_at":null,"venue":null},{"id":"7155c44a-bc23-496e-a58d-18d656149bf8","name":"It's my party","organization_id":"cb98289d-03ad-4b20-abcd-c1b18ba56414","venue_id":"612c2803-c4c4-4f8d-aac9-af77dea5905a","created_at":"2018-10-05T13:56:28.872973","event_start":"2018-11-13T12:00:00","door_time":null,"status":"Published","publish_date":"2018-10-05T13:56:28.915057","promo_image_url":null,"additional_info":null,"age_limit":null,"cancelled_at":null,"venue":{"id":"612c2803-c4c4-4f8d-aac9-af77dea5905a","region_id":null,"organization_id":null,"is_private":false,"name":"Test venue_1538747789","address":null,"city":null,"state":null,"country":null,"postal_code":null,"phone":null,"created_at":"2018-10-05T13:56:28.831603","updated_at":"2018-10-05T13:56:28.831603"}}]
```


mikethetike:
Happy with this. If the queries do not hit the benchmarks in future, we can optimize it then

PR #380, opened 2018-10-09
Sw Index result pagination data structure for list of interested users in event

SWvheerden:
Moved the result over to return the new standard response for indexes.


api/src/models/paging.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[derive(Serialize)]
+///struct used to indicate paging information and search query information
+pub struct Paging{
+    pub page : u64,
+    pub limit : u64,
+    pub sort : bool,
+    pub dir : SortingDir,
+    pub total : u64,
+    pub tags : Vec<serde_json::Value>,
```
mikethetike: I think this should be a `Vec<(string, Vec<string>)>` or a Map of some kind

api/src/models/paging.rs
```diff
@@ -0,0 +1,26 @@
...
+    Asc,
+    Desc,
+}
+
+#[derive(Serialize)]
+///struct used to indicate paging information and search query information
+pub struct Paging{
+    pub page : u64,
+    pub limit : u64,
+    pub sort : bool,
```
mikethetike: Shouldn't there always be a sort? Also it should probably be a string e.g. `name`, `price`

db/tests/unit/event_interest.rs
```diff
@@ -113,8 +113,8 @@ fn list_interested_users() {
         let result = EventInterest::list_interested_users(
             primary_event.id,
             primary_user.id,
-            request_from_index,
-            request_to_index,
+            request_from_index as u64,
```
mikethetike: Don't these need to be updated to `page` and `limit`?

api/src/models/paging.rs
```diff
@@ -1,26 +1,73 @@
...
+}
+
+#[derive(Serialize, Deserialize)]
+///struct used to indicate paging information and search query information
+pub struct PagingParameters{
+    pub page : Option<u64>,
+    pub limit : Option<u64>,
+    pub sort : Option<String>,
+    pub dir : Option<SortingDir>,
+    pub total : Option<u64>,
```
mikethetike: When would you pass in `total` as a parameter?

api/src/models/paging.rs
```diff
@@ -1,26 +1,73 @@
...
+    pub tags : Option<Vec<SearchParams>>,
+}
+
+impl Paging {
+    pub fn new(received : &PagingParameters)-> Self{
+        let default_page = if let Some(i) = received.page {i} else {0};
```
mikethetike: Maybe `clippy` has a way of shortening these lines

api/src/models/paging.rs
SWvheerden: Never. Because we have to pass back most of these we have two options:
1: As is currently with the field option, but it is populated on return.
2: We duplicate the struct for return with the total added.

api/src/models/paging.rs
SWvheerden: clippy isnt picking up anything here

api/src/models/paging.rs
mikethetike: So can it be removed?

mikethetike:
@SWvheerden Don't forget to add some Postman tests or edit the existing ones

Issue #376, opened 2018-10-08
Add an fee_in_cents to events


Krakaw:
A single fee for an event

PR #383, opened 2018-10-09
Add flat fee per event

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/376

This PR adds an optional flat fee for an event. This flat fee is applied to any order that includes any number of items from the given event. 

I renamed the existing Fees OrderItem type to FeesPerUnit (these are the previous fees that are applied for each unit of an order item based on the organizations fee_schedule) and created a new OrderItem type called EventFees. The logic to cater for these EventFees when adding and removing tickets and the Show Cart logic has been updated to account for these new fees as well.


db/src/models/order_items.rs
```diff
@@ -114,22 +124,24 @@ impl OrderItem {
            oi.unit_price_in_cents,
            oi.item_type,
            CASE
-             WHEN item_type = 'Fees' THEN 'Fees'
+             WHEN item_type = 'PerUnitFees' THEN 'Ticket Fees'
+             WHEN item_type = 'EventFees' THEN 'Event Fees - ' || e.name
              ELSE e.name || ' - ' || tt.name END AS description
         FROM order_items oi
+           INNER JOIN events e ON event_id = e.id
```
mikethetike: This doesn't seem right. I think moving it out of the nested Left join maybe return a different result set. Perhaps there should be two joins to `events`

philipr-za: When the SQL Wizard speaks we obey.

Issue #169, opened 2018-08-24
A User should be able to purchase tickets using a previously saved payment method


zmelamed:
child of big-neon/docs#22

Timo614:
Most of this should be set via #362 as it allows the saving of Stripe Customer data to be reused for a future purchase. Guess we just need to surface the cards via an endpoint that returns some basic data (ie card ending 2222) via https://stripe.com/docs/api#list_cards and then allow the frontend to pass along the selected card to make the purchase. Or if we're not surfacing the a set of cards for them to pick just charges using the default (although seems to make sense to pass that list back with one including a default flag).

Will grab this once the other card is merged unless someone else does beforehand.

Timo614:
I've spent some time with the logic and noticed we only allow one Stripe card right now using whatever is set as the default for the customer. I guess for the MVP that's probably fine so will move forward with this approach and not implement the list of cards for the customer to then pass in when selecting Stripe (ie: multiple card options for that provider). We can circle back with a different card to add that capability in the future as that seems out of scope here now that I've spent some time with this logic. 

So for this going to:

1. When a new token is used and stored the payment method is updated to use that in place of the old (the logic has a uniqueness constraint on the payment method)
2. Add endpoint to return payment methods for use in checkout
3. Allow the payment method to be used in place of the token when the customer has previously saved it

PR #387, opened 2018-10-09
Purchase tickets using saved methods

Timo614:
Fixes: #169

Changes:
- Fixed issue with using saved customer token / needed to use 'customer' parameter in place of 'source' for Stripe API
- Fixed issue where logic would have an error if you attempted to checkout using an already saved payment option (it now updates the payment option to use the new token)
- Adds endpoint to return payment methods that are saved for the user
- Adds `PaymentMethod` type for checkout which uses a stored payment method in place of the token for purchasing
- Adds postman tests for the saved card flow and purchasing using stored payment method

Note:
- We're mixing terminology a bit so we may want to refactor that a little bit in the future (the payment method's `provider` is their reuse token and the `name` field there is the `provider` we use with checkout.

We may want to also make new issues for the following:
- Add means to update existing payment method
- Allow payment method to be deleted with endpoint
- Display more information about payment method (last 4 etc?) specific to providers (felt out of scope here)
- May be worth us finding some way to add functional tests for Stripe by mocking it somehow but perhaps not needed with postman

Examples:

Saving a new payment on an existing stored payment method:
```
curl http://localhost:9000/cart/checkout --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI1MmNiYTQ0Ny1kZjg2LTRlZTktOGQxNi1hMmE5ZWRmYzdkN2IiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTM5MTI0Mzg3fQ.klWE3Dg5Agehv+MVNQKO0Bgk34DXOyraou4AaBVHguk" -X POST  --header "Content-Type: application/json" --data '{"amount": 1000, "method": {"type": "Card", "provider": "stripe", "token": "tok_1DJTZwJIAfNqtxWr9socAZE6", "save_payment_method": true, "set_default": true}}'
{"payment_id":"06bceb3c-ab28-4d67-91a2-6484d1938e97"}
```

Making a payment using a saved method:
```
curl http://localhost:9000/cart/checkout --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI1MmNiYTQ0Ny1kZjg2LTRlZTktOGQxNi1hMmE5ZWRmYzdkN2IiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTM5MTI0Mzg3fQ.klWE3Dg5Agehv+MVNQKO0Bgk34DXOyraou4AaBVHguk" -X POST  --header "Content-Type: application/json" --data '{"amount": 1000, "method": {"type": "PaymentMethod", "provider": "stripe"}}'
{"payment_id":"f6eeafee-b3c5-4ad6-a9eb-6b61724e1e09"}
```

Checking saved payment methods:
```
curl http://localhost:9000/payment_methods --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI1MmNiYTQ0Ny1kZjg2LTRlZTktOGQxNi1hMmE5ZWRmYzdkN2IiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTM5MTI0Mzg3fQ.klWE3Dg5Agehv+MVNQKO0Bgk34DXOyraou4AaBVHguk"
[{"name":"stripe","is_default":true}]
```

Stripe dashboard event log showing behavior.

screen shot 2018-10-09 at 6 47 37 pm
screen shot 2018-10-09 at 6 47 56 pm

1. Lowest is the normal checkout with saving where it stores the customer record and card
2. Above that is the logic where a new payment is made and stored for the given customer so the old card is deleted and a new one added ([this is Stripe's behavior on their end when it updates for a new souce](https://stripe.com/docs/api#update_customer-source) although we can support multiple cards via the cards API in the future if we add that but it's out of scope for this card and likely not an MVP need)

> A Token’s or a Source’s ID, as returned by Elements. Passing source will create a new source object, make it the new customer default source, and delete the old customer default if one exists. If you want to add additional sources instead of replacing the existing default, use the card creation API. Whenever you attach a card to a customer, Stripe will automatically validate the card.

3. Top payment is using the payment method to reuse the saved card


db/src/models/payment_methods.rs
```diff
@@ -34,6 +42,49 @@ impl PaymentMethod {
...
+            )
+    }
+
+    pub fn update(
+        &self,
+        attributes: &PaymentMethodEditableAttributes,
+        conn: &PgConnection,
+    ) -> Result<PaymentMethod, DatabaseError> {
+        let query =
+            diesel::update(self).set((attributes, payment_methods::updated_at.eq(dsl::now)));
```
mikethetike: Can we store a DomainEvent here to keep track of the old `provider_data`?

db/src/models/users.rs
```diff
@@ -214,6 +214,29 @@ impl User {
...
+        conn: &PgConnection,
+    ) -> Result<PaymentMethod, DatabaseError> {
+        let mut payment_methods = PaymentMethod::find_for_user(self.id, Some(name), conn)?;
+        if payment_methods.is_empty() {
+            Err(DatabaseError::new(
+                ErrorCode::NoResults,
+                Some("No payment method found for user"),
+            ))
+        } else {
+            Ok(payment_methods.remove(0))
```
mikethetike: Shouldn't we check for `is_default`? or is that handled somewhere else?

Timo614: Have it returning if it's the default via the payment methods index but could modify the cart logic to have the provider name (ie: stripe) as an optional field and if it's not there just fallback to the default. Will make that change. :+1:

db/src/models/payment_methods.rs
Timo614: Sounds good, will add that

api/src/controllers/cart.rs
```diff
@@ -146,6 +149,37 @@ pub fn checkout(
         PaymentRequest::External { reference } => {
             checkout_external(&connection, &mut order, reference, &req, &user)?
         }
+        PaymentRequest::PaymentMethod { provider } => {
+            let provider = match provider {
+                Some(provider) => provider.clone(),
+                None => match user
```
Timo614: Broke this apart over using `unwrap_or_else` as I couldn't return the failure state when there wasn't a default set on the account using that method (given it would need to return a string).

One note: it does end up doing an extra query here as the logic below ends up fetching the payment method again based on the provided name here. Since it's just one small SQL query it seemed worth doing this over adding some logic to pass around an optional `PaymentMethod` to the `checkout_payment_processor` method for the case where the default is being used. If we find it worth it later we could further refactor this to remove the extra SQL call but I think the cleaner interface is worth it for now since it's pretty minor cost but could be convinced otherwise if anyone feels strongly about it.

PR #390, opened 2018-10-10
Add paging info to artists

SWvheerden:
Let all artist calls that return a Vec<> use the paging struct


api/src/controllers/artists.rs
```diff
@@ -1,18 +1,27 @@
...
     let artists = match user {
         Some(u) => Artist::all(Some(u.id()), connection.get())?,
         None => Artist::all(None, connection.get())?,
     };
-    Ok(HttpResponse::Ok().json(&artists))
+    let query_parameters = Paging::new(&query_parameters.into_inner());
+    let artist_count = artists.len();
+    let payload = Payload {
+        data: artists,
+        paging: Paging::clone_with_new_total(&query_parameters, artist_count as u64),
```
mikethetike: Maybe we should return a `Paging` object with `page=0, limit=artist_count, total=artist_count`

PR #392, opened 2018-10-10
Update tari client and BN-Api to make use of new chain message structure, message signatures and authorisation

### LINKED ISSUE #338: Add management of private/public keys to wallet implementation for Asset and Token creation and transfer
Once the signing is implemented in the mockchain we will need to generate public/private key pairs for use when creating Assets and Transferring and Redeeming tokens on the mockchain via the Tari Client.

### LINKED ISSUE #365: Update tari client to use new chain message structure

neonknight64:
Overview of changes:
- Updated cart checkout and creation of ticket types to use the private and public keys of an orgs wallet to do transfers
- Updated the wallet structures and db tables to include a private key, public key and default flag
- Added functions to the wallet to add, find and create wallets for users and orgs
- Added cryptographic library for message signing and verifications
- Added new message structures
- Modified Client to make use of private and public keys

Closes #365 
Closes #338


api/src/controllers/cart.rs
```diff
@@ -158,18 +162,22 @@ pub fn checkout(
             *save_payment_method,
             *set_default,
             &state.service_locator,
-        ),
-    }?;
+        )?,
+    };
 
     if payment_response.status() == StatusCode::OK {
```
mikethetike: This should actually check that the order is complete, not that the payment was successful. I think it's fine to create an issue for this to be done later though

api/src/controllers/cart.rs
```diff
@@ -158,18 +162,22 @@ pub fn checkout(
...
-                token_ids.clone(),
-                user.id().hyphenated().to_string(),
-            )?,
+            Some(a) => {
+                let wallet_id=walletid_per_asset.get(asset_id).unwrap().clone();
+                let org_wallet = Wallet::find(wallet_id, connection.get())?;
+                state.config.tari_client.transfer_tokens(&org_wallet.secret_key, &org_wallet.public_key,
+                                                         &a,
+                                                         token_ids.clone(),
+                                                         user.id().hyphenated().to_string(),
```
mikethetike: Should this not be the user's wallet, not his id?

db/src/models/wallets.rs
```diff
@@ -3,34 +3,74 @@ use diesel;
...
+    pub fn create_for_user(user_id: Uuid, name: String, conn: &PgConnection) -> NewWallet {
+        let (secret_key, public_key) = tari_client::cryptographic_keypair();
+        let default_flag: bool;
+        match Wallet::find_for_user(user_id, conn) {
+            Ok(v) => if v.len() == 0 {
+                default_flag = true;
+            } else {
+                default_flag = false;
+            },
+            Err(_e) => default_flag = true,
```
mikethetike: Change this method to return a Result<NewWallet, DatabaseError> and propogate this error up with `?`

db/src/models/wallets.rs
```diff
@@ -3,34 +3,74 @@ use diesel;
...
+    ) -> NewWallet {
+        let (secret_key, public_key) = tari_client::cryptographic_keypair();
+        let default_flag: bool;
+        match Wallet::find_for_organization(organization_id, conn) {
+            Ok(v) => if v.len() == 0 {
+                default_flag = true;
+            } else {
+                default_flag = false;
+            },
+            Err(_e) => default_flag = true,
```
mikethetike: Same

api/src/controllers/cart.rs
```diff
@@ -117,15 +117,19 @@ pub fn checkout(
 
     let order_items = order.items(connection.get())?;
 
-    //Assemble token ids for each asset in the order
+    //Assemble token ids and ticket instance ids for each asset in the order
     let mut tokens_per_asset: HashMap<Uuid, Vec<u64>> = HashMap::new();
+    let mut walletid_per_asset: HashMap<Uuid, Uuid> = HashMap::new();
```
philipr-za: Just a style thing but if we are sticking to snake case this should be wallet_id_per_asset. walletid reads weirdly.

db/src/models/wallets.rs
mikethetike: You can probably merge the `.commit()` method in here into one method. Usually it's in two steps: 1. Create the object, then 2. Save to the database. Because you are already hitting the database, it should just be one method

db/src/models/wallets.rs
```diff
@@ -3,34 +3,74 @@ use diesel;
...
+        };
         NewWallet {
             user_id: Some(user_id),
             name,
+            secret_key: tari_client::convert_bytes_to_hexstring(&secret_key),
```
philipr-za: If you use: use tari_client::{convert_bytes_to_hexstring, cryptographic_keypair}; as your include you can exclude all these references to tari_client in the body of the code.

db/src/models/wallets.rs
```diff
@@ -41,6 +81,61 @@ impl Wallet {
...
+            for wallet in wallets {
+                if wallet.default_flag {
+                    result_wallet = wallet;
+                    break;
+                }
+            }
+        } else {
+            //Create default wallet for user
+            result_wallet =
+                Wallet::create_for_user(user_id, String::from("Default"), conn).commit(conn)?;
```
mikethetike: We usually use `"Default".to_string()`

db/tests/unit/events.rs
```diff
@@ -561,6 +566,10 @@ fn add_ticket_type() {
 fn ticket_types() {
     let project = TestProject::new();
     let event = project.create_event().finish();
+    let wallet_id =
+        Wallet::find_default_for_organization(event.organization_id, &project.get_connection())
```
mikethetike: It might be useful to create a helper method on `Event` `event.issuer_wallet(&conn)` which does this code. It seems to be repeated a couple time

api/src/controllers/cart.rs
philipr-za: Ja, I think the owner should actually be the PuclicKeyHash of the user?

api/src/controllers/cart.rs
neonknight64: Changed to wallet_id_per_asset

api/src/controllers/cart.rs
neonknight64: Ticket created

api/src/controllers/cart.rs
neonknight64: Made the correct, the public key of the user will be used as for the new owner.

db/src/models/wallets.rs
neonknight64: Made the changes

db/src/models/wallets.rs
neonknight64: Perfect, made the improvement

db/src/models/wallets.rs
neonknight64: Made the change

db/src/models/wallets.rs
neonknight64: Made the change

db/tests/unit/events.rs
neonknight64: Perfect, made the change

PR #407, opened 2018-10-11
Added paging struct to venue calls

SWvheerden:
Added paging wrapper to venue calls


api/src/controllers/venues.rs
```diff
@@ -25,15 +34,28 @@ pub fn show(
...
         }
         None => Venue::find_for_organization(None, organization_id.id, connection.get())?,
     };
-    Ok(HttpResponse::Ok().json(&venues))
+    let venues_count = venues.len();
+    let mut payload = Payload {
+        data: venues,
+        paging: Paging::clone_with_new_total(&query_parameters, venues_count as u64),
+    };
+    payload.paging.limit = venues_count as u64;
```
Krakaw: I'm not quite following why this is being set to the number of venues we find?
Shouldn't this be set either by the client in the params `?limit=n` or the default value of x (defined by env or some other method)?

Issue #403, opened 2018-10-11
Change PUT `/events/{id}/artists` to accept an `artists` param which will be the array


Jasonvdb:
Right now PUT `/events/{id}/artists` receives just one array as a param. 

Need it to receive a key/value with `artists` value being an array. 

This is because of the way bn-api-node passes through params. 

Like {artists: ['Bon Jovi']}

Timo614:
Fixed via #408

PR #408, opened 2018-10-11
Sw change put `/events/{id}/artists` to accept an `artists` param which will be the array

SWvheerden:
Fixes issue #403


api/src/controllers/events.rs
```diff
@@ -85,12 +85,17 @@ pub struct AddArtistRequest {
...
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, Default)]
 pub struct UpdateArtistsRequest {
     pub artist_id: Uuid,
     pub set_time: Option<NaiveDateTime>,
 }
 
+#[derive(Deserialize, Debug, Default)]
+pub struct UpdateArtistsRequestList {
+    pub update_req: Vec<UpdateArtistsRequest>,
```
Timo614: Shouldn't we make this `artists`?

SWvheerden: Its not updating the complete artist. 
Just this: pub set_time: Option<NaiveDateTime>
I beleive doing it this way is a much smaller request. 

This was done on request for @Jasonvdb ,  So perhaps he and @mikethetike  can make a comment here?

Timo614: I was just commenting on the name of the parameter not its type as the associated issue and this PR's title call it `artists` but it's named `update_req` here.

SWvheerden: Well I am open to change the name, but I would steer away from "artist" as that would conflict with the artist struct. 

as to the PR name, its the exact name of the issue that was open

mikethetike: I agree with @Timo614 here. I think `artists` will make more sense to the front end developers consuming this service

PR #410, opened 2018-10-12
Added more Ticket Types

mikethetike:
Just a simple change to add multiple ticket types for an event


mikethetike:
Yeah, I added it in to try and replicate a bug and then thought it might be worth keeping

Issue #175, opened 2018-08-24
A user must be able to select multiple tickets to send as a group to a user via QR code


zmelamed:
child of big-neon/docs#21

Issue #174, opened 2018-08-24
A user must be able to send an individual ticket to a user via QR code


zmelamed:
child of big-neon/docs#21

Jasonvdb:
I have the UX ready in bn-web, whoever picks this up please let me know what data I need to add to the QR code so the mobile app can sign the transaction.

PR #417, opened 2018-10-15
Transferring tickets via QR code

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/174
Closes https://github.com/big-neon/bn-api/issues/175

This PR implements the endpoints for a user to authorise the transfer of tickets that they have purchased for a given period. A transfer authorization data object is returned which can be embedded in a QR code. Another user can then receive these tickets by provided the transfer authorization data to a second endpoint.

The POST /tickets/transfer endpoint accepts a JSON body with an array of ticket_ids (Must be purchased by the currently logged in user) and the time this transfer is valid for e.g.
```
{
	"ticket_ids":[
		"b760e94e-3ee5-4ff2-ba98-614cef93334c",
		"675fb132-5920-449e-872c-c4b212821976"],
	"validity_period_in_seconds":3600
}
```
If the tickets are purchased and available this endpoint will return the Transfer Authorization Data which looks as follows:
```
{
    "transfer_key": "fabe3a4a-5a60-4dea-9030-2b52dbcb5e61",
    "sender_user_id": "0f60defd-16f3-4dd7-a9e9-f6a7256a032a",
    "num_tickets": 2,
    "signature": "304402203f46e575c3e7b1e2ed795437a3b17e248f70c6d10f129d8a3f26da896793f5a702202ff5c31b09bfd0b52e3ab6918de98a80bbe7398085dfaee6e2dab9ed1e430e32"
}
```
The signature is computed using the sending users private key stored in their default wallet.

Another user can then post this data to the POST /tickets/receive endpoint while they are logged in. The signature will be verified by looking up the sending users public key. If all the data is valid, the transfer expiry time has not elapsed and the tickets are still available for transfer they will be transferred to the current user.


db/src/models/ticket_instances.rs
```diff
@@ -358,6 +379,156 @@ impl TicketInstance {
...
+            .to_db_error(ErrorCode::UpdateError, "Could not update ticket instance")?;
+        }
+
+        //Generate transfer_key and store keys and set transfer_expiry date
+        let transfer_key = Uuid::new_v4();
+        let transfer_expiry_date =
+            Utc::now().naive_utc() + Duration::seconds(validity_period_in_seconds as i64);
+
+        diesel::update(
+            ticket_instances::table.filter(ticket_instances::id.eq(dsl::any(ticket_ids.clone()))),
```
mikethetike: Perhaps add an explicit check here that the user on the ticket instance is still the current user and that the transfer key is still null

db/src/models/ticket_instances.rs
```diff
@@ -358,6 +379,156 @@ impl TicketInstance {
...
+        //Perform transfer
+        let receiver_wallet = Wallet::find_default_for_user(receiver_user_id, conn)?;
+
+        diesel::update(
+            ticket_instances::table
+                .filter(ticket_instances::id.eq(dsl::any(ticket_ids_to_transfer))),
+        ).set((
+            ticket_instances::wallet_id.eq(receiver_wallet.id),
+            ticket_instances::updated_at.eq(dsl::now),
+        )).execute(conn)
```
mikethetike: Add check that rows affected is equal to the number of tickets sent

db/src/models/ticket_instances.rs
```diff
@@ -358,6 +379,156 @@ impl TicketInstance {
...
+                ErrorCode::BusinessProcessError,
+                Some("Cannot transfer tickets.".to_string()),
+            ));
+        }
+        //Perform transfer
+        let receiver_wallet = Wallet::find_default_for_user(receiver_user_id, conn)?;
+
+        diesel::update(
+            ticket_instances::table
+                .filter(ticket_instances::id.eq(dsl::any(ticket_ids_to_transfer))),
```
mikethetike: Again, add concurrency check

db/src/models/ticket_instances.rs
```diff
@@ -358,6 +379,156 @@ impl TicketInstance {
...
+            return Err(DatabaseError::new(
+                ErrorCode::BusinessProcessError,
+                Some("User does not own all requested tickets".to_string()),
+            ));
+        }
+
+        //If any of the requested ticket instances have an existing transfer key all instances of that transfer key
+        //should be cleared.
+        let mut existing_transfer_keys: Vec<Uuid> = Vec::new();
+        for t in &tickets {
```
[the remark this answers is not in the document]
philipr-za: Agreed, I used this configuration to keep the code neater by using the contains in the loop as the outer data structure is a composite struct not just Uuid's. Happy to change if you think the performance saving will be worthwhile?

db/src/models/ticket_instances.rs
```diff
@@ -358,6 +379,156 @@ impl TicketInstance {
...
+            if ticket_ids.contains(&t.id) && t.transfer_key.is_some() {
+                existing_transfer_keys.push(t.transfer_key.unwrap());
+            }
+        }
+        if existing_transfer_keys.len() > 0 {
+            diesel::update(
+                ticket_instances::table
+                    .filter(ticket_instances::transfer_key.eq(dsl::any(existing_transfer_keys))),
```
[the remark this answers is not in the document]
philipr-za: Removed this section after we realized that we check that the number of transfers is checked on receiving them and if they are not all there the transfer cannot be received.

db/src/models/ticket_instances.rs
philipr-za: Added the check for wallet_id and updated_at. Cannot really assume that the transfer_key is null as it will not be on a second transfer.

db/src/models/ticket_instances.rs
philipr-za: Done

db/src/models/ticket_instances.rs
philipr-za: Done

PR #432, opened 2018-10-17
Transfer tokens on chain after successful sending of tickets from user to user

### LINKED ISSUE #415: Transfer tokens on the mockchain after successful sending of tickets from user to user
After a user receives/claims a ticket transfer then the token ownership should be transferred to the user on the mockchain

Blocked by https://github.com/big-neon/bn-api/issues/175

neonknight64:
Modified the receive_transfer tickets controller to also transfer the tokens of tickets on the chain. The receive_ticket_transfer test had to be updated as well.

Closes #415


db/src/models/ticket_instances.rs
```diff
@@ -455,16 +455,17 @@ impl TicketInstance {
         transfer_authorization: TransferAuthorization,
         receiver_user_id: Uuid,
         conn: &PgConnection,
-    ) -> Result<(), DatabaseError> {
+    ) -> Result<(Wallet, Wallet, Vec<TicketInstance>), DatabaseError> {
```
mikethetike: The data returned here seems very specific to the caller of this method which makes it less reusable. Should we maybe move the retrieval of this data into a second method?

PR #442, opened 2018-10-18
Allow user to create a ticket hold

### LINKED ISSUE #423: Create a hold list
Child of big-neon/docs#19
Child of #421

Holds will take a ticket out of the inventory and (TBC transfer the block of tickets to another wallet) these can then be paid for by the third party (with a code / at the door)

mikethetike:
Closes #423 

Add the API action to create a ticket hold. I also changed the asset name in the block chain to use the event id instead of the event name because there were occassionally duplicates


mikethetike:
I still need to add some test for these methods.

PR #444, opened 2018-10-19
Allow updating of ticket capacity for an event

### LINKED ISSUE #146: An `Organization Admin` and `Organization User` to update ticket allocations for an event
child of big-neon/docs#19

neonknight64:
Overview of changes:
- Modified the ticket types update controller so that the ticket capacity can be updated locally and on the chain
- Added two chain functions:  modify_asset_increase_supply and modify_asset_nullify_tokens
- Added two ticket type model functions for counting ticket instances: ticket_count and valid_ticket_count
- Added function find_valid_tickets_for_asset_id for returning all local ticket instances that are still available 
- Add function nullify_tickets for marking local tickets as unusable
- Modified the TicketInstance function create_multiple so that you can provide the starting tari id

Closes #146


api/src/controllers/ticket_types.rs
```diff
@@ -175,14 +176,82 @@ pub fn update(
...
+            let asset = Asset::find_by_ticket_type(&ticket_type.id, connection)?;
+            let tickets = TicketInstance::find_valid_tickets_for_asset_id(asset.id, connection)?;
+            if (nullify_ticket_count as usize) < tickets.len() {
+                let org_wallet =
+                    Wallet::find_default_for_organization(event.organization_id, connection)?;
+                //Nullify tickets locally
+                let ticket_instance_ids: Vec<Uuid> = (0..nullify_ticket_count)
+                    .map(|i| tickets[i as usize].id)
+                    .collect();
+                TicketInstance::nullify_tickets(&ticket_instance_ids, connection)?;
```
mikethetike: I think it will be better to call a sql query that ensures the locking like we do for releasing reserved ticket instances, and then to return the ids of the tickets we nullified, and then only use those ids to nullify the data on the block chain

neonknight64: Perfect, I made this change

db/src/models/ticket_instances.rs
```diff
@@ -159,8 +159,21 @@ impl TicketInstance {
             .to_db_error(ErrorCode::QueryError, "Could not load ticket instances")
     }
 
+    pub fn find_valid_tickets_for_asset_id(
```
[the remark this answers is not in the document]
neonknight64: Not using this function anymore as I made the changes of your next comment

db/src/queries/nullify_tickets.sql
```diff
@@ -0,0 +1,22 @@
+UPDATE ticket_instances
+SET
+  status = 'Nullified'
+WHERE id IN (SELECT t.id
+             FROM ticket_instances AS t
+             WHERE t.asset_id = $1 AND t.status = 'Available'
```
mikethetike:
```suggestion
             WHERE t.asset_id = $1 AND t.status = 'Available' AND t.hold_id IS NULL
```

db/src/queries/nullify_tickets.sql
```diff
@@ -0,0 +1,22 @@
...
+             LIMIT $2 FOR UPDATE SKIP LOCKED)
+    RETURNING
+      id,
+      asset_id,
+      token_id,
+      ticket_holding_id,
```
mikethetike:
```suggestion
      hold_id,
```

mikethetike: This PR will requires #442

Issue #422, opened 2018-10-15
Create a list of comps


Krakaw:
Child of big-neon/docs#19
Child of #421

Comps are also taken out of inventory but they have a flag set / their price set to 0 and they are transferred to their destination wallet

Timo614:
Just confirming the behavior here / confirming the direction I've been going is correct for this one:

- An event is created with x number of tickets of varying ticket types
- An organization owner or member can create comps for a given ticket type which keeps the tickets in the organization's wallet and marks them as being Comp tickets which makes them unsellable
- The organization can transfer those tickets to another user

So they never go through the order process just are moved around as free tickets from the organization's wallet to the user in questions?

"Comps are also taken out of inventory" so from that I was thinking that ticket types have assets created for them and then the comps request marks a portion of them as unsellable and available for giving out as comps? Or should the logic work by allowing a comp request to create new assets as if they weren't included to begin with? Guess I'm wondering what the flow is here for the generation of these if we're taking them out of the available set or if the available set never included them prior to their addition if that makes sense.

Timo614:
@zmelamed can you confirm the flow above?

mikethetike:
I think it's important that the Order is still created for these comps, so that the financial reports can reflect them easily. The order's amount should just be 0. The API/front end shouldn't explicitly create an order, it should be created in the backend

zmelamed:
Hey @Krakaw @Timo614 @mikethetike , I put together some flows here (also added this to big-neon/docs#19) to get across how this might work with turning a hold into a named list ... is this possible with our implementation... it's essentially taking a hold, and then adding a list of names which are assigned allotments of that hold (kind of holding part of the hold)... These names would be returned to the scanner app and box office to allow those tickets to be claimed or checked in without them having to be on the users device... Then eventually we could create a link to share with an external list admin (such as the band who has tickets held) where they could self manage the names on that "list" whcih is also described in this flow to some degree. 
![holds-lists-flows](https://user-images.githubusercontent.com/456953/47399648-56b40900-d6ee-11e8-9352-38a4a30f2578.png)

mikethetike:
So a comp list is basically just a hold with a 100% discount?

zmelamed:
Yes... a hold on tickets with a 100% discount and no fee. A held allotment of complimentary tix=a comp hold, and if you have a list of names attached to that you get a comp list.

Timo614:
[...]

PR #445, opened 2018-10-19
WIP: Complimentary tickets

Timo614:
Closes: #422 

WIP, will add more later. I left a note in the associated issue but handling this by marking a subset of tickets as complimentary and then allowing those to be transferred by any organization member with the ComplimentaryTransfer scope. Still need to add tests and make sure there aren't any areas I've missed with the logic. May end up adjusting this further as well flow wise depending upon if the direction I was going here is incorrect (clarifying via 422).


mikethetike:
I added some comments to #422. I don't know why wafflebot assigned me to this though...

Timo614:
@mikethetike waiting for Zach's comments before I touch this further to get clarification there on the flow (and also working on other responsibilities at the moment so won't be able to work on bigneon for at least another day) but can switch it to whatever flow we think is best.

That's fine switching it to use a Comp wallet instead and using an order. I hadn't released we wanted to have an order for these and the associated issues were a bit light on details so went with the direction of making it a new ticket type status versus purchased as these weren't 'Purchased' per se.

Once the associated issue is updated with whatever the business team thinks is best will circle back and adjust to use an order instead.

mikethetike:
> a bit light on details

:joy: yip

Timo614:
Reviewing the associated issue and flow Zach mentioned. Going to close this PR and make a new branch and PR for the changes mentioned there as it's hold based so this logic won't be useful for it.

PR #457, opened 2018-10-25
Make Cart more concurrent

### LINKED ISSUE #352: Add concurrency check for add tickets to cart

mikethetike:
Closes #352 

Sorry for the big pull request. This ended up being a really large amount of work. I think we will need to look at refactoring the Order structures into something more manageable because it's really clunky dealing with it at the moment.

Any feedback on style is also welcome. 

This is a really import part of the system, so please take some time to review it carefully


api/src/controllers/cart.rs
```diff
@@ -57,9 +52,12 @@ pub fn add(
         .group_by(|request_item| request_item.ticket_type_id)
     {
         let quantity = request_items.fold(0, |sum, request_item| sum + request_item.quantity);
+
         cart.add_tickets(ticket_type_id, quantity, connection)?;
     }
 
+    cart.update_event_fees(connection)?;
```
mikethetike: I've moved the event fees update out of the 'add tickets' section because it was calculating them for every item that was added. This means that other parts of the code that call it need to call this as well.

api/src/controllers/cart.rs
```diff
@@ -74,7 +72,7 @@ pub fn remove(
 ) -> Result<HttpResponse, BigNeonError> {
     let connection = connection.get();
     // Find the current cart of the user, if it exists.
-    let current_cart = Order::find_cart_for_user(user.id(), connection).optional()?;
+    let current_cart = Order::find_cart_for_user(user.id(), connection)?;
```
mikethetike: I see the remove method must still be updated. I would like to merge these two into a single 'update` with just a ticket type and a quantity, but that might need to be in another PR

db/src/bin.rs
```diff
@@ -1,45 +1,29 @@
...
+#![deny(unused_imports)]
+// Unused results is more often than not an error
+#![deny(unused_must_use)]
+#![deny(unused_extern_crates)]
 
 #[macro_use]
 extern crate diesel_migrations;
+extern crate bigneon_db;
 extern crate clap;
-
```
mikethetike: These were no longer necessary when using bigneon_db as a crate

db/src/models/holds.rs
```diff
@@ -105,6 +105,19 @@ impl Hold {
                 "Could not load organization for hold",
             )
     }
+
+    pub fn find_by_redemption_code(
```
mikethetike: This will be used at a later stage when adding tickets from a hold

db/src/models/orders.rs
```diff
@@ -49,19 +51,21 @@ impl NewOrder {
 }
 
 impl Order {
-    pub fn create(user_id: Uuid, order_type: OrderTypes) -> NewOrder {
```
mikethetike: `create` has been removed so that we can update the user row as well

db/tests/unit/orders.rs
```diff
@@ -412,11 +385,9 @@ fn find_for_user_for_display() {
     order2
         .add_external_payment("test".to_string(), user.id, 500, project.get_connection())
         .unwrap();
-    let order3 = project.create_order().for_user(&user).finish();
 
     assert_eq!(order1.status, OrderStatus::Paid.to_string());
     assert_eq!(order2.status, OrderStatus::PartiallyPaid.to_string());
-    assert_eq!(order3.status, OrderStatus::Draft.to_string());
```
mikethetike: It is no longer possible to have two open orders that are carts

db/src/models/orders.rs
```diff
@@ -70,17 +74,77 @@ impl Order {
...
+        let cart_id: Vec<R> = diesel::sql_query(query)
+            .bind::<sql_types::Uuid, _>(user.id)
+            .bind::<sql_types::Timestamp, _>(Utc::now().naive_utc() + Duration::minutes(15))
+            .get_results(conn)
+            .to_db_error(ErrorCode::QueryError, "Could not find or create cart")?;
+
+        if cart_id.is_empty() || cart_id[0].id.is_none() || cart_id.len() > 1 {
+            // Another thread has created a cart
+            return DatabaseError::concurrency_error(&format!(
+                "Possible race condition when creating a cart for a user. ID returned: {}",
```
philipr-za: It is just an error so no big deal but the text implies you want to print an ID of a cart but you return the number of carts.

mikethetike: Good point

PR #458, opened 2018-10-25
Used sql_query bind instead of format! for invites

Krakaw:


db/src/models/organization_invites.rs
```diff
@@ -129,38 +120,47 @@ impl OrganizationInvite {
...
+            .to_db_error(ErrorCode::QueryError, "Cannot find organization invite")
     }
 
     pub fn get_invite_details(
         token: &Uuid,
         conn: &PgConnection,
     ) -> Result<OrganizationInvite, DatabaseError> {
-        let expiredate = Utc::now().naive_utc() - Duration::days(INVITE_EXPIRATION_PERIOD_IN_DAYS);
+        let expiry_date = Utc::now().naive_utc() - Duration::days(INVITE_EXPIRATION_PERIOD_IN_DAYS);
+        let query = r#"
```
Timo614: May be worth switching this to use diesel's DSL to avoid the need to handle the bind for this one since it's a simple query.

Something like:

```
        DatabaseError::wrap(
            ErrorCode::AccessError,
            "No valid token found",
            organization_invites::table
                .filter(organization_invites::security_token.eq(token))
                .filter(organization_invites::created_at.gt(expiry_date))
                .filter(accepted.is_null())
                .get_result(conn),
        )
```

PR #472, opened 2018-10-29
Add list holds route and simplify paging

### LINKED ISSUE #450: List all holds for an OrgMember

mikethetike:
Closes #450 

Added a list of holds for an event.

I also simplified the paging interface, so that it requires fewer lines in each method


Timo614:
Will review first thing tomorrow if no one has a chance before then to take a look. Sorry about that meant to take a look earlier but it slipped my mind.

PR #476, opened 2018-10-29
Modifications for invite issues

Timo614:
While going through for the Comps branch I noticed a way to specify more custom declarations in the SQL select aspect of the Diesel DSL so went to quickly switch this over to use that over custom SQL as the query itself outside of the select was pretty straightforward from a diesel perspective. 

I noticed a bug when I went to do it with the way it was binding the interval value for the expiration date. It seems if you bind the variable as part of the interval call it is ignored so the results for a record created today with 7 days added for expiration were just a single day:

```
{"expires_at":"2018-10-30T17:37:00.355207","inviter_name":"System Administrator","organization_name":"Jazzy_1540834220"}
```

With the fix:

```
{"expires_at":"2018-11-05T17:37:00.355207","inviter_name":"System Administrator","organization_name":"Jazzy_1540834220"}
```

There may be other ways to accomplish it like casting things around but this seemed the most straightforward approach to me at the time. The bind while using SQL fragments acts as an inline bind which is why it looks the way it does there versus other binds we are using in the code.


mikethetike:
Nice. Didn't know you could bind in a select statement

PR #478, opened 2018-10-30
Allow an OrgMember to split holds

### LINKED ISSUE #473: Split a hold
child of big-neon/docs#19

mikethetike:
Closes #473 

I also added the ticket_type_id directly onto the holds, otherwise it is too difficult to determine which tickets to reserve in the new split


db/src/models/holds.rs
```diff
@@ -99,6 +102,38 @@ impl Hold {
...
+            self.event_id,
+            redemption_code,
+            discount_in_cents,
+            end_at,
+            max_per_order,
+            hold_type,
+            self.ticket_type_id,
+        ).commit(conn)?;
+
+        TicketInstance::add_to_hold(
```
Timo614: With the Comps branch I had the logic do a validation when the hold's `set_quantity` was called to verify that the new quantity wasn't going to drop below the number of comps we have assigned (since they are like a hold of a hold). We may need to move the validation somehow and just reverify in this case that the old hold is still valid after the modification or we may end up with a hold that has less quantity available than the held comps. Since we do the rollback could just make the validation its own method and call it both there and here.

mikethetike: I was relying on the TicketInstance::reserve SQL LIMIT to check if there was not enough left in the hold, but I see that doesn't sort out the comps problem. One way would be to add `comp_id` onto the ticket instance to reserve them. Another would be to make each comp a mini hold that has a parent hold.

At the moment it seems that if you have the access code for the hold that the comps belong to, then you can buy through all the tickets and then into the comps allocation

Timo614: I'll go the `comp_id` route via a new issue as the mocks show them as part of one hold so the mini hold approach would force them to be seen as individual ones (or require some additional logic to combine all mini holds for display etc which would get complicated). I'll file a new issue for that aspect and making sure the splitting of holds doesn't eat into that supply. There's an issue for purchasing comps via the redemption code (#482) so we can handle the case of those codes not eating into the assigned set there. Will make a note of it in that issue.

PR #498, opened 2018-10-31
Add available to list of holds

mikethetike:


api/src/controllers/events.rs
```diff
@@ -517,5 +517,41 @@ pub fn holds(
     )?;
     let holds = Hold::find_for_event(path.id, conn)?;
 
-    Ok(HttpResponse::Ok().json(Payload::new(holds, query.into_inner().into())))
+    #[derive(Serialize)]
+    struct R {
```
Timo614: May be worth DRYing this up with a new struct that's reused here and in holds controller. Could have a method for it that takes the hold and connection and returns this type.

PR #499, opened 2018-10-31
Try and speed up docker builds and reduce image size

CjS77:
Several wins in this PR wrt improving docker build times and image size

Firstly, `.dockerignore` massively reduces the build context from 3-5GB to <10MB

Then we use multibuilds to extract more gains:

* We first create a dummy project with all the dependencies and compile it. Since this list _doesn't change often_, it is stored as an image and can be used as a cache. No more downloading and compiling on each build.
* Then we _copy_ the binaries to a fresh, smaller base image. `1.30-slim` is 950MB vs 2.5GB, but we can potentially reduce this further (in production) by using Alpine (<100 MB)


CjS77:
Looks like aligning versions fpr `reqwest` breaks things 🤦‍♂️

PR #500, opened 2018-11-01
Event ticket price min and max cache

### LINKED ISSUE #440: Add min and max price to events
When Adding or changing ticket types, set the event min and max price fields as a caching field - this is so that when displaying events you can display this event costs between `min` and `max`

Krakaw:
Closes #440 

Added the fields min_ticket_price_cache and max_ticket_price_cache.
Any time the ticket prices are changed the cache is updated.


api/src/controllers/events.rs
```diff
@@ -131,6 +131,8 @@ pub fn index(
         age_limit: Option<i32>,
         cancelled_at: Option<NaiveDateTime>,
         venue: Option<Venue>,
+        min_ticket_price_cache: Option<i64>,
```
mikethetike: For the db models etc. I think `min_ticket_price_cache` is fine, but when it goes over the wire to the client, `min_ticket_price` is better

mikethetike:
```suggestion
        min_ticket_price: Option<i64>,
```

api/src/controllers/events.rs
```diff
@@ -164,6 +166,8 @@ pub fn index(
             top_line_info: event.top_line_info,
             age_limit: event.age_limit,
             cancelled_at: event.cancelled_at,
+            min_ticket_price_cache: event.min_ticket_price_cache,
```
mikethetike:
```suggestion
            min_ticket_price: event.min_ticket_price_cache,
```

api/src/controllers/events.rs
```diff
@@ -226,6 +229,8 @@ pub fn show(
         ticket_types: Vec<UserDisplayTicketType>,
         total_interest: u32,
         user_is_interested: bool,
+        min_ticket_price_cache: Option<i64>,
```
mikethetike:
```suggestion
        min_ticket_price: Option<i64>,
```

api/src/controllers/events.rs
```diff
@@ -252,6 +257,8 @@ pub fn show(
         ticket_types: display_ticket_types,
         total_interest,
         user_is_interested: user_interest,
+        min_ticket_price_cache: event.min_ticket_price_cache,
```
mikethetike:
```suggestion
        min_ticket_price: event.min_ticket_price_cache,
```

db/src/models/events.rs
```diff
@@ -117,6 +128,40 @@ impl Event {
...
+                prices.push(ticket_pricing.price_in_cents);
+                //                if ticket_pricing.price_in_cents < min_ticket_price_cache {
+                //                    min_ticket_price_cache = ticket_pricing.price_in_cents.clone();
+                //                }
+                //                if ticket_pricing.price_in_cents > max_ticket_price_cache {
+                //                    max_ticket_price_cache = ticket_pricing.price_in_cents.clone();
+                //                }
+            }
+        }
+        let min_ticket_price_cache = prices.clone().into_iter().min();
```
mikethetike: I think if you use `.iter()` instead of `.into_iter()` you won't have to `.clone` here

api/src/controllers/events.rs
Krakaw: Agreed.

Issue #446, opened 2018-10-22
Create discounts codes


Krakaw:
See big-neon/docs#19

Timo614:
@zmelamed @Krakaw couple clarifications for Discount codes:

1. Are discount codes applied before fees? Or fee inclusive? So if the discount code is for $10 and the ticket is $8 but $9 with fees it'd be 100% free? Or would cost $1 for the fee and only discount at most to the ticket price?
2. What sort of flow are you envisioning for setting discount codes? An endpoint to add it to the cart and then it's reflected as that total line item? So endpoints to add and clear discount code? What happens if an order is in process and the discount code is made invalid (reaches end date, runs out of max uses)? Does the cart view return an error on the frontend and just wait to have it cleared via that endpoint but be otherwise unable to checkout in that state due to an error given the associated discount code is no longer valid?
3. > When a user applies a discount code, the Ticket Types pricing shown to them within the app and during checkout should reflect the discounted price, and the total discounts should be shown as a line item in the checkout summary.

So if my understanding in the second item and this text is correct they'd add the discount code to apply to their current cart. Then the event ticket types endpoint would show the applicable ticket pricing with the discount applied. Then in the cart view the tickets would reflect their normal pricing but the total discounts line would show the full discounts applied across all tickets. Is this correct?

Just to visualize what my understanding of this is... 

If there are two ticket types for $4 and for $8 and there's a discount code for both that gives $6 off then:
- The ticket types related pricing shows $0 and $2
- If they buy one of each checkout shows line items for $4 and $8 but with a total discount of $10 ($4 off and $6 off)
4. We only allow for one discount code per order right?
5. What behavior should we show when a user reaches their `Max per User` but is buying additional tickets? The ticket types endpoint should revert to regular pricing? What happens if they're only 2 tickets away from that with an order for 3 tickets? When they add them to the cart they would have seen the discounted pricing via the ticket types endpoint but the cart would only reflect that pricing for 2 of them? Or is that okay?

zmelamed:
@Timo614 Thank you for this, all really great insights and questions. Please push back on my responses if you find more holes. 

1. Discount codes are applied before fees, impacting the face value, thus potentially lowering the fee if it knocks it into a lower range on the fee schedule.
Discounts should not be able to be lower than face of the tickets they're related to. If they were the order should be treated as a comp. No fees.

[...]

PR #501, opened 2018-11-02
Create and manage discounts

Timo614:
Fixes: #446

This branch addresses the creation and management aspect of #446. 

Primarily:
1. CRUD related endpoints for discount creation and updating (linked to events as per the mockups)
2. Validations for discounts for data integrity where needed
3. Ability to associate ticket types for a given event with the discounts via a vector of `ticket_type_ids`
4. Associated unit, functional, and postman tests for discounts behavior

Outside of scope:
- Most of what we discussed in https://github.com/big-neon/bn-api/issues/446#issuecomment-434578054 -- I was just clarifying for many for future use cases but most of that is for the use of discount codes which is a big issue in it of itself.

Validations:
- Start before end date
- Redemption code 6 char min and confirm no other model level validations missing
- Redemption code unique on a per event level (constraint and validation) across multiple models
- Confirm ticket type is valid for discount

Future validations (when we deal with those issues / implementations) but not done here in this branch to restrict scope (making a note here so we keep track of it as considered adding them here but opted not):
- Order is within discount window during purchasing
- Ticket association with discount code to check max per user
- Max uses does not exceed discount max uses via association lookup (constraint and validation)

Not sure if we want to do the following (probably worth their own issues if so):
1. Switch holds to use discount_id and apply the discount when the hold is activated? Perhaps using the same redemption_code as the hold?
2. Switch the validation for redemption code to constrict it to just a given event and remove uniqueness constraint on redemption code per table we currently have? One downside of that approach is if multiple events have active redemption codes with the same value they'd be triggered and show discounts (although not likely to happen). Just noticed the docs mentioned to make it unique per event but right now we have it system unique.

Note:
1. Added a method to set a discount on an order but leaving the work there for pricing, etc. for another issue to constrain scope here.
2. For now only made the redemption codes unique across discounts and holds -- wasn't sure what the comp specific redemption codes shown in the mocks were used for but should be easy enough to add if those are used in the same logic.
3. Tried and failed to get diesel's dsl working with an subquery resulting in an array for `Discount::find_for_event` so ended up with plain SQL. May be worth refactoring later but I spent way too long attempting to get it to work and reading through the docs so decided it wasn't worth the effort since the plain SQL works just as well.

```
discounts::table
    .filter(discounts::event_id.eq(event_id))
    .select((
        discounts::id,
        discounts::name,
        discounts::event_id,
        discounts::redemption_code,
        discounts::max_uses,
        discounts::discount_in_cents,
        discounts::start_date,
        discounts::end_date,
        discounts::max_tickets_per_user,
        discounts::created_at,
        discounts::updated_at,
        dsl::array::<dUuid, _>(
            ticket_type_discounts::table
                .filter(ticket_type_discounts::discount_id.eq(discounts::id))
                .select(ticket_type_discounts::ticket_type_id)
        ),
    )).get_results(conn)
    .to_db_error(
        ErrorCode::QueryError,
        "Could not retrieve discounts for event",
    )
```


Timo614:
I just looked over the mockups / writeup for access codes and it sounds near identical to discount codes with a few small variations (discount is optional not required, gates access to a ticket allocation).

I may switch this over to make it more generic then naming the table `codes` or something along the line with a `code_type` of discount since it seems like we'd have to reuse a lot of the same logic here.

Issue #437, opened 2018-10-17
User should be able to specify a Hold in the call to add to cart


mikethetike:
connected to #446

PR #507, opened 2018-11-02
Change Add to Cart to Set Quantity in cart and add hold/comp to cart

mikethetike:
Connected to  #437 

Overview
---
I've removed the add/remove to cart api methods and replaced them with a set quantity method. This will prevent any duplicate calls or out of order calls from corrupting the user's cart. I've also changed it to calculate the fees at the end.

Notes
---

This PR doesn't sort out all of the issues with discounts, holds and comps, but there is some work done on it


mikethetike:
I still need to add some tests

db/src/models/orders.rs
```diff
@@ -162,51 +163,158 @@ impl Order {
...
-        let event = Event::find(ticket_type.event_id, conn)?;
-        let organization = Organization::find(event.organization_id, conn)?;
-
-        let fee_schedule_range = FeeSchedule::find(organization.fee_schedule_id, conn)?
-            .get_range(ticket_pricing.price_in_cents, conn)?
-            .unwrap();
+    ) -> Result<(), DatabaseError> {
+        self.lock_version(conn)?;
+        let mut mapped = vec![];
+        for item in items.iter().enumerate() {
```
philipr-za: For readability maybe use a tuple here [i.e. for (ind, item) in items.iter().enumerate() ], would be easier to understand without having to look at the enumerate docs.

db/src/models/orders.rs
```diff
@@ -162,51 +163,158 @@ impl Order {
...
 
-        TicketInstance::reserve_tickets(
-            &order_item,
-            &self.expires_at,
-            ticket_type_id,
-            None,
-            quantity,
-            conn,
-        )
+        Ok(())
```
philipr-za: Crikey, this is quite a function....looks alright though.

db/src/models/organizations.rs
```diff
@@ -73,7 +73,7 @@ pub struct OrganizationEditableAttributes {
     pub postal_code: Option<String>,
     pub phone: Option<String>,
     pub fee_schedule_id: Option<Uuid>,
-    pub event_fee_in_cents: Option<i64>,
+    pub event_fee_in_cents: Option<Option<i64>>,
```
Timo614: Is this so we can set it to `null`? If so we'll need that serde custom deserializer to treat the logic differently depending upon if the attribute is present or not.

`#[serde(default, deserialize_with = "deserialize_some")]`

Otherwise it'll always deserialize to `None` whether or not the attribute is present.

db/migrations/00000000000031_add_ticket_instances_to_orders/down.sql
```diff
@@ -11,9 +13,3 @@ ALTER TABLE order_items
 
 ALTER TABLE order_items
   DROP COLUMN parent_id;
-
-ALTER TABLE order_items
-  ADD ticket_type_id Uuid NULL REFERENCES ticket_types(id);
-
-
-CREATE INDEX index_order_items_ticket_type_id ON order_items (ticket_type_id);
```
Timo614: I noticed with the change to push the ticket_type_id into the create db migration we tossed this index. Maybe worth adding back there if we're doing any lookups on the order items by the `ticket_type_id`.

db/src/models/order_items.rs
```diff
@@ -22,23 +22,27 @@ pub struct OrderItem {
...
 }
 
 impl OrderItem {
-    pub(crate) fn find(id: Uuid, conn: &PgConnection) -> Result<OrderItem, DatabaseError> {
-        order_items::table
-            .filter(order_items::id.eq(id))
-            .first(conn)
-            .to_db_error(errors::ErrorCode::QueryError, "Could not find order item")
-    }
+    //    pub(crate) fn find(id: Uuid, conn: &PgConnection) -> Result<OrderItem, DatabaseError> {
```
Timo614: May be worth just nuking this logic since it'd be pretty easy to add back if the need arises later unless we're commenting it out because we intend to add it back in some other branch but the unused code error build logic was going off here and we don't need it yet?

db/src/models/order_items.rs
```diff
@@ -215,21 +225,21 @@ impl OrderItem {
...
-            .filter(order_items::order_id.eq(order_id))
-            .filter(order_items::ticket_pricing_id.eq(ticket_pricing_id))
-            .filter(order_items::item_type.eq(OrderItemTypes::Tickets.to_string()))
-            .first(conn)
-            .to_db_error(
-                errors::ErrorCode::QueryError,
-                "Could not retrieve order item",
-            )
-    }
+    //    pub(crate) fn find_for_ticket_pricing(
```
Timo614: Just same as above where it may be worth just nuking this as we can always add back via commit history if a need arises unless it's something with an active branch and we're just commenting out to avoid the build error logic.

db/src/models/orders.rs
```diff
@@ -162,51 +163,158 @@ impl Order {
...
+            });
+        }
 
-        let order_item = match OrderItem::find_for_ticket_pricing(self.id, ticket_pricing.id, conn)
-            .optional()?
+        for mut current_line in self
+            .items(conn)?
+            .into_iter()
+            .filter(|t| t.item_type() == OrderItemTypes::Tickets)
         {
```
Timo614: It may be worth adding a validation (here or in another card) related to `max_per_order` for the hold with the db via a constraint and something on the code side so it bubbles up as a validation error.

Did a quick search but couldn't find anything (it's a big card though so may have missed it and not sure if we already had something in place but I don't think we did yet since we're adding it to the cart here so we might have been waiting for this logic to add it).

db/src/models/orders.rs
```diff
@@ -162,51 +163,158 @@ impl Order {
...
+                        // TODO: Move this to an external processer
+                        let fee_schedule_range = ticket_type
+                            .fee_schedule(conn)?
+                            .get_range(ticket_pricing.price_in_cents, conn)?;
+                        if Some(ticket_pricing.id) != current_line.ticket_pricing_id {
+                            let order_item = NewTicketsOrderItem {
+                                order_id: self.id,
+                                item_type: OrderItemTypes::Tickets.to_string(),
+                                quantity: matching_line.quantity as i64,
+
```
Timo614: May be worth nuking the extra spaces here and below. Not a huge deal / don't really mind if we keep it just seemed different compared to the other struct syntax we've historically used unless we were highlighting the `ticket_type_id` for some reason.

db/migrations/00000000000021_order_items/up.sql
```diff
@@ -2,7 +2,7 @@ CREATE TABLE order_items (
   id uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
   order_id uuid NOT NULL REFERENCES orders (id) ON DELETE CASCADE,
   item_type text not null,
-  ticket_type_id Uuid not null references ticket_types,
+  ticket_type_id Uuid null references ticket_types,
```
Timo614: I noticed we made this nullable. I'd wager it should never be null when we have an `item_type` of tickets though right? May be worth adding a constraint here to check that and to add a validation in code. We call `unwrap` during the ticket updating process so just thinking it'd be worth that db integrity check to make sure we never panic on that.

db/src/models/orders.rs
```diff
@@ -219,78 +327,48 @@ impl Order {
...
-                    new_event_fee.unit_price_in_cents = event.fee_in_cents.unwrap();
-                    new_event_fee.commit(conn)?;
-                } else if organization.event_fee_in_cents.is_some() {
-                    new_event_fee.unit_price_in_cents = organization.event_fee_in_cents.unwrap();
-                    new_event_fee.commit(conn)?;
+            let event = Event::find(event_id.unwrap(), conn)?;
+            let organization = Organization::find(event.organization_id, conn)?;
+            for o in items {
+                match o.item_type() {
+                    OrderItemTypes::Tickets => o.update_fees(conn)?,
```
Timo614: Are we saving the pricing aspect for another card? The 100% free comps and `discount_in_cents` for the hold?

I remember Zach mentioning that fees are calculated on the price after it has been adjusted for the discount so we'd probably want the ticket total price to be calculated on an item level prior to this?

db/src/models/orders.rs
```diff
@@ -162,51 +163,158 @@ impl Order {
...
+                self.expires_at,
+                new_line.ticket_type_id,
+                hold_id,
+                new_line.quantity,
+                conn,
+            )?;
+        }
 
-        order_item.update_fees(conn)?;
+        self.update_fees(conn)?;
```
Timo614: Mentioned below https://github.com/big-neon/bn-api/pull/507/files#r231199931 but we may need to calculate the new pricing prior for the hold and comp since the fees are calculated based on the end pricing for that. Unless we're saving that for another issue.

db/src/models/orders.rs
```diff
@@ -219,78 +327,48 @@ impl Order {
...
+
+            let mut new_event_fee = NewFeesOrderItem {
+                order_id: self.id,
+                item_type: OrderItemTypes::EventFees.to_string(),
+                event_id: Some(event.id),
+                unit_price_in_cents: 0,
+                quantity: 1,
+                parent_id: None,
+            };
+            if event.fee_in_cents.is_some() {
```
Timo614: More of a @zmelamed question but if the order only contains comps should the event fee just not be included? But if they include multiple ticket types some with comps some without it should?

db/src/models/orders.rs
mikethetike: Good point. Removed

db/src/models/orders.rs
mikethetike: Yeah this will have to be another card. This is already a huge PR

db/migrations/00000000000021_order_items/up.sql
mikethetike: Added the constraint

db/src/models/orders.rs
mikethetike: Good point. I'll need to check that - New ticket

Issue #448, opened 2018-10-22
Create access codes


Krakaw:
See big-neon/docs#19

Issue #509, opened 2018-11-02
Apply discount codes and purchase


Timo614:
https://github.com/big-neon/bn-api/issues/446#issuecomment-434578054

#501 takes care of the creation and management of discount codes but we'll need to also add logic to use them for the cart and displaying of prices.

Needed validations:
- Order is within discount window during purchasing
- Ticket association with discount code to check max per user
- Max uses does not exceed discount max uses via association lookup (constraint and validation)

Timo614:
Going to hold off actually as we have #447 open so not sure if the current discount code implementation is still what we intend to go with?

PR #518, opened 2018-11-07
WIP: Access codes

Timo614:
Closes: #448

This PR attempts to address the associated issue by adding a new `CodeTypes::Access` type for use with the API. In addition several smaller changes were made in the branch.

I originally intended to look into adding the ticket type pricing and add to cart logic here but #509 will take care of most of that and it's waiting on #507 so figured it was better to just get this out as is and we can iterate on it. In addition my schedule for the next week is going to be erratic so don't want it holding things up. I'll make a note in #509 to also add the gating functionality of the `Access` type when we do the ticket types pricing logic changes.

Todo:
- Still need to spend a little more time with the branch to make sure nothing is amiss.
- Going to add pagination support to the codes endpoint (will make a new issue for the events and other endpoints currently just returning the pagination struct without doing anything on the db side to support it)
- Make a note about adding access gating to #509
- Confirm the db validations added won't cause any issues (I can't add tests at the moment so can't fully verify but will once over them again and then in #509 we can add some tests to further validate it)

Changes:
- Adds support to make `discount_in_cents` nullable adjusts validations around that
- Moved the `code_id` off of the `orders` and `ticket_instances` tables in favor of tracking via the `order_items` as the add to cart logic coming in #507 used that table to track redemption code associations with the order. I was able to pull together what I needed for the validations based on that new field (as we can just group by order and count those to check against `max_uses` as done in the validation).
- Fixed codes endpoint to take `code_type` and mimic events index setup
- Adds cascading deletions to `comps` if the associated `holds` row is deleted
- Clears Comps when Hold type is changed (there's a todo there to prevent changing the type when Comp already used so that aspect wasn't handled here)
- Cleans up the validation logic a little bit to fit a standard format across the models

Questions:
- Are we going to ever hit the codes endpoint without a code type? Or was that done just temporarily? Wondering if I can nuke the option there or if we'll have a reason to pull all the discount and access codes at once (perhaps some export? although I feel we'd want more control of the output then).
- Is this new format correct for the codes search parameters? Not sure how the tags are supposed to work so mimicked the events endpoint. Are the tags supposed to be how we transmit that data or just additional data returned about how the query worked?

Validations:
- Check quantity of order items against code max_tickets_per_user
- Check number of orders against code max uses
- Confirms discount is present for `CodeTypes::Discount` but not required (but still usable) for `CodeTypes::Access`

Notes:
- I didn't add any order item creation validation tests as there's no way to create one with a code at the moment. That will be possible with #509, however.
- The codes logic doesn't currently support the paging parameters but neither does most of the code (it seems like that feature is mostly just for showing the correct format for now? Nothing I could find seems to be hooked up to use it). Will be at least adding support for this as mentioned in the todo above.
- Tried to switch the query parameter logic to use serde flatten but it appears it fails with actix-web as it treats them as strings so going to hold off on that for now. Will file an issue with actix-web later about it once I have a chance to make a concise test case.

https://serde.rs/attr-flatten.html
```
#[serde(flatten)]
paging_parameters: PagingParameters,

curl "http://localhost:9000/events?page=2" -i
HTTP/1.1 400 Bad Request
{ "level": "WARN", "target": "actix_web::pipeline", "message": "Error occured during request handling: invalid type: string "2", expected u32" }
{ "level": "DEBUG", "target": "actix_web::pipeline", "message": "Error { err: "invalid type: string \"2\", expected u32" }" }
```


Timo614:
I spent awhile looking into changing the syntax to match your ideal format and I don't think we'll be able to use that without some upstream changes to support it. `/events/{id}/codes?tags[type]=discount`

The problem is [here](https://github.com/nox/serde_urlencoded/blob/4d220818f758a0797f754c8ca33b117ee981d5e5/src/de.rs#L155) the upstream serde_urlencoded crate (used by the Query extractor in actix-web) treats multi dimensional keys as if they are part of the same key name for serde's `MapDeserializer` logic.

The easiest way to see this in action is by sending a different type than what the parameter is expecting. If serde thought it found that key it would complain about the type. It doesn't however if you try to send anything to `?tags[type]=` despite many of the endpoints taking that field. 

So if I then rename the tags element of the `PagingParameters` struct to 
`#[serde(rename = "tags[type]")]`

and I send the same value it will error out as it's expecting a `HashMap`:
`Error { err: "invalid type: string \"Discount\", expected a map" }`

It just treats these complex keys as if they are individual field names and isn't aware of any complex array structures associated with them.

We may be able to get around that with a custom deserializer that handles the keys differently for the struct in question. Will be a larger change though as I'm not really that familiar with serde so leaving this as is for now. Also not sure if it'll play nice with how the QueryExtractor works as it does that all internally to actix-web so perhaps we won't have a say in the matter. We could make an issue upstream for serde_urlencoded, however.

For now in my changes I'm using the serde `flatten` logic to pass along the additional elements:
https://serde.rs/field-attrs.html#serdeflatten

So you'd specify type as an additional field along with the paging parameters and it gets bundled into the tags array automatically.

Accessing the tag in code:
```
    let mut code_type: Option<CodeTypes> = None;
    if let Some(value) = query.tags.get("type") {
        code_type = serde_json::from_value(value.clone())?;
    }
```

No type provided:
```
curl "http://localhost:9000/events/b9498b3d-6530-43cc-825a-d24f0365eae7/codes" --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI4ZjU3Y2I5NC00MGMxLTRiYTQtOGNiNy1mM2VjYzAyZDZkYjUiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQxNjMyNTIzfQ.29n0I5wf3MrtPvW65jXEWwj4tLq2RKdRp+HtpMKKxrc" -i

{"data":[{"id":"04fb57e7-c180-455c-8ea7-95e1d75a57f0","name":"Access Tickets","event_id":"b9498b3d-6530-43cc-825a-d24f0365eae7","code_type":"Access","redemption_code":"AccessDiscountCode1541631494","max_uses":10,"discount_in_cents":null,"start_date":"2018-01-01T12:00:00","end_date":"2019-01-01T12:00:00","max_tickets_per_user":10,"created_at":"2018-11-07T22:58:14.433295","updated_at":"2018-11-07T22:58:14.433295","ticket_type_ids":["3d7f106c-c1f7-4620-9901-3c5bdcb27d1b"]}],"paging":{"page":0,"limit":1,"sort":"","dir":"Asc","total":1,"tags":{}}}
```

Discount provided (no codes on file):
```
curl "http://localhost:9000/events/b9498b3d-6530-43cc-825a-d24f0365eae7/codes?type=Discount" --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI4ZjU3Y2I5NC00MGMxLTRiYTQtOGNiNy1mM2VjYzAyZDZkYjUiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQxNjMyNTIzfQ.29n0I5wf3MrtPvW65jXEWwj4tLq2RKdRp+HtpMKKxrc" -i

{"data":[],"paging":{"page":0,"limit":0,"sort":"","dir":"Asc","total":0,"tags":{"type":"Discount"}}}
```

Access provided (one code on file):
```
curl "http://localhost:9000/events/b9498b3d-6530-43cc-825a-d24f0365eae7/codes?type=Access" --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJraWQiOm51bGwsImFsZyI6IkhTMjU2In0.eyJzdWIiOiI4ZjU3Y2I5NC00MGMxLTRiYTQtOGNiNy1mM2VjYzAyZDZkYjUiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQxNjMyNTIzfQ.29n0I5wf3MrtPvW65jXEWwj4tLq2RKdRp+HtpMKKxrc" -i

{"data":[{"id":"04fb57e7-c180-455c-8ea7-95e1d75a57f0","name":"Access Tickets","event_id":"b9498b3d-6530-43cc-825a-d24f0365eae7","code_type":"Access","redemption_code":"AccessDiscountCode1541631494","max_uses":10,"discount_in_cents":null,"start_date":"2018-01-01T12:00:00","end_date":"2019-01-01T12:00:00","max_tickets_per_user":10,"created_at":"2018-11-07T22:58:14.433295","updated_at":"2018-11-07T22:58:14.433295","ticket_type_ids":["3d7f106c-c1f7-4620-9901-3c5bdcb27d1b"]}],"paging":{"page":0,"limit":1,"sort":"","dir":"Asc","total":1,"tags":{"type":"Access"}}}
```

Note:
- I didn't change the events index to use the tag like setup (figure we can do this in whatever card we'll make eventually to make the pagination work).

In addition I adjusted some logic in this branch further with my latest commit:
- Made the enum parsing a custom error type (originally was parsing a string back from it but after I changed the Paging logic kept it in place as it's the better choice here regardless).
- Adds redemption code uniqueness across all models check for comps as well. One oddity there is the mockups don't show a way to set it. We may want to do a follow up issue where we retry and generate a new code if it ends up conflicting. Right now I've just adjusted the SQL function to take the model into account.

We can spend more time with the new validations once we add the ability to associate the codes with the cart. Just was limited given the permission model from further adding testing for the validations added here.

Krakaw:
That makes sense, since the tags functionality hasn't really been implemented anywhere and using it is going to be a massive pain, I'm going to create a new issue to start a discussion for a replacement in #526

db/src/models/comps.rs
```diff
@@ -211,6 +210,15 @@ impl Comp {
             diesel::delete(self).execute(conn),
         )
     }
+
+    pub fn destroy_from_hold(hold_id: Uuid, conn: &PgConnection) -> Result<usize, DatabaseError> {
+        // TODO: prevent deletion of comps that have been claimed
```
mikethetike: I think we should make a ticket for this so we don't forget

Timo614:
> I see the endpoint takes in a ticket_type_id and not a pricing. Does this mean that the current pricing will be used?

@mikethetike For which endpoint? The codes creation? Yeah the requirements mentioned associating these with ticket types themselves so the discount would come out of the current pricing.

PR #531, opened 2018-11-09
Update hold quantity

mikethetike:
- Removed unused route
- Update quantity in update-hold method
- return 422 Validation Error instead of 500


api/tests/functional/holds.rs
```diff
@@ -158,47 +137,48 @@ fn update_with_validation_errors() {
...
-
-    let expected_json = json!({
-        "error": "Validation error",
-        "fields":{
-            "quantity":[{"code":"assigned_comp_count_greater_than_quantity","message":null,"params":{}}],
-        }
-    }).to_string();
-    assert_eq!(body, expected_json);
-}
+// TODO redo this test
```
SWvheerden: I am going to assume this test is broken, but functionality is not?
Perhaps add a reason why a redo is needed and this is commented out, as well as the date this was done

db/src/utils/errors.rs
```diff
@@ -183,6 +183,20 @@ impl DatabaseError {
         ))
     }
 
+    pub fn validation_error<T>(
+        field: &'static str,
```
SWvheerden: is there a reason the lifetime should be static?

api/tests/functional/holds.rs
mikethetike: I think this is fixed in a different PR

Issue #511, opened 2018-11-05
Domain Messages


Krakaw:
Domain messages are a pub/sub messaging system that will take triggers from the domain events table and action them. These triggers can be time based or event based.

The domain_messages table will require the following fields

```
id - uuid
domain_event_id - if this was triggered by a domain event
domain_message_type - Enum of the message types we will be sending
communication_channel_type - Enum of the available communication channel
main_table - what table will the main_table_id relate to
main_table_id - The id of the relevant entity
scheduled_at - Timestamp to run
expires_at - Timestamp to expire
last_attempted_at - The last time we attempted to run this domain message type
attempt_count - The number of attempts
max_attempt_count - The maximum number of attempts
status - Enum of status
blocked_until - Timestamp of the next available run time for a retry
created_at - Timestamp
updated_at - Timestamp
```

The domain message triggers will be cron based as well as event driven. This allows us to attempt to send a domain message immediately, such as in the case of a user registration, as well as triggering scheduled messages, such as the doors to the event are opening.

PR #537, opened 2018-11-09
Adding async Domain Events and Domain Action queuing and monitoring system

### LINKED ISSUE #512: Domain Message Handler Categories
The domain message type handler will contain the category type embedded in the trait / struct. This will compare itself to the users' notification settings to decide whether it should be sent to that user or not.

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/511
Closes #512
 Still a major WIP.

The goal of this PR is to implement the mechanism to be able to register a Domain Event (i.e. Issue a Ticket hold for a limited time promotion OR A user registering) and that Event will create one or more Domain Actions that will be executed, asynchronously, either immediately or at some time in the future (e.g. After the promotion is completed return the remaining tickets from the hold to the general ticket pool OR immediately send a welcome email). The Domain Action Monitor will be a system running on a separate threadpool that will periodically check if there are Pending Actions that are due and if there are spawn an asynchronous attempt to complete the Action,.

The database model was written and the start of the controllers for both Events and Actions was started. The main challenge is figuring out how to run the Domain Action Monitor asynchronously in a threadpool separate from the Actix workers but keep the code modular. I still haven't figured this out. The next challenge will be to find an easy and modular way to launch asynchronous attempts at executing Actions which will take the form of Futures spawned onto the threadpool. It will be easier to Actions that involve Reqwest requests because they are inherently Futures and handle the task::current().notify() stuff in the Futures polling methods. Writing our own Futures will involve managing the notify action ourselves. I have included a rough example of each but still much to figure out on that front.


api/src/events/domain_actions.rs
```diff
@@ -79,7 +78,7 @@ impl DomainActionMonitor {
     // }
 }
 
-//This is a Reqwest task so Reqwest will handle the tokio::task::current().notify() stuff.
+//This is a Request task so Request will handle the tokio::task::current().notify() stuff.
```
mikethetike: I think this was intentionally spelled 'Reqwest', referring to the library

philipr-za: Correct, we are the Reqwest library for all our external API calls (check out the Sendgrid controller in bn-api).

api/src/events/domain_actions.rs
```diff
@@ -13,62 +14,66 @@ use std::io::{self, Cursor};
...
 pub struct DomainActionMonitor {
-    database_url: String,
+    config: Arc<Config>,
     pub interval: u64,
     runtime: Runtime,
 }
 
 impl DomainActionMonitor {
-    pub fn new(database_url: String, poll_period_in_secs: u64) -> DomainActionMonitor {
+    pub fn new(conf: Arc<Config>, poll_period_in_secs: u64) -> DomainActionMonitor {
```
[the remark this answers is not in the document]
SWvheerden: yes, I think so. The threads are moved, this we need a reference counter.

api/src/events/domain_actions.rs
```diff
@@ -0,0 +1,134 @@
...
+        let mut runtime = Runtime::new().unwrap();
+        loop {
+            //Check for actions that are due to be processed
+            let mut pending_actions = match DomainAction::find_pending(None, &connection) {
+                Ok(val) => val,
+                Err(_) => break,
+            };
+            // //Process actions
+            for action in pending_actions {
+                println!("Pending Action: {}", action.domain_action_type);
```
philipr-za: This should probably be a jlog! action rather than a println!

db/src/models/domain_actions.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn set_done(&self, conn: &PgConnection) -> Result<DomainAction, DatabaseError> {
+        diesel::update(self)
+            .set((
+                domain_actions::status.eq(DomainActionStatus::Success.to_string()),
+                domain_actions::updated_at.eq(dsl::now),
+            )).get_result(conn)
+            .to_db_error(ErrorCode::UpdateError, "Could not update Domain Action")
+    }
+
+    pub fn Inc_fail_count(&self, conn: &PgConnection) -> Result<DomainAction, DatabaseError> {
```
mikethetike: `inc_fail_count`

api/src/events/domain_actions.rs
```diff
@@ -0,0 +1,134 @@
...
+                        Duration::from_secs(1800),
+                    );
+                    runtime.spawn(timeout.map_err(|_| {
+                        println!("Pending Action:  failed");
+                    }));
+                }
+                _ => log!{Error, "Domain Action Monitor has a task without a corresponding match"},
+            };
+            }
+            thread::sleep(Duration::from_secs(10))
```
philipr-za: This should be the the structs `interval` value

api/src/events/domain_actions.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    pub fn start(&mut self) {
+        let config = self.config.clone();
+        thread::spawn(move || DomainActionMonitor::poll_database(config.clone()));
+    }
+}
+
+//This is a Reqwest task so Reqwest will handle the tokio::task::current().notify() stuff.
+//This must be spawned on the Domain Action Handler runtime.
+fn lost_password_handler() -> impl Future<Item = (), Error = ()> {
```
philipr-za: These were just examples of implementing a one-shot future. If they are not being used they should come out.

api/src/events/domain_actions.rs
```diff
@@ -0,0 +1,134 @@
...
+            let mut body = Cursor::new(body);
+            let _ = io::copy(&mut body, &mut io::stdout()).map_err(|err| {
+                println!("stdout error: {}", err);
+            });
+        })
+}
+
+//An example future that handles its own poll() and task::current().notify()
+//this toy future spins for the specified time and then returns.
+struct PurchaseCompletedFuture {
```
philipr-za: Again, take this out.

api/src/events/domain_actions.rs
```diff
@@ -0,0 +1,134 @@
...
+    done_time: Duration,
+    start_time: Instant,
+}
+
+impl Future for PurchaseCompletedFuture {
```
philipr-za: Take out this future example

api/src/events/domain_actions.rs
```diff
@@ -0,0 +1,134 @@
...
+            println!("Future has completed {:?}", self.done_time);
+            return Ok(Async::Ready(()));
+        }
+
+        task::current().notify();
+        Ok(Async::NotReady)
+    }
+}
+
+//This handler will need to spawn the future on a Tokio runtime.
```
philipr-za: Take out this future example

api/src/events/domain_actions.rs
```diff
@@ -0,0 +1,134 @@
...
+//How we get this runtime in here is part of what we still need to figure out.
+fn purchase_completed_handler(mut runtime: Runtime) {
```
philipr-za: Take out this future example.

db/scripts/init-test.sh
```diff
@@ -5,6 +5,6 @@ source .env
 set +o allexport
 # Run this script to update the database to the latest migration version
 diesel -V 2> /dev/null || cargo install diesel_cli --no-default-features --features postgres
-diesel database reset --database-url=$TEST_DATABASE_ADMIN_URL
-diesel setup --database-url=$TEST_DATABASE_ADMIN_URL
-diesel migration run --database-url=$TEST_DATABASE_ADMIN_URL
+diesel database reset --database-url=$DATABASE_URL
```
philipr-za: The other script in this directory `init-rest.sh` already has this.

db/src/models/domain_actions.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn find(id: Uuid, conn: &PgConnection) -> Result<DomainAction, DatabaseError> {
+        DatabaseError::wrap(
+            ErrorCode::QueryError,
+            "Error loading domain action",
+            domain_actions::table.find(id).first::<DomainAction>(conn),
+        )
+    }
+
+    pub fn set_busy(&self, conn: &PgConnection) -> Result<bool, DatabaseError> {
+        let timeout = Utc::now().naive_utc() + Duration::seconds(DOMAIN_ACTION_TIMEOUT_SEC);
```
philipr-za: This works but I actually think this could be a single update db call. Update the DB and if the conditions for setting busy are right the number of returned rows will be 1 and if the conditions were false you will get 0 rows updated.

db/src/models/domain_actions.rs
philipr-za: *inc_fail_count

db/src/models/domain_actions.rs
```diff
@@ -0,0 +1,179 @@
...
+use diesel;
+use diesel::expression::dsl;
+use diesel::prelude::*;
+use models::enums::*;
+use schema::*;
+use serde_json;
+use utils::errors::*;
+use uuid::Uuid;
+
+const DOMAIN_ACTION_TIMEOUT_SEC: i64 = 1800; //30 mins
```
Krakaw: Lets move this into the individual actions as each one will require a different timeout

mikethetike:
This should be ready for review. Open to any feedback

api/src/utils/communication.rs
```diff
@@ -84,63 +97,85 @@ impl Communication {
...
+                                )
                             }
-                            CommunicationType::Sms => Err(ApplicationError::new(
-                                "SMS communication not implemented".to_string(),
-                            ).into()),
-                            CommunicationType::PushNotification => Err(ApplicationError::new(
-                                "Push notifications not implemented".to_string(),
-                            ).into()),
-                        }
+                            _ => panic!("Not implemented"),
```
Timo614: May be worth returning an ApplicationError in place of the panic since this returns a BigNeonError anyhow. I'd think this could kill the DomainActionMonitor thread otherwise.

api/src/domain_events/routing.rs
```diff
@@ -0,0 +1,42 @@
...
+        match self.routes.insert(action_type, executor) {
+            Some(_) => Err(DomainActionError::Simple(
+                "Action type already has an executor".to_string(),
+            )),
+            _ => Ok(()),
+        }
+    }
+
+    pub fn get_executor_for(&self, action_type: DomainActionTypes) -> &dyn DomainActionExecutor {
+        self.routes.get(&action_type).unwrap().as_ref()
```
Timo614: May be worth having this return an error if the get call returns `None` here. Not sure how likely of a scenario that is as I feel like we'd add the executor as part of any pull request adding a new DomainActionTypes value just noticed the unwrap.

api/src/domain_events/routing.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+pub fn set_up_executors(router: &mut DomainActionRouter, conf: &Config) {
+    router
+        .add_executor(
+            DomainActionTypes::Communication,
+            Box::new(SendCommunicationExecutor::new(conf.clone())),
+        ).unwrap();
```
Timo614: May be worth propagating the error up here in case the add_executor fails (although I guess it's not the end of the world if it exists already).

api/src/domain_events/send_communication.rs
```diff
@@ -0,0 +1,25 @@
...
+impl SendCommunicationExecutor {
+    pub fn new(config: Config) -> SendCommunicationExecutor {
+        SendCommunicationExecutor { config }
+    }
+}
+
+impl DomainActionExecutor for SendCommunicationExecutor {
+    fn execute(&self, action: DomainAction, conn: PgConnection) -> ExecutorFuture {
+        let future = Communication::send_async(&action, &self.config);
+        let e = ExecutorFuture::new(action, conn, Box::new(future));
```
Timo614: Not a huge deal but may be worth just putting `ExecutorFuture::new(action, conn, Box::new(future))` on this line if we're just returning the executor right away anyhow.

api/src/utils/sendgrid.rs
```diff
@@ -35,6 +38,32 @@ pub fn send_email(
...
+    sg_message.from = SGEmail::from(&source_email_address);
+
+    let mut msg_personalization = SGPersonalization::new();
+    for email_address in dest_email_addresses {
+        msg_personalization.to.push(SGEmail::from(email_address));
+    }
+    sg_message.personalizations.push(msg_personalization);
+
+    let mut msg_content = SGContent::new();
+    if body.is_some() {
```
Timo614: Not a huge deal but could use `if let Some(body) = body {` to avoid the unwrap.

api/src/utils/communication.rs
```diff
@@ -84,63 +97,85 @@ impl Communication {
...
-                                    ).into())
-                                }
+                                let source_address =
+                                    communication.source.as_ref().unwrap().get_first().unwrap();
+
+                                sendgrid::send_email_template_async(
+                                    &config.sendgrid_api_key.clone(),
+                                    &source_address,
+                                    &destination_addresses,
+                                    &communication.template_id.clone().unwrap(),
```
Timo614: Noticed we nuked the checks above to make sure these had values. May be worth adding them back or having a db constraint to make sure we never create communications missing them to avoid a chance to panic here.

api/src/utils/communication.rs
```diff
@@ -84,63 +97,85 @@ impl Communication {
...
-                                        &self.title,
-                                        &self.body,
-                                    )
-                                } else {
-                                    Err(ApplicationError::new(
-                                        "Email source not specified".to_string(),
-                                    ).into())
-                                }
+                                let source_address =
+                                    communication.source.as_ref().unwrap().get_first().unwrap();
```
Timo614: Same thing as below here regarding the check that source had a value.

db/src/models/domain_actions.rs
```diff
@@ -0,0 +1,188 @@
...
+            attempt_count: 0,
+            max_attempt_count,
+            status: DomainActionStatus::Pending.to_string(),
+        }
+    }
+
+    pub fn domain_action_type(&self) -> DomainActionTypes {
+        self.domain_action_type
+            .parse::<DomainActionTypes>()
+            .unwrap()
```
Timo614: May be worth handling via a result to avoid a panic here if it ends up failing parsing for whatever reason. This is called via DomainActionMonitor I think probably could just have some From trait logic to convert it to a DomainActionError there.

api/src/utils/communication.rs
mikethetike: It's probably better to remove the other CommunicationType enum values until we actually use them. Then when they are added it'll generate a compile time error

api/src/domain_events/routing.rs
mikethetike: I've updated it to generate a compile error (hopefully), but otherwise to crash at a higher level, which will be recorded on the database row. The processor will keep running

api/src/domain_events/send_communication.rs
mikethetike: 👍

api/src/utils/sendgrid.rs
mikethetike: 👍

mikethetike: Cleaned up some of the borrowing/moving in this file as well

api/src/utils/communication.rs
mikethetike: A better way to do this is to change CommunicationType into an enum with the fields we need. That's a bit of extra work, so I'll create a ticket and then come bback to it

mikethetike: Created #588

api/src/utils/communication.rs
mikethetike: See other comment

mikethetike:
The clap crate was to allow us to pass in some command line arguments to run either the HTTP API server or the Domain Action manager, so that we can split up the instances when it is hosted in production. It should also make debugging either component easier. This PR was just getting so big though, so it will have to be done in another issue

mikethetike:
Created #589 for allowing the api server to run either the API or the domain actions

Timo614:
Saw we rolled back the crate updates so took a quick look.

Looks like the actix-web in master sets QueryConfig for the Query extractor as public (private in the latest crate version) so once they do another crate release we should be able to update actix-web and the other crates without running into test issues (aside from the initial work to modify our extractor logic in the tests to use that in place of the tuple like setup they had previously).

mikethetike:
@Timo614 Yeah, I've been struggling with the postman tests. Basically, whenever there is an HTTP call, the whole process shuts down with a Seg fault. I initially hoped it was one of the tokio/actix/future libs that we were using and that we could update it, but it looks like it's actually got something to do with running two tokio runtimes. I'm still digging through it.

PR #546, opened 2018-11-12
Validation max per order

### LINKED ISSUE #503: Set max tickets per customer on ticket types
A user must be limited by a field set by an org owner for a ticket type.
Something like: `max_tickets_per_customer`.

Also needs to be validated when adding to cart.

### LINKED ISSUE #521: Add validation for max per order/max per hold

Krakaw:
Closes #503 
Closes #521


api/src/controllers/ticket_types.rs
```diff
@@ -27,6 +27,7 @@ pub struct CreateTicketTypeRequest {
     pub end_date: NaiveDateTime,
     pub ticket_pricing: Vec<CreateTicketPricingRequest>,
     pub increment: Option<i32>,
+    pub limit_per_person: i32,
```
mikethetike: How do you specify no limit?

db/src/models/orders.rs
```diff
@@ -304,6 +328,27 @@ impl Order {
...
+            let quantities_ordered =
+                Order::quantity_for_user_for_event(&self.user_id, &limit_check.event_id, &conn)?;
+
+            if &limit_check.limit_per_person > &0
+                && quantities_ordered.contains_key(&limit_check.ticket_type_id)
+            {
+                match quantities_ordered.get(&limit_check.ticket_type_id) {
+                    Some(ordered_quantity) => {
+                        if ordered_quantity > &limit_check.limit_per_person {
+                            return Err(DatabaseError::new(
```
mikethetike: It might be better to use an ErrorCode::ValidationError here. It'll make the frontend easier to manage

db/src/models/orders.rs
```diff
@@ -366,6 +411,38 @@ impl Order {
         Ok(())
     }
 
+    pub fn available_order_limit_for_user(
```
mikethetike: is this intended?

db/src/models/orders.rs
```diff
@@ -304,6 +328,27 @@ impl Order {
             )?;
         }
 
+        for limit_check in check_ticket_limits {
+            let quantities_ordered =
+                Order::quantity_for_user_for_event(&self.user_id, &limit_check.event_id, &conn)?;
```
mikethetike: Is the limit per event or per ticket type?

db/src/queries/quantity_of_tickets_per_user_per_event.sql
```diff
@@ -0,0 +1,17 @@
+SELECT
+    order_items.ticket_type_id,
+    CAST(COUNT(ticket_instances.id) AS Integer) as total_quantity
+FROM
+    orders
+LEFT JOIN
+    order_items ON (order_items.order_id = orders.id)
+RIGHT JOIN
```
mikethetike: I think this can be an inner join. I'm not sure whether the right join might return unexpected results

db/src/models/orders.rs
Krakaw: Forgotten code.

db/src/models/orders.rs
Krakaw: Per event, per ticket type

mikethetike: Ah I see that now. Silly me

db/src/queries/quantity_of_tickets_per_user_per_event.sql
Krakaw: Made it an inner join

PR #550, opened 2018-11-13
Box office only ticket_pricing

### LINKED ISSUE #504: Price at door for ticket types
The designs have a `price at door` inout when creating ticket types. Not sure what the plan to accommodate this is? Set per ticket type or an additional ticket_pricing?

Krakaw:
Closes #504 
Added a `is_box_office_only` field to ticket_pricing, if this is true then it allows an overlap of dates in the validation. It however is not available for checkout through the normal cart flow.
TODO: Add a box office endpoint that can use the ticket_pricing that has `is_box_office_only` set to true and add items to the cart.


api/src/controllers/ticket_types.rs
```diff
@@ -16,6 +16,7 @@ pub struct CreateTicketPricingRequest {
     pub price_in_cents: i64,
     pub start_date: NaiveDateTime,
     pub end_date: NaiveDateTime,
+    pub is_box_office_only: Option<bool>,
```
mikethetike: Is this not just true or false? When would this value be null?

Krakaw: Backwards compatibility, if the client isn't sending `is_box_office_only` it will assume it is false.
We can remove this though and just update the clients?

PR #552, opened 2018-11-14
Add video url to events

### LINKED ISSUE #475: Add video_url to events
Designs show a promo video is needed on event create/update

sdbondi:
Closes #475 

Added video_url to events and in the `event.create` method.

Extras:
Added `--seed` flag to `init-reset.sh` which, if given, will run `cargo run seed -c $DATABASE_URL`


db/src/models/events.rs
```diff
@@ -65,6 +66,7 @@ pub struct NewEvent {
         message = "Top line info must be at most 100 characters long"
     ))]
     pub top_line_info: Option<String>,
+    pub video_url: Option<String>,
```
philipr-za: Add the URL validation derivation as per promo_image_url above

db/src/models/events.rs
```diff
@@ -109,6 +111,7 @@ pub struct EventEditableAttributes {
...
```
philipr-za: Add the URL validation here too.

db/src/models/events.rs
```diff
@@ -120,6 +123,7 @@ impl Event {
         event_start: Option<NaiveDateTime>,
         door_time: Option<NaiveDateTime>,
         publish_date: Option<NaiveDateTime>,
+        video_url: Option<String>,
```
philipr-za: I dont think this should be here. The other optionals like promo_image_url, additional_info, age_limit etc are not included. I think the flow from the front end is to create it with minimal stuff and then update the remaining stuff.

sdbondi: Ah makes sense, seemed funny to me too :)

Issue #483, opened 2018-10-30
Audit unwrap usage to prevent panics / properly handle errors


Timo614:
Cayle mentioned he saw some panics in the logs. Those are likely to be caused by the logic unwrapping an error result or none value. I noticed in some cases we are unwrapping logic instead of just returning an error which may be bubbling up.

`curl http://localhost:9000/users/me --header "Authorization: Bearer" -i` panics as the logic assumes there will always be a token there after the bearer.

A search of `.unwrap` constrained to `bn-api/api/src` brings up some results so we should look into those. Some seem legit but a few of them look like they could cause issues. We should also confirm whether the db has any issues like this as I noticed quite a few unwraps in that logic as well.

There's also a `panic!` called from inside of `create_payment_processor` which could cause the cart to panic if an invalid method is sent as well.

PR #554, opened 2018-11-15
Audit unwrap usage to prevent panics

Timo614:
Closes: #483

This branch attempts to address the issues mentioned in the associated issue by removing a lot of use cases of unwrap in favor of explicitly matching the results or by bubbling up the errors via result return types where they previously just panicked. 

Main Changes:
- Removed unwraps in favor of safer code where possible
- Switched the JWT library to a more recent crate with more activity from contributors (the other was years old without much work on it) as the existing one's error support was pretty dated. I switched the algorithm to use HS256 instead of RS256 as the latter for this crate requires converting the key to a DER encoded key. It should be easy enough to do in the future if we feel strongly about using RS256 but didn't want to interfere with the current workflow of things by requiring that.
- Switched the error handling code over to macros to reduce duplication there.

Notes:
- Left test code alone (only searched in `/db/src` and `/api/src`) as it didn't seem worth adjusting those given the panics would happen in the process of a single test case which would then fail
- Left server related unwraps alone as those would trigger during initial run if at all
- Left unwraps from configuration setup as those would panic and fail immediately during start
- Didn't touch any uses of `unwrap_or` as they won't panic
- Left time difference unwraps in place as they are checking the difference from unix epoch
- There's some events code to avoid an n+1 query (by fetching the venues all at once to be reused across events returned) so left that in place as it filters to the `is_some()` ones first
- Didn't touch the threading code in organization_invites as my local machine isn't able to connect to my SMTP host at the moment so couldn't test it
- update_quantities has a few unwraps in it but I believe we have a constraint to force ticket cart items to have the ticket_type_id so should be fine


api/src/errors/big_neon_error.rs
```diff
@@ -4,76 +4,45 @@ use bigneon_db::utils::errors::*;
...
-        BigNeonError(Box::new(e))
-    }
-}
-
-impl From<AuthError> for BigNeonError {
-    fn from(e: AuthError) -> Self {
-        BigNeonError(Box::new(e))
-    }
-}
+macro_rules! error_conversion {
```
mikethetike: Nice!

Issue #459, opened 2018-10-25
Event should not be able to be published if it has no ticket types


mikethetike:

Timo614:
Just confirming: So for this we're just checking if the ticket type exists right? Not if there's existing pricing for it yet?

Timo614:
Closing, was mentioned in #562 that this functionality may not be needed.

Will reopen if we decide later it's worth having and can rebase the associated PR.

PR #562, opened 2018-11-15
Prevent event publishing if no ticket types exist

Timo614:
Closes: #459

This PR deals with the associated issue by adding another validation to the events publish logic. I adjusted the logic there so that the venues publishing results could have errors appended to it prior to returning so they'd all be present in the same response.

In addition added some test logic for it to the unit tests and adjusted the postman logic to take it into account (moving things around so it'd pass after the ticket types were added and adding an assertion).

I added an open question in the associated issue: not sure if we should be only allowing publishing if ticket pricing exists for these ticket types? Or is it fine just that they exist? Not sure if we'd ever have a hold or something for a ticket type without pricing for comps?

Can quickly add an assertion to make sure ticket pricing exists if that's the desired behavior.


zmelamed:
@Timo614 what about the case where an event is externally ticketed? That should technically be listed as published with no ticket types associated with it as it'll be ticketed outside of our system.

Timo614:
@zmelamed how are those currently represented? Just as an event that's normally added and published but doesn't have a ticket type added for it?

If so we may want to just nix the issue and this PR. 

We could add some new attribute to designate between normal events and externally ticketed but if the only difference is that this type doesn't have ticket types it seems like a better call to just let any get published and if a ticket type is forgotten someone can add it after the fact. Not sure the best call there.

zmelamed:
@Timo614 Yeah I don't really fully see an issue with publishing an event without tickets. I could see a case where a show is announced before ticket inventories are decided on. With that said I had thought there was an is_external and external_url attribute on the event interfaces as there are those options in the event creation views, I'm not seeing it in the docs now though so I may be mistaken.

Timo614:
Going to close PR but will keep branch alive in case we want to resurrect in the future.

Will close the associated issue as well.

Issue #566, opened 2018-11-16
Add the logging of roles attempting to access endpoints outside of their scopes


Krakaw:
If a user attempts to access an endpoint that is not defined by their scope specifically log their details.
```
User ID
User Name
Their scopes
Accessed Scopes
IP
URL
```

PR #570, opened 2018-11-16
Log unauthorized endpoint access

Timo614:
Closes: #566 

This branch closes the associated issue by adding some additional logging to the system when a user attempts to access a resource they do not have permissions to access.

An example (for an admin only endpoint with no organization associated) can be seen below:
```
{ "level": "WARN", "target": "logging", "message": "Unauthorized access attempt", "accessed_scope":"artist:write","global_scopes":["event:interest","order:read","ticket:transfer"],"ip_address":"127.0.0.1:51716","method":"POST","url":"/venues","user_id":"4c4c47e8-f0f2-48e4-8ec7-13b48c33816e","user_name":"Mike Surname" }
```

One for organization access posting a new venue specifying the `organization_id` (I temporarily moved the `VenueWrite` scope to the `OrgOwner` role to test the response (normally it'd be a scope you could do for that level):
```
{ "level": "WARN", "target": "logging", "message": "Unauthorized access attempt", "accessed_scope":"venue:write","global_scopes":["event:interest","order:read","ticket:transfer"],"ip_address":"127.0.0.1:51767","method":"POST","organization_id":"788c9d9d-f6eb-42e0-b9df-8a4e050b3b3d","organization_scopes":["artist:write","code:read","code:write","comp:read","comp:write","event:interest","event:view-guests","event:write","hold:read","hold:write","order:read","org:read","ticket:admin","ticket:transfer"],"url":"/venues","user_id":"4c4c47e8-f0f2-48e4-8ec7-13b48c33816e","user_name":"Mike Surname" }
```

- IP address is determined by: https://github.com/actix/actix-web/blob/3b536ee96c0118e8cf452b28f6acab6085db22a6/src/info.rs#L144
- Still looking into actix-web to see if it's possible to get the full URL returned (as it looks like it's currently returning a path for the URI method)
- When accessing a controller action that requires an organization the `organization_scopes` and `organization_id` are returned in the response as well.
- Not logging data for fail authorization (bad token, etc) as it's a normal action in the flow of user access as their temporary tokens expire (indicating to refresh)
- Opted to log the information in the auth user implementation as it's possible to pass the request logic to it during initialization which allowed me to leave the controller action signatures alone / didn't need to extract the `HttpRequest<AppState>` for every controller action.
- Standardized the response for unauthorized to use the same messaging where applicable

Changes:
- Cleaned up the authorization logic and switched the logic to use `requires_scope_for_organization` and `requires_scope` across the controller actions
- Logging basic information on refresh token refresh (as there's no user)


api/src/controllers/organizations.rs
```diff
@@ -49,7 +48,7 @@ pub struct NewOrganizationRequest {
 pub fn index(
     (connection, query_parameters, user): (Connection, Query<PagingParameters>, User),
 ) -> Result<HttpResponse, BigNeonError> {
-    if user.has_scope(Scopes::OrgAdmin, None, connection.get())? {
+    if user.global_scopes.contains(&Scopes::OrgAdmin.to_string()) {
```
mikethetike: Any reason this isn't `user.requires_scope()`?

api/src/helpers/application.rs
```diff
@@ -1,28 +1,38 @@
...
-    unauthorized_with_message("Unauthorized")
+    unauthorized_with_message("User does not have the required permissions")
 }
 
 pub fn unauthorized_with_message(message: &str) -> Result<HttpResponse, BigNeonError> {
     warn!("Unauthorized: {}", message);
-    let error: BigNeonError = AuthError {
-        reason: message.to_string(),
-    }.into();
+    let error: BigNeonError = AuthError::new(message.into()).into();
```
mikethetike: Maybe we should log here as well. Unless it gets trapped somewhere else?

api/src/controllers/organizations.rs
Timo614: With the diff it's a bit hard to see but this logic doesn't actually fail if it's not present just displays something else:
https://github.com/big-neon/bn-api/blob/2754255bc1c4c5d637a3f17bcf1b5c7a3e2a1177/api/src/controllers/organizations.rs#L53

I made has_scope private so switched it to just check if the global scope existed. I guess I could have used requires_scope and had it `if let Err(_) = requires_scope()` etc instead. Will switch that.

api/src/helpers/application.rs
Timo614: Actually we refactor this a little. In some cases I logged via their auth user record since it had additional user information for the purposes of logging while this could be called on a method without a user. Will change this to accept and optional AuthUser and log via that if applicable or with normally logging if not.

The only cases we wouldn't want to really log would be for where the user token is no longer valid due to exp (as it would add a lot of junk to the logs as that's an expected case every 15 minutes). That should still be fine though as that goes through the normal error flow for BigNeonError converting the error to a response and won't go through this method (that flow doesn't log automatically).

api/src/controllers/organizations.rs
Timo614: Modified this as per above. Uses `if user.requires_scope(Scopes::OrgAdmin).is_ok()` now instead of the custom check for the global scopes.

api/src/helpers/application.rs
Timo614: Modified this so the `application::unauthorized` method takes a `HttpRequest<AppState>` reference and an optional `AuthUser`. If the auth user exists it logs it with the additional user context otherwise just logs the request data.

Does add a little extra to the signature but since it's just an extractor it only affects the unit tests which pass in the extractor values so not a big deal in the grand scheme of things and only required for actions that return unauthorized responses (which isn't too common).

PR #579, opened 2018-11-21
Changed the spotify results to mimic artists

### LINKED ISSUE #540: Using spotify artist search as secondary search
When searching for an artist, if none is found in the database do a search for the artist using spotify. If it is found and selected store the required fields in or database for use later on

Krakaw:
Closes #540 

Removed the Enum that differentiated between an artist and a spotify result. Instead a `spotify_id` field has been added to the artists table.

Search results return a `CreateArtist` struct, if it is from spotify the `spotify_id` will be populated if it is a local result then the `id` field will be populated.

The `/spotify/{id}` endpoint has been removed instead, passing a `CreateArtist` to `POST /artists/create` if the `spotify_id` field is set then it will attempt to fetch the artist details from spotify.

### DIFF
--- a/api/src/controllers/artists.rs
+++ b/api/src/controllers/artists.rs
@@ -4,15 +4,9 @@ use bigneon_db::models::*;
 use db::Connection;
 use errors::*;
 use helpers::application;
-use models::{ExternalPathParameters, PathParameters, WebPayload};
+use models::{CreateArtist, PathParameters, WebPayload};
 use server::AppState;
 use utils::spotify::*;
-
-#[derive(Serialize, Debug)]
-pub enum ArtistOrSpotify {
-    Artist(Artist),
-    Spotify(SpotifyArtist),
-}
 
 pub fn search(
     (state, connection, query_parameters, user): (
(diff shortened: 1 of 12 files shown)


api/src/controllers/artists.rs
```diff
@@ -4,26 +4,21 @@ use bigneon_db::models::*;
...
-
 pub fn search(
     (state, connection, query_parameters, user): (
         State<AppState>,
         Connection,
         Query<PagingParameters>,
         Option<User>,
     ),
-) -> Result<WebPayload<ArtistOrSpotify>, BigNeonError> {
+) -> Result<WebPayload<CreateArtist>, BigNeonError> {
```
mikethetike: Maybe name this `CreateArtistRequest` to match the other methods that take this kind of parameter. Also prevents confusion with `NewArtist`

api/src/models/artist_spotify_or_local.rs
```diff
@@ -0,0 +1,69 @@
+use bigneon_db::models::{Artist, NewArtist};
```
mikethetike: This file should probably be called `CreateArtist` or `CreateArtistRequest`

api/src/controllers/artists.rs
Krakaw: Thanks, renamed

api/src/models/artist_spotify_or_local.rs
Krakaw: Done

PR #587, opened 2018-11-23
Fill events dashboard data with 0 sales if there was no data on that day

mikethetike:


db/src/models/events.rs
```diff
@@ -508,14 +509,28 @@ impl Event {
...
+
         let mut result = vec![];
 
-        for s in summary {
-            result.push(DayStats {
-                date: s.date,
-                sales: s.sales.unwrap_or(0),
-                tickets_sold: s.ticket_count.unwrap_or(0),
-            });
+        for s in 0..31 {
```
Krakaw: Do we always want to return exactly 30 days, what about using n as our limit here, and change the response data key to "data" instead of specifically "last_30_days"?

mikethetike: Good point

PR #590, opened 2018-11-23
Redeem takes event_id

### LINKED ISSUE #580: When posting to redeem tickets return the guest details on success
Instead of a success boolean, make the success object the same as the `GET /tickets/redeem/` result.

### LINKED ISSUE #581: Add event id to POST /tickets/redeem
A ticket should be directly linked to the event_id it is being redeemed for.
Such as a person with scanning access has 2 events in their venue but in different halls, you don't want to scan in a valid ticket for the wrong event.

Krakaw:
Closes #581 
Closes #580 

Moved the endpoint for redeeming a ticket to `/events/{id}/redeem/{ticket_instance_id}`

Added the postman tests for redemption.


api/src/controllers/tickets.rs
```diff
@@ -138,7 +148,8 @@ pub fn redeem(
                                                                                &a,
                                                                                vec![ticket.token_id as u64],
                             )?;
-                            Ok(HttpResponse::Ok().json(json!({"success": true})))
+
+                            Ok(HttpResponse::Ok().json(json!({"success": true, "data": &redeemable_ticket})))
```
mikethetike: Maybe use the HTTP status to indicate success and just have the data as the body

api/src/controllers/tickets.rs
```diff
@@ -148,7 +159,7 @@ pub fn redeem(
             },
         Err(e) => Ok(HttpResponse::new(StatusCode::INTERNAL_SERVER_ERROR)
```
mikethetike: We should really throw a better error here, doesn't need to be done in this ticket though

api/src/routing.rs
```diff
@@ -53,6 +53,8 @@ pub fn routes(app: &mut CorsBuilder<AppState>) -> App<AppState> {
         r.method(Method::DELETE).with(events::remove_interest);
     }).resource("/events/{id}/publish", |r| {
         r.method(Method::POST).with(events::publish);
+    }).resource("/events/{id}/redeem/{ticket_instance_id}", |r| {
+        r.method(Method::POST).with(tickets::redeem);
```
mikethetike: I think move the `tickets::redeem` method to the events controller as `events::redeem` otherwise the `id` parameter looks like it is a ticket id

api/src/controllers/tickets.rs
Krakaw: Good idea.

api/src/routing.rs
Krakaw: Agreed, moved.

Issue #544, opened 2018-11-12
Create event view fields for admin


Krakaw:
GET /organizations/{id}/events/{id}
@jasonvdb needs extra fields for bn-web
https://app.goabstract.com/projects/20f4bd60-ab00-11e8-b694-6fcb86ba37e4/branches/master/collections/78dec09d-bd7a-4539-9163-cad557673a91/layers/45bb0766-34b1-4400-9de3-da6372efdaeb/presentation

Jasonvdb:
Have built the layout, can just plugin the data when it's ready.

For the top chart maybe we need a separate API call that accepts a date range or just a single month to return stats for.

The dummy data I'm working with looks like this:

```
[
  { day: 1, ticketSales: 99, revenue_in_cents: 3000 },
  { day: 2, ticketSales: 120, revenue_in_cents: 4000 }
];
```

PR #597, opened 2018-11-27
Add range query to event dashboard endpoint

Timo614:
Closes: #544

This pull request adds range query parameters to the events dashboard endpoint. Previously the endpoint just returned the last 30 days regardless but with this change if the `start_utc` or `end_utc` is specified it uses those instead of the default values. It uses the start and end of dates respectfully so will include any orders from that day. The default is 29 days ago for the `start_utc` and today for the `end_utc`.

I modified the results to return `day_stats` instead of `last_30_days` as it depends on the query.

Also modified the day stats returned to match the associated issue (`date`, `revenue_in_cents`, `ticket_sales`). Opted not to add the `day` field as it can be determined by the index. Let me know if we want to add it though. Also opted not to add pagination as it's expected to be used all on one page as part of the graph.

Let me know if we want to change any of this as I'm not sure how far along the frontend has advanced with the existing implementation. The changes themselves are mostly minor but also added some tests so we'd catch any regressions there or if we need to expand on the date range logic in the future in any way.


Timo614:
@Jasonvdb this is merged just a heads up on the modifications for the endpoint.

PR #603, opened 2018-11-27
Add Replace Cart api endpoint

### LINKED ISSUE #602: Create an api method for replacing a user's current cart

mikethetike:
Closes #602 
It also adjusts the expiry time, so that if a cart is empty, then the expiry time is reset when the first item is added


db/src/models/orders.rs
```diff
@@ -325,14 +345,35 @@ impl Order {
                             current_line.update(conn)?;
                         }
                     }
+                } else {
+                    if remove_others {
```
Timo614: Not a huge deal but could just make this `} else if remove_overs {`.

db/src/models/orders.rs
```diff
@@ -189,13 +189,33 @@ impl Order {
...
 
+    pub fn extend_expiry(&mut self, conn: &PgConnection) -> Result<(), DatabaseError> {
+        let affected_rows = diesel::update(
+            orders::table.filter(
+                orders::id
+                    .eq(self.id)
+                    .and(orders::version.eq(self.version))
+                    .and(orders::expires_at.gt(Utc::now().naive_utc())),
+            ),
+        ).set(orders::expires_at.eq(Utc::now().naive_utc() + Duration::minutes(15)))
```
Timo614: Probably worth updating the 15 here and in the other uses to make it a constant so if we decide to adjust this later on it's easier to do across the board.

db/src/models/orders.rs
```diff
@@ -325,14 +345,35 @@ impl Order {
...
                 }
             }
             if let Some(index) = index_to_remove {
                 mapped.remove(index);
             }
         }
 
+        // if the cart is empty at this point, it is effectively a new cart, extend the time
+
+        if self.items(conn)?.len() == 0 {
```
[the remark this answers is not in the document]
mikethetike: Yeah, it wasn't really part of the purpose of this PR. I should have actually done it in as a whole seperate issue. I think that in #604, we should think about and handle all the cases

PR #622, opened 2018-11-30
Added redeem key to the /tickets response

### LINKED ISSUE #613: Add redeem_key to each ticket in /tickets
It would be very useful for the mobile app if redeem_key came back in the tickets index response on each ticket. We use this in generating the QR Code, and it would save us making extra API calls for each individual ticket.

Krakaw:
Closes #613


db/src/models/ticket_instances.rs
```diff
@@ -63,6 +63,7 @@ impl TicketInstance {
                 events::id,
                 events::venue_id,
                 ticket_instances::status,
+                ticket_instances::redeem_key,
```
mikethetike: In `show_redeemable_ticket` we only show this if the redeem_date > now()
```
  if ticket_data.redeem_date.is_some()
            && ticket_data.redeem_date.unwrap() > Utc::now().naive_utc()
        {
            ticket_data.redeem_key = None; //Redeem key not available yet. Should this be an error?
        }
```

Should we not do the same?

Krakaw: Is there any harm in showing the redeem key always, the ticket can only actually be redeemed once the redeem_date has passed anyway?

mikethetike: Fair enough

Krakaw: I agree with you, let's keep the display of the redeem key standard throughout the system. I've added the logic to replace the redeem_key with None based on the above snippet.

PR #623, opened 2018-11-30
Add Box office check out

### LINKED ISSUE #612: Add user/guest details to order when checking out at Box Office

mikethetike:
Closes #612


api/src/controllers/orders.rs
```diff
@@ -29,3 +31,57 @@ pub fn show(
...
+        .collect();
+    let mut results = vec![];
+    for item in order
+        .items(conn)?
+        .iter()
+        .filter(|t| t.item_type() == Ok(OrderItemTypes::Tickets))
+    {
+        if order.user_id != user.id() && order.on_behalf_of_user_id != Some(user.id()) {
+            if item.event_id.is_none()
+                || !orgs.contains(&Event::find(item.event_id.unwrap(), conn)?.id)
```
Timo614: I think this should be `.organization_id` above as right now it's checking to see if the event id is present in the orgs vector.

api/src/controllers/cart.rs
```diff
@@ -245,28 +273,48 @@ pub fn checkout(
...
 
-    let payment = order.add_external_payment(
-        reference.to_string(),
-        user.id(),
-        checkout_request.amount,
-        connection,
-    )?;
+    let mut guest: Option<DbUser> = None;
+
+    if email.is_some() {
```
Timo614: Could use the `if Some(email) = email {` syntax but not a big deal either way.

api/src/controllers/cart.rs
```diff
@@ -245,28 +273,48 @@ pub fn checkout(
...
+        guest = DbUser::find_by_email(email.as_ref().unwrap(), conn).optional()?;
+    };
+    if guest.is_none() {
+        if phone.is_some() {
+            guest = DbUser::find_by_phone(phone.as_ref().unwrap(), conn).optional()?;
+        }
+    }
+    if guest.is_none() {
```
Timo614: Could just make this an else statement for the above check for phone having some value. Not a huge deal either way though.

api/src/controllers/cart.rs
```diff
@@ -245,28 +273,48 @@ pub fn checkout(
...
+        guest = Some(DbUser::create_stub(
+            first_name, last_name, email, phone, conn,
```
Timo614: More of a question for @zmelamed but how are terms and conditions and privacy policy acceptance going to be accepted by the stubbed users? Do we need to add a flow for having them accept prior to becoming full users in the system / when they go to login initially? I imagine normal users will go through a terms checkbox during their setup at some point.

Or will that be handled at the box office when they go to make their purchase via some notice?

api/src/controllers/orders.rs
mikethetike: Good find.

api/src/controllers/cart.rs
mikethetike: It's intentional this time. if you use Some(x), it will move the value and then you can't use it in line 307

api/src/controllers/cart.rs
mikethetike: Again, it's intentional. The find_by_phone might also return a `None`

api/src/controllers/cart.rs
mikethetike: This is a very good point. At the moment I don't think we have a terms and conditions agreement at any stage.

Issue #604, opened 2018-11-27
Make expiry time on empty carts a very long time


mikethetike:
When a cart is first created, and it is empty, make it not expire, or have a really long expiry time. #603 already resets the expiry time when an item is added to an empty cart. When the last item is removed from a cart, it should reset the expiry time to not expire

PR #626, opened 2018-11-30
Remove expiration from empty carts

Timo614:
Closes: #604

This branch closes the associated issue by setting a cart's expiration to `null` by default, sets it on cart update if it doesn't yet exist, and sets it back to `null` if all items are removed.


db/src/models/orders.rs
```diff
@@ -213,6 +244,17 @@ impl Order {
 
         let current_items = self.items(conn)?;
 
+        // Set cart expiration time if not currently set (empty carts have no expiration)
+        if self.expires_at.is_none() {
```
mikethetike: I believe this should be done around line 402, so that if you remove all tickets of one type, and add new tickets of another type, then the expiry resets. This is particularly import for when the `ReplaceCart` method is used by the apps.

db/src/models/orders.rs
```diff
@@ -358,12 +400,6 @@ impl Order {
             }
         }
 
-        // if the cart is empty at this point, it is effectively a new cart, extend the time
-
-        if self.items(conn)?.len() == 0 {
```
mikethetike: This is the check I am referring to in the previous comment

db/src/models/orders.rs
Timo614: Sounds good, moved it below and modified the logic a bit.

Issue #594, opened 2018-11-26
When signing in, email should not be case sensitive


Krakaw:
Currently emails are case sensitive.

Krakaw:
`tolower(email) = email.tolower()`

PR #629, opened 2018-12-03
force user emails to lowercase to allow case insensitivity

SWvheerden:
#594


db/src/models/users.rs
```diff
@@ -118,10 +118,15 @@ impl User {
         password: &str,
     ) -> NewUser {
         let hash = PasswordHash::generate(password, None);
+        let lower_email = if let Some(i) = email {
```
mikethetike:
```suggestion
        let lower_email = email.map(|e| e.to_lowercase())
```

PR #645, opened 2018-12-04
Fix intermittent panic when replacing cart items

### LINKED ISSUE #586: Panic when trying to replace items in a cart

philipr-za:
Fixes #586 

bn-api would panic sometimes when replacing carts using the PUT /cart endpoint if you had multiple items in the cart. I think it was intermittent because it depended on the order of the order_items in the db as the cart was updated. Sometimes the update process would remove the second element in the `mapped` vector without compensating the index values still in the vector and then try remove the value at index 1 when the vector only had 1 value in it.

Fixed by not removing items from the vector but using an Option on the index value in the tuple to indicate that the item had been dealt with.


mikethetike:
Nice find!

Issue #627, opened 2018-12-02
Remove purchase limit errors when checking out through box office


Jasonvdb:
Because the tickets are added to the box office guy's/girl's cart first. If there's a limit on the number of tickets a user can buy, the box office person can't sell more than that number of tickets.

philipr-za:
Closed by https://github.com/big-neon/bn-api/pull/651

PR #651, opened 2018-12-05
Exclude on-behalf-of tickets from count when checking ticket limits for box office

philipr-za:
#627 

Fixed by excluding order-items from orders with a non-null on-behalf-of-user field when counting how many tickets a user has bought or has in their cart. This still means that if a box office user tries to add more than the limit of tickets to a single cart it will hit the limit but it wont count previous orders on behalf of other users when checking for limits.

Added postman test to confirm that the limit is enforced in a single order and that previous on-behalf-of orders are excluded when checking the limit.


philipr-za:
Cool, that makes sense. The other user could be at the limit and then go to the box office and buy further tickets. Also a user could hit the box office multiple times to go over the limit though will need to be in multiple orders.

I will make a ticket describing that.

philipr-za:
Closed by https://github.com/big-neon/bn-api/pull/651

PR #681, opened 2018-12-06
#678 Make venue fields required

### LINKED ISSUE #229: Add google places ID and coordinates to venues
Linking to google maps on the web using the place_id is more accurate. Right now it's making a google maps link using the address, but that could go wrong.

Storing the coordinates might come in handy later I'm sure so just popping that in there.

`google_place_id`
`lat`
`lng`

Requires big-neon/bigneon#20

### LINKED ISSUE #678: Make Venue fields required

mikethetike:
Closes #678 
Closes #229 

Also will create a new Region if a region with the name = venue.State does not exists


db/migrations/00000000000048_make_venue_fields_not_null/down.sql
```diff
@@ -0,0 +1,41 @@
...
+    ALTER COLUMN country SET NULL;
+
+ALTER TABLE venues
+    ALTER COLUMN postal_code SET NULL;
+
+ALTER TABLE venues
+    ALTER COLUMN state SET NULL;
+
+
+UPDATE venueS
```
Krakaw:
```suggestion
UPDATE venues
```

db/migrations/00000000000048_make_venue_fields_not_null/up.sql
```diff
@@ -0,0 +1,39 @@
+INSERT INTO regions(id, name)
+VALUES ('00000000-0000-0000-0000-000000000000', 'Other');
+
+UPDATE venueS
+SET state       = COALESCE(state, 'Unknown'),
+    reGion_id   = COALESCE(reGion_id, '00000000-0000-0000-0000-000000000000'),
```
Krakaw: region_id   = COALESCE(reGion_id, '00000000-0000-0000-0000-000000000000'),

PR #685, opened 2018-12-06
Update fee schedule min_price column for consistency

### LINKED ISSUE #679: Update the min_price column in the fee_schedule_range table to min_price_in_cents
Update this column name to be consistent with other price columns.

Timo614:
Closes #679 

This branch closes the associated issue by renaming the `fee_schedule_ranges#min_price` database field to `min_price_in_cents` to be consistent with the other fields there for tracking currency.

Opted to do it in a separate migration as it looks like that's the direction we've been going lately (versus updating the old column in its initial migration).

Note the following two endpoints responses will be modified by this change so callers should be updated:

```
Organizations#show_fee_schedule `GET` `/organizations/{id}/fee_schedule`
Organizations#add_fee_schedule `POST` `/organizations/{id}/fee_schedule`
```

I did a search for the other uses and I don't believe we display the `min_price` in any other locations.


Timo614:
Will need to be rebased once the FMT changes come in. Currently erroring out due to the massive number of adjustments the latest FMT logic has.

PR #688, opened 2018-12-07
Sw change fees to not null

SWvheerden:
Changed fees so that they cannot be null. Empty fees are counted as 0.


Cargo.lock
```diff
@@ -1,29 +1,29 @@
 [[package]]
 name = "actix"
-version = "0.7.5"
+version = "0.7.8"
```
mikethetike: I think undo the Cargo.lock mass upgrades

SWvheerden:
check fee incoming parameters as well

db/src/models/orders.rs
```diff
@@ -568,11 +568,11 @@ impl Order {
                 quantity: 1,
                 parent_id: None,
             };
-            if let Some(event_fee_in_cents) = event.fee_in_cents {
-                new_event_fee.unit_price_in_cents = event_fee_in_cents;
+            if event.fee_in_cents > 0 {
```
mikethetike: You can remove this whole `if` statement and replace with 
```suggestion

 new_event_fee.unit_price_in_cents = event.fee_in_cents;
                new_event_fee.commit(conn)?;	                
```

db/src/models/organizations.rs
```diff
@@ -90,8 +90,8 @@ pub struct OrganizationEditableAttributes {
     pub postal_code: Option<String>,
     #[serde(default, deserialize_with = "deserialize_unless_blank")]
     pub phone: Option<String>,
-    #[serde(default, deserialize_with = "double_option::deserialize")]
-    pub event_fee_in_cents: Option<Option<i64>>,
+    #[serde(default, deserialize_with = "deserialize_unless_blank")]
```
mikethetike: I think this attribute is only used for strings

db/src/models/ticket_instances.rs
```diff
@@ -57,6 +57,7 @@ impl TicketInstance {
             .inner_join(wallets::table.on(ticket_instances::wallet_id.eq(wallets::id)))
             .inner_join(events::table.on(ticket_types::event_id.eq(events::id)))
             .filter(ticket_instances::id.eq(id))
+            .filter(events::status.ne(EventStatus::Draft.to_string()))
```
mikethetike: This has slipped in from a different PR

db/src/models/ticket_instances.rs
```diff
@@ -99,6 +100,7 @@ impl TicketInstance {
             .inner_join(events::table.on(ticket_types::event_id.eq(events::id)))
             .filter(ticket_instances::id.eq(id))
             .filter(events::id.eq(event_id))
+            .filter(events::status.ne(EventStatus::Draft.to_string()))
```
mikethetike: Same

db/src/models/ticket_instances.rs
```diff
@@ -140,6 +142,7 @@ impl TicketInstance {
                     end_time.unwrap_or_else(|| NaiveDate::from_ymd(3970, 1, 1).and_hms(0, 0, 0)),
                 ))
                 .filter(wallets::user_id.eq(user_id))
+                .filter(events::status.ne(EventStatus::Draft.to_string()))
```
mikethetike: Same

db/src/test/builders/event_builder.rs
```diff
@@ -112,7 +112,7 @@ impl<'a> EventBuilder<'a> {
         .unwrap();
 
         let update_event_fee = EventEditableAttributes {
-            fee_in_cents: self.fee_in_cents,
+            fee_in_cents: self.fee_in_cents.unwrap_or(0),
```
mikethetike: Is this needed?

SWvheerden: yes this is, self in this case is optional.

db/migrations/00000000000051_make_fees_not_null/down.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE organizations
+    ALTER COLUMN event_fee_in_cents DROP NOT NULL;
```
mikethetike:
```suggestion
    ALTER COLUMN event_fee_in_cents SET NULL;
```

db/migrations/00000000000051_make_fees_not_null/down.sql
```diff
@@ -0,0 +1,9 @@
...
+ALTER TABLE organizations
+    ALTER COLUMN event_fee_in_cents DROP DEFAULT;
+
+ALTER TABLE events
+    ALTER COLUMN fee_in_cents DROP NOT NULL;
```
mikethetike:
```suggestion
    ALTER COLUMN fee_in_cents SET NULL;
```

db/migrations/00000000000051_make_fees_not_null/down.sql
SWvheerden: up to you how we do this, I am happy both ways, this was just the way postgress documentation mentioned it.

PR #698, opened 2018-12-10
Merge spotify result and client data

Krakaw:
When creating an artist from spotify, merge in any other data that was sent along with it such as bio's and 3rd party urls

### DIFF
--- a/db/src/models/artists.rs
+++ b/db/src/models/artists.rs
@@ -74,6 +74,42 @@ impl NewArtist {
                 .get_result(conn),
         )
     }
+
+    pub fn merge(&mut self, new_artist: NewArtist) {
+        if self.bio == "" {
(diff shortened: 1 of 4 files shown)


db/src/models/artists.rs
```diff
@@ -74,6 +74,42 @@ impl NewArtist {
                 .get_result(conn),
         )
     }
+
+    pub fn merge(&mut self, new_artist: NewArtist) {
+        if self.bio == "" {
+            self.bio = new_artist.bio;
+        }
+        if let None = self.image_url {
```
mikethetike: OMG isn't there a better way of doing this

Krakaw: One would think! This mostly worries me because if we change the fields that we expect for a `NewArtist` you just have to magically know to update it here too.

Issue #702, opened 2018-12-11
Return org roles in `/organizations/{id}/users`


Jasonvdb:
We need to show all the roles a user has in the member list.

Related to: https://github.com/big-neon/bn-api/issues/701

sdbondi:
Fixed by https://github.com/big-neon/bn-api/pull/774

PR #704, opened 2018-12-11
Sw return org roles in org users request

SWvheerden:
this allows the web to see roles assigned to a user
#702


Jasonvdb:
@SWvheerden does this return the org lever roles? OrgOwner/OrgMember/OrgBoxOffice/OrgScanner

SWvheerden:
It should return these:
string_enum! { Roles [Admin, OrgMember, OrgOwner, OrgBoxOffice, OrgScanner, User] }

Jasonvdb:
Don't you need `users.get_roles_by_organization()` for that?

SWvheerden:
Not as far as I can pickup

PR #709, opened 2018-12-11
Modified show_from_venues to display only the active events

### LINKED ISSUE #707: show_from_venues is returning cancelled published events
When you open a venue on bigneon to see what events are hosted by this venue, it also show previously published cancelled events.

I believe the problem is with show_from_venues function of the /venues/{id}/events entry point of bn-api

neonknight64:
The show_from_venues function had an issue where it displayed previously published but cancelled events.
Overview of changes:
- Added find_all_active_events_for_venue memberber function for Event
- Added venue unit test find_active_for_venue

Closes #707


db/src/models/events.rs
```diff
@@ -289,6 +289,22 @@ impl Event {
         )
     }
 
+    pub fn find_all_active_events_for_venue(
```
mikethetike: Is the previous method `find_all_events_for_venue()` used by anything else anymore?

neonknight64: It is used by two tests, one to test itself and another to test something else.

PR #712, opened 2018-12-11
Refactor Org Roles

mikethetike:
Probably worth double checking this one


api/src/controllers/organization_invites.rs
```diff
@@ -40,62 +40,47 @@ pub fn create(
 ) -> Result<HttpResponse, BigNeonError> {
     let connection = connection.get();
     let organization = Organization::find(path.id, connection)?;
-    auth_user.requires_scope_for_organization(Scopes::OrgWrite, &organization, connection)?;
+    auth_user.requires_scope_for_organization(Scopes::OrgManageUsers, &organization, connection)?;
 
-    let invite_args = new_org_invite.into_inner();
+    if &new_org_invite.role != &Roles::OrgOwner && &new_org_invite.role != &Roles::OrgMember {
```
SWvheerden: nit: I think add brackets here, && & looks a bit funny

api/src/controllers/organization_invites.rs
```diff
@@ -151,7 +136,7 @@ pub fn accept_request(
             if valid_for_acceptance {
                 let accept_details = invite_details.change_invite_status(1, connection)?;
                 let org = Organization::find(accept_details.organization_id, connection)?;
-                org.add_user(u.id(), None, connection)?;
+                org.add_user(u.id(), vec![Roles::OrgMember], connection)?;
```
SWvheerden: Should this not be the invite role?
Here you assume OrgMember, but as I understand it, it can be both orgmember or orgowner?

api/src/controllers/organizations.rs
```diff
@@ -213,7 +195,7 @@ pub fn add_user(
     let connection = connection.get();
     let organization = Organization::find(path.id, connection)?;
     user.requires_scope_for_organization(Scopes::OrgWrite, &organization, connection)?;
-    organization.add_user(add_request.user_id, None, connection)?;
+    organization.add_user(add_request.user_id, vec![Roles::OrgMember], connection)?;
```
SWvheerden: how do we add other roles?
Would we not rather use the same function, and just pass in the roles?

api/tests/functional/base/organization_invites.rs
```diff
@@ -50,43 +50,6 @@ pub fn create(role: Roles, should_test_succeed: bool) {
     }
 }
 
-pub fn create_for_existing_user_via_user_id(role: Roles, should_test_succeed: bool) {
```
SWvheerden: is there a reason for removing test rather than updating?

api/tests/functional/base/organization_invites.rs
```diff
@@ -119,39 +82,6 @@ pub fn create_for_new_user(role: Roles, should_test_succeed: bool) {
     }
 }
 
-pub fn create_failure_missing_required_parameters(role: Roles, should_test_succeed: bool) {
```
SWvheerden: is there a reason for removing test rather than updating?

api/tests/functional/base/organizations.rs
```diff
@@ -348,36 +341,6 @@ pub fn add_artist(role: Roles, should_test_succeed: bool) {
     }
 }
 
-pub fn update_owner(role: Roles, should_succeed: bool) {
```
SWvheerden: is there a reason for removing test rather than updating?

db/migrations/00000000000052_org_roles/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table organization_invites
+    add role text not NULL
```
SWvheerden: add a default?

SWvheerden: missing ";"

db/migrations/00000000000052_org_roles/down.sql
```diff
@@ -0,0 +1,2 @@
+alter table organization_invites
+    drop column role
```
SWvheerden: missing ";"

db/migrations/00000000000053_add_role_to_organization_invite/down.sql
```diff
@@ -0,0 +1 @@
+LOLZ YOU CAN'T GO BACK NOW
```
SWvheerden: add comment marker?

db/migrations/00000000000053_add_role_to_organization_invite/up.sql
```diff
@@ -0,0 +1,21 @@
+alter table organization_users
+    rename column "role" to old_role;
+
+
+alter table organization_users
+    add "role" text[] not NULL;
```
SWvheerden: default?

db/src/models/artists.rs
```diff
@@ -187,7 +186,6 @@ impl Artist {
                 .filter(
                     organization_users::user_id
                         .eq(u.id)
-                        .or(organizations::owner_user_id.eq(u.id))
```
SWvheerden: dont removal of these add security risk?

api/src/controllers/organization_invites.rs
mikethetike: This is the same as the previous behaviour, I've just moved the `None => Roles::OrgMember` that happened inside the method

api/src/controllers/organizations.rs
mikethetike: See other comment.

api/tests/functional/base/organization_invites.rs
mikethetike: I removed the functionality of inviting via user_id. All invites must be done via email

api/tests/functional/base/organization_invites.rs
mikethetike: I removed the user_id parameter and made user_email non-optional, so this case no longer exists

api/tests/functional/base/organizations.rs
mikethetike: Owners are now added the same as other org members, via specifying a role, so this method no longer exists on Organization

db/migrations/00000000000052_org_roles/down.sql
mikethetike: ; is only required for multiple queries in a single file

db/migrations/00000000000052_org_roles/up.sql
mikethetike: Good point. Updated

db/migrations/00000000000053_add_role_to_organization_invite/down.sql
mikethetike: It is intentional that this doesn't compile, it should generate an error so that the user knows they should manually intervene

db/migrations/00000000000053_add_role_to_organization_invite/up.sql
mikethetike: I think this defaults to an empty array

db/src/models/artists.rs
mikethetike: the field no longer exists in the db

Issue #484, opened 2018-10-30
Improve json deserialization errors to include more relevant information


Timo614:
It was noticed in the IRC chat that at times the json deserialization errors returned by actix-web during the extraction process are a bit useless for debugging what could be wrong with the request.

It appears as though we can create a custom error handler for responding to these sorts of issues:
https://actix.rs/actix-web/actix_web/dev/struct.JsonConfig.html

We should attempt to make the responses more helpful for clients debugging why their API interaction is not working.

Timo614:
The example associated appears to be used on a per route basis. Seeing if there's any way to do it globally as that wouldn't work in practice.

Timo614:
It looks like we will have to do it per route. You have to select the extractor config from the tuple items as well which is dependent on the ordering in the method itself.

Existing logic (per route for any using json extractor):
```
        r.method(Method::POST).with(venues::create);
```

With config:
```
        // The c.0 is to get into the parameter tuple and the .1 is the location in the tuple for the enum extractor
        r.method(Method::POST).with_config(venues::create, |c| {
            (c.0).1.error_handler(json_error);
        });
```

Master behavior:
```
curl --header "Content-Type: application/json" --request POST --data '{"email":"superuser@test.com","password":"password"}' http://localhost:9000/venues --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI5MWFmZDcyMS1iNmY0LTRmMWUtOTYwOC01ZTMwYjJjZTg4MWMiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQ0NTUxOTQxfQ.O34a3v8cZm7Hb7EVsRscRM3I9ut8zTsOngGZJqv3zY4" -i
HTTP/1.1 400 Bad Request
access-control-allow-origin: http://localhost:3000
vary: Origin
x-app-version: 0.1.22
content-length: 0
date: Tue, 11 Dec 2018 17:59:34 GMT
```

Local branch with the new config:
```
[...]

PR #713, opened 2018-12-11
Improve json error response

Timo614:
Closes: #484

This pull request fixes the associated issue by returning json in the response when there is a json error caused during the json extractor. 

To accomplish this I'm using a custom extractor that uses the actix-web JsonBody logic to handle deserialization but provides an error handling function which returns the json error. I originally was going to go the route of using route configs but it's messy, error prone, and it requires specifying tuple positions for the config so it's bound to make controller action modifications a pain. This approach just requires using our custom `Json` extractor.

Master behavior:
```
curl --header "Content-Type: application/json" --request POST --data '{}' http://localhost:9000/venues --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI5MWFmZDcyMS1iNmY0LTRmMWUtOTYwOC01ZTMwYjJjZTg4MWMiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQ0NTUxOTQxfQ.O34a3v8cZm7Hb7EVsRscRM3I9ut8zTsOngGZJqv3zY4" -i
HTTP/1.1 400 Bad Request
vary: Origin
x-app-version: 0.1.22
content-length: 0
date: Tue, 11 Dec 2018 17:59:34 GMT
```

Local branch with the new config:
```
curl --header "Content-Type: application/json" --request POST --data '{}' http://localhost:9000/venues --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI5MWFmZDcyMS1iNmY0LTRmMWUtOTYwOC01ZTMwYjJjZTg4MWMiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQ0NTUyOTUxfQ.JXQD6QQGzmP8xCmPv546-vjy16HfWutLwGHqmZWi6XE" -i
HTTP/1.1 400 Bad Request
content-length: 52
vary: Origin
x-app-version: 0.1.22
date: Tue, 11 Dec 2018 18:14:40 GMT

{"error":"missing field `name` at line 1 column 52"}
```


Timo614:
An alternative approach could be to make a new extractor that we use in place of the existing extractor.

It looks like the majority of the work done by the extractor is done in the JsonBody struct so may quickly see if that approach works better. Would feel a bit more straightforward as we wouldn't need to specify the tuple location for the JsonConfig object when handling the config / the routes would be more straightforward.

Issue #689, opened 2018-12-07
Encrypt api keys while at rest


Krakaw:
Use a symmetric key encryption (re-use one of the available libraries we're already using)

PR #716, opened 2018-12-12
Encrypt organizations API keys

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/689

Implement encryption of organization API key fields when written to DB and allow for decryption when reading from DB.


api/src/controllers/organizations.rs
```diff
@@ -54,14 +55,22 @@ pub struct NewOrganizationRequest {
 }
 
 pub fn index(
-    (connection, query_parameters, user): (Connection, Query<PagingParameters>, User),
+    (state, connection, query_parameters, user): (
```
mikethetike: I don't think this is necessary in the index method

api/src/controllers/organizations.rs
```diff
@@ -71,12 +80,19 @@ pub fn index(
 }
 
 pub fn index_for_all_orgs(
-    (connection, query_parameters, user): (Connection, Query<PagingParameters>, User),
+    (state, connection, query_parameters, user): (
```
mikethetike: Also not necessary

PR #725, opened 2018-12-13
Sw modify postman tests to test for min role

SWvheerden:


SWvheerden:
I had to modify the one api call because the doorman scope is linked to an organization. The api call is there to get all events the doorman can checkin, but we need to know which organization he is linked to so that we can confirm his rights, then give that event, which events he can scan. 

With the previous admin rights the call had, it overrides the checking that allows the admin to get the rights

api/src/controllers/events.rs
```diff
@@ -69,9 +69,15 @@ impl From<SearchParameters> for Paging {
...
+    (conn, parameters, query, auth_user): (
+        Connection,
+        Path<PathParameters>,
+        Query<SearchParameters>,
+        User,
+    ),
 ) -> Result<HttpResponse, BigNeonError> {
-    auth_user.requires_scope(Scopes::RedeemTicket)?;
+    let org = Organization::find(parameters.id, conn.get())?;
+    auth_user.requires_scope_for_organization(Scopes::RedeemTicket, &org, conn.get())?;
```
Timo614: I think this logic may be a bit flawed for confirming access since it only checks the passed in organization id but the events returned may be from multiple organizations given it's not filtering to them.

I think we could either:
1. We remove the requires scope logic above, remove the organization id that is now being passed in, and adjust `find_events_with_access_to_scan` to filter by organizations the user has the `Scopes::RedeemTicket` for (just looping over their organizations and filtering down to the ones they have that scope for). Then if the user has none it just returns an empty array for available events.
2. We pass in the organization id to `find_events_with_access_to_scan` so that it filters to just the given passed in organization.

I'm not sure how the doorman app is used which may determine which is more ideal. I'd think that a doorman may be in that role for multiple organizations at a given venue so it may be ideal that they don't have to filter down by specific organizations so the first option may be a better call.

@Krakaw do you know how the app is set to work in that regards? Should the doorman have to specify the organization he's checking in for when getting this information?

api/src/controllers/events.rs
```diff
@@ -69,9 +69,15 @@ impl From<SearchParameters> for Paging {
...
-    (conn, query, auth_user): (Connection, Query<SearchParameters>, User),
+    (conn, parameters, query, auth_user): (
+        Connection,
+        Path<PathParameters>,
+        Query<SearchParameters>,
+        User,
+    ),
 ) -> Result<HttpResponse, BigNeonError> {
-    auth_user.requires_scope(Scopes::RedeemTicket)?;
+    let org = Organization::find(parameters.id, conn.get())?;
```
mikethetike:
```suggestion
     let org = Organisation::find(parameters.id, conn.get())?;
```

api/src/controllers/events.rs
Krakaw: In most cases a user will usually only be a part of a single organization, however they could be in multiple. The app just wants a list of available events for this user to scan without jumping through hoops to find those events. Option 1 definitely seems like the best way to go.

SWvheerden:
Cwl, will quickly implement option 1 tonight

Issue #485, opened 2018-10-30
Add constraints in db for enum related fields values


Timo614:
We currently do not have constraints in place to prevent a db record from being created with an invalid status or type where applicable for the enums related fields. Ideally we could just switch the structs to use an enum value instead and have serde / diesel handle the translation back and forth between the string value to limit what the API could set. 

It would also be good if we added db constraints to prevent the fields from being set to values that aren't applicable although if we manage to have serde and diesel handle the enum type we won't have to worry about that as heavily (still good to prevent database integrity issues though).

PR #727, opened 2018-12-14
WIP: Enum improvements

Timo614:
Closes: #485

This PR closes the associated issue by modifying the way our enum values are stored and retrieved from the database to remove the need to manually set values to strings and parse out the enums in favor of having diesel handle it for us.

This has the benefit of allowing code like: `assert_eq!(event.status().unwrap(), EventStatus::Published);`

To become: `assert_eq!(event.status, EventStatus::Published);`

It has another benefit in that it allows us to restrict the likelihood of bad data getting into the database. Since these are strongly typed and limited to what we have available via the enum there's never a step in the logic where a string value not associated with the enum could be set.

I've converted some of the enums so far but am still working through the list so this is currently a WIP. Would like to do a little bit of additional testing as well.


Timo614:
Opted not to add constraints on the db itself as it'll add overhead whenever we introduce new types requiring a migration to modify the constraints. Since the enum values are strongly typed this enforces integrity albeit someone could access the db itself and put bad values there directly. Could still add db constraints if we'd like the additional verification for that scenario though.

PR #743, opened 2018-12-15
Added event:scan back to scopes

### LINKED ISSUE #742: Re-introduce the event:scan permission

Krakaw:
Closes #742 
The app is currently checking if the user has event:scan before allowing
the scanner to open


db/src/models/scopes.rs
```diff
@@ -103,25 +115,19 @@ fn get_scopes_for_role(role: Roles) -> Vec<Scopes> {
...
         OrgAdmin => {
             let mut roles = vec![Scopes::OrgWrite, Scopes::UserRead, Scopes::OrgManageUsers];
             roles.extend(get_scopes_for_role(OrgMember));
-            roles.extend(get_scopes_for_role(Roles::OrgBoxOffice));
-            roles
-        }
-        OrgBoxOffice => {
-            let mut roles = vec![Scopes::EventViewGuests, Scopes::OrderMakeExternalPayment];
-            roles.extend(get_scopes_for_role(Roles::DoorPerson));
+            //Remove the duplicate event:scan from OrgBoxOffice because OrgMember contains it
```
philipr-za: I think there is a dedup function for vectors that if you sort first it will remove all dups. I think there are other dups because of this extending so .sort().dedup() will sort those out too.

Krakaw: Awesome thanks, I'll keep that in mind for next time. The scopes are all manually entered in the tests (i think they are sorted, but that is an issue for post-MVP :P

Issue #740, opened 2018-12-14
Add client_event_fees_in_cents and company_event_fees_in_cents to organizations and events


philipr-za:
Add two extra event_fee fields to organizations to be applied to events on creation.

PR #763, opened 2018-12-18
Split organization event_fees_in_cents into company_event_fees_in_cents

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/740

Splits the `event_fees_in_cents` in Organizations and Events into `client_event_fees_in_cents` and `company_event_fees_in_cents`. Made it so that the summed event_fees are not directly editable in either struct but are calculated from the client and company fees on creation and update.

This will not break the front end but it will ignore any attempt to directly edit the summed `event_fees_in_cents` or `fees_in_cents` in Organizations or Events respectively so the front end will need to be updated to rather edit the new fields.


db/migrations/00000000000055_add_client_and_company_event_fees/down.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE organizations DROP COLUMN client_event_fee_in_cents;
+ALTER TABLE organizations DROP COLUMN company_event_fee_in_cents;
+
+ALTER TABLE events ADD client_fee_in_cents BIGINT NULL;
```
Krakaw: This should be DROP

db/migrations/00000000000055_add_client_and_company_event_fees/down.sql
```diff
@@ -0,0 +1,5 @@
...
+ALTER TABLE events ADD company_fee_in_cents BIGINT NULL;
```
Krakaw: This should be DROP

PR #765, opened 2018-12-18
Create and save domain events when an organization is created

mikethetike:
Also added some logging to be picked up on the server for logins and refreshes


db/src/models/enums.rs
```diff
@@ -75,7 +75,10 @@ macro_rules! string_enum {
 string_enum! { AssetStatus [Unsynced] }
 string_enum! { CodeTypes [Access, Discount] }
 string_enum! { CommunicationChannelType [Email, Sms, Push]}
-string_enum! { DomainEventTypes [OrderBehalfOfUserChanged, PaymentCreated, PaymentCompleted, PaymentMethodCreated, PaymentMethodUpdated, UserRegistration, LostPassword, PurchaseCompleted]}
+string_enum! { DomainEventTypes [FeeScheduleCreated, OrderBehalfOfUserChanged, OrganizationCreated,
+PaymentCreated,
```
Timo614: Not a huge deal but may be worth adjusting this formatting here as it looks a bit odd with the two rows with just one items in them (unless fmt forced this which it may have).

PR #770, opened 2018-12-18
Allow Publish at a future date

mikethetike:


db/src/models/events.rs
```diff
@@ -222,16 +223,28 @@ impl Event {
...
-        &self,
+        self,
         attributes: EventEditableAttributes,
         conn: &PgConnection,
     ) -> Result<Event, DatabaseError> {
         attributes.validate()?;
+
+        let mut event = attributes;
+
+        if self.status == EventStatus::Published {
```
Timo614: So this scenario is to cover situations where an event has already been published but lacks the publish date? 

Is there a path for that? The event editable attributes logic above doesn't include the status so I had thought the only way to get into that state would be via the publish method and that always sets a published date.

Not sure if it's needed since one can only change the publish_date via the update call but just confirming.

We may want to prevent a published event from changing its publish date though as I don't think there's some way to unpublish an event.

Timo614: Oh I see the new event flow does accept EventStatus so this scenario could appear from that flow. I wonder if we should just modify its commit logic to set the publish date there if it's coming in as published.

mikethetike:
@Timo614 
Re: preventing a publishing date from being edited means that if they choose a date too far in the future, then they cannot change it. Currently there is no "Unpublish" functionality, so this is currently the only way address that

PR #775, opened 2018-12-18
Dont add fees for zero cost tickets

philipr-za:
Closes https://github.com/big-neon/bn-api/issues/773

If a ticket has zero cost don't even try look up the fee schedule. 
If all the tickets in a cart for an event have zero fees then don't add Event fees.


db/src/models/order_items.rs
```diff
@@ -72,45 +72,54 @@ impl OrderItem {
...
-            .get_range(self.unit_price_in_cents, conn)?;
+        let ticket_pricing = TicketPricing::get_current_ticket_pricing(ticket_type.id, conn)?;
 
-        // If the hold is a comp, then there are no fees.
-        if let Some(hold_id) = self.hold_id {
-            let hold = Hold::find(hold_id, conn)?;
-            if hold.hold_type == HoldTypes::Comp {
-                if let Some(fee_item) = fee_item {
-                    order.destroy_item(fee_item.id, conn)?;
+        if ticket_pricing.price_in_cents > 0 {
```
mikethetike: Rather use self.unit_price_in_cents > 0

mikethetike: Also might be worth getting the lowest fee schedule range and checking for less than that

philipr-za: Done and done.

So if the unit_price_in_cents is lower than the lowest minimum price in the schedule it will not apply PerItemFees but if there is even one non zero OrderItem the EventFees will be applied.

Issue #711, opened 2018-12-11
Add missing roles


Timo614:
Related to #701

Need to add additional roles to the organizations and adjust functional tests to take them into account.

Org Owner - Can manage team (including adding admins), view all reports, update org details, create events, manage tickets, manage holds, sell tickets through box office, check people into events
Org Admin - Can manage team (cannot add other admins), view all reports, update org details, create events, manage tickets, manage holds, sell tickets through box office, check people into events
Org Member - View Events List/Dashboard, View all reports without fees, create events, manage tickets, manage holds, sell tickets through box office, check people into events
Box Office - View Events List/Event Dashboard, view orders, view tickets, view holds, sell tickets through box office, check people into events
Doorperson - View Events List, view holds, check people into events

Krakaw:
@mikethetike is redoing roles

Timo614:
Gotcha, will reassign this his way.

mikethetike:
I think I have done most of these. Might be worth someone doing a review to make sure that they are correct

Timo614:
Will do a review of these today.

PR #779, opened 2018-12-19
Adjust tests for new roles

Timo614:
Closes: #711

This pull request closes the associated issue by adjusting a few controller actions to change their scope requirements and with new tests for the new roles introduced into master to confirm that access.

- I've adjusted the `events#dashboard` route to have a new scope `DashboardRead` as it was previously using `EventWrite` and box office needs dashboard access.
- Switched `tickets#show` and `tickets#show_redeemable_ticket` to use `TicketRead` in place of `TicketAdmin` so access can be granted to Door People and Box Office.
- Switched a few tests that were tested across roles to be only tested once (mostly validation ones as we already confirm endpoint access with the success case done across the roles).

This does increase the test time a bit as we're covering more roles now. Perhaps in the future it'd be worth us having some way to only run `Admin` related tests locally but then having Travis do them all. Most of Travis's test time is devoted to the building of the crates so it shouldn't add too much time there relatively on a per test run perspective.


Timo614:
I think we should be set here with the changes I made in this PR. Any additional access issues we can probably tackle with a new issue and quickly address. I think we're set though between the scopes they had access to before and this.

PR #826, opened 2019-01-01
pre-push migrations check

Krakaw:
A rudimentary check that pulls the most recent migration filename from the master repo and compares it to the latest local version on your machine.
If your local version is lower than the remote version it will abort the commit. This would usually happen when you're working on a large PR, create your migration and another, smaller PR with a migration generated after yours gets merged it. If that happens your migration won't get run as the beta / studio migrations will be ahead of yours.


Krakaw:
Diesel doesn't care about the order of the migrations, it checks which ids have been run (the ID being the filename) so this PR is closed.

Issue #785, opened 2018-12-20
Order Support - Refunds - A org member should be able to refund ticket purchases


zmelamed:
An org member should be able to view the order details page for any order.

The order detail page should show a table (similar to the table in the box office guest list view) listing each ticket in that order (with fees, face value, and total price listed for each ticket) and should also have any event fee that was paid if applicable.

The org member should be able to select all or just one of the table items which will enable a Refund button.

After clicking refund button, the org member should be prompted to confirm the refund and  should receive an email that the order was refunded, and the items refunded should still appear in that order detail list, but there should be a label indicated that the item was refunded.

Timo614:
I've been working on this one for a little bit hoping to wrap it up next week. I've run into a few scenarios I was hoping to clarify though @zmelamed if you have a chance:

- Should comps be allowed to be refunded to put them back in the available list? Or only allow refunds on paid orders?
- Are there any situations where we should cutoff when refunds are allowed? Or can a purchased and redeemed ticket be refunded?
- Normally tickets are returned to available inventory after refunding right? Or should that be an option as part of the refund to allow refunded but still existing refunds?
- Transferred tickets are not eligible for any refunds right? (I assume so as the person ordering no longer has access).
- Should fees be recalculated given the pricing will be different? What if the lower item quantity ends up having higher fees? Should we only refund part of the ticket price and then modify the fee order item to reflect the new fee? It may get a bit confusing.
- Not really a problem for the MVP as it's only a mockchain for now but we'll likely need to make refunds contingent on the wallet having and transferring out the tickets back to our organization wallet right?
- Event fees are automatically removed if there are no existing tickets left for the event right? (I'm assuming such with the code at the moment but confirming here).

zmelamed:
@Timo614, this is so helpful, I really appreciate this

- Should comps be allowed to be refunded to put them back in the available list? Or only allow refunds on paid orders?
  - A refund of a comp should be possible, and would just reverse the comp order if that makes sense
- Are there any situations where we should cutoff when refunds are allowed? Or can a purchased and redeemed ticket be refunded?
  - Purchased and redeemed tickets can be refunded
- Normally tickets are returned to available inventory after refunding right? Or should that be an option as part of the refund whether to allow the user to keep the ticket?
  - I've posed this question to the business team but for now assume the ticket will go back to inventory after refunding.
- Transferred tickets are not eligible for any refunds right? (I assume so as the person ordering no longer has access).
  - Correct
[...]

PR #846, opened 2019-01-04
Refunds

Timo614:
Closes: #785

This branch attempts to resolve the associated issue by adding in the capability to refund individual tickets and their fees. The logic can be a bit complex at times due to the fact tickets are no longer associated with an order item if they are released to be purchased again. I track which tickets did exist though so it can generate the order details page listing the ones that are now refunded. Order items have a refunded_quantity field which is used to keep track of if they have been refunded.

Would be good to have a few eyes on these changes as refunding is a bit complex given all its needs (triggering the refund back to the payment method, transferring tokens on the mockchain, calculating the amount needed to refund ie if we need to remove the event fees as well, etc.).

Based some of the logic here based on the responses to [these questions](https://github.com/big-neon/bn-api/issues/785#issuecomment-450429192).

Questions:
- In the event a ticket is already Redeemed I don't release it back to inventory. Should refunding still take it back to the organization's wallet? Just don't allow purchasing again? I am taking it at the moment to prevent transferring etc. later on but it's redeemed so they likely can't be reused at the venue.
- How will partial payment system handle refunding automatically?
- If refund fails with Stripe should we stop the entire process and transfer wallet items back? Issue credit?
- If the wallet transfer succeeds for some tickets but fails for others should we attempt to send back the ones that did succeed? Reset things back to an unrefunded state prior to reverting the db transactions?

Not covered / new issues (scope is already pretty large here so think it may be best to follow up with further adjustments and review what we have now):
- Receiving an email for the order getting refunded template (I lack access to the sendgrid account being used by staging). Haven't had a chance to test this as I don't have a sendgrid account but should work as it mimics the order purchase email with some slight adjustments and compiles without issue.
- Handling wallet issues such as an invalid secret preventing transfer (an issue for post mvp likely).
- Some additional logic for the purposes of settlement to adjust for refunds outside of the initial settlement window for an order?

Examples:

An order details call response (with one of the ticket refunded):
```
curl http://localhost:9000/orders/3a48cfcd-ec2b-42d2-8c7d-8c42e7b32dea/details -i --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI5Yjk1ZDJhYS1iNzk4LTQxNDgtYWVkOC03YjAyYmU0MjY1NTkiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQ2NTY3NDg2fQ.Pi5YxA1wY2TxSJEtKo2c2HoiAQ_XxmmDJ861A4xY8FY"
{"items":[{"description":"It's my party - General Admission_1546566522","fees_price_in_cents":10,"order_item_id":"72ff77ff-a9dc-446f-99de-d1ce238060d8","refundable":true,"status":"Purchased","ticket_instance_id":"725709dd-a57a-4f9b-af62-e5f32bc94da9","ticket_price_in_cents":3000,"total_price_in_cents":3010},{"description":"It's my party - General Admission_1546566522","fees_price_in_cents":0,"order_item_id":"72ff77ff-a9dc-446f-99de-d1ce238060d8","refundable":false,"status":"Refunded","ticket_instance_id":"80a1812a-35e2-4755-a001-3681441887c4","ticket_price_in_cents":0,"total_price_in_cents":0},{"description":"Event Fees - It's my party","fees_price_in_cents":100,"order_item_id":"e56ef34c-f82c-4ac3-aa6c-0808b10e808e","refundable":true,"status":"Purchased","ticket_instance_id":null,"ticket_price_in_cents":0,"total_price_in_cents":100}],"order_contains_tickets_for_other_organizations":false}
```

First ticket in order item triggering just ticket and its fee item:
```
curl --header "Content-Type: application/json" --request PATCH --data '{"items":[{"order_item_id": "9d438b4a-49a4-4ff7-a0dc-c016d5284ce6", "ticket_instance_id": "deccfe52-c11e-4fff-b16b-0005a3207cfd"}]}' http://localhost:9000/orders/f34816ff-537b-4671-a692-c1224a349c02/refund -i --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJjOTgzMWU2My03MjRmLTQxNDAtODk2My1hZGNhODExNWVjMGEiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQ2NTY5MDY2fQ.EDl0qAkc0Jjv5CMVh5rxA0ekquk1qYidBhHSxGsTkaE"
{"amount_refunded":3010}
```

Second ticket in order item triggering event fee to also refund:
```
curl --header "Content-Type: application/json" --request PATCH --data '{"items":[{"order_item_id": "9d438b4a-49a4-4ff7-a0dc-c016d5284ce6", "ticket_instance_id": "c8b5dccb-c6e7-4cf1-94bc-d0feb13f330e"}]}' http://localhost:9000/orders/f34816ff-537b-4671-a692-c1224a349c02/refund -i --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJjOTgzMWU2My03MjRmLTQxNDAtODk2My1hZGNhODExNWVjMGEiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTQ2NTY5MDY2fQ.EDl0qAkc0Jjv5CMVh5rxA0ekquk1qYidBhHSxGsTkaE"
{"amount_refunded":3110}
```

image

Domain Event refund logging:
```
select * from domain_events where event_type = 'PaymentRefund';
                  id                  |  event_type   |     display_text     |                                                                                                                                                               event_data                                                                                                                                                               | main_table |               main_id                | published_at |         created_at         |         updated_at         |               user_id                
--------------------------------------+---------------+----------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------+--------------------------------------+--------------+----------------------------+----------------------------+--------------------------------------
 b683e47a-a298-45ea-8ab6-755dda33fbf0 | PaymentRefund | Payment was refunded | {"amount":3010,"balance_transaction":"txn_1Dq2x2DLD9JqYEf35tOyIji7","charge":"ch_1Dq2wyDLD9JqYEf3JRV27YnK","created":1546886152,"currency":"usd","id":"re_1Dq2x2DLD9JqYEf3Id9Re5am","metadata":{},"object":"refund","reason":null,"receipt_number":null,"source_transfer_reversal":null,"status":"succeeded","transfer_reversal":null} | Payments   | 26335e5a-c409-4a51-bfb2-4b28028011e7 |              | 2019-01-07 18:35:57.592561 | 2019-01-07 18:35:57.592561 | df133d3f-8132-48ef-a976-63a1216274c8
 296c8312-8577-4d98-a5a9-a27f856ab8bc | PaymentRefund | Payment was refunded | {"amount":3110,"balance_transaction":"txn_1Dq2x4DLD9JqYEf3zzyYftnM","charge":"ch_1Dq2wyDLD9JqYEf3JRV27YnK","created":1546886154,"currency":"usd","id":"re_1Dq2x4DLD9JqYEf3QddaXqCi","metadata":{},"object":"refund","reason":null,"receipt_number":null,"source_transfer_reversal":null,"status":"succeeded","transfer_reversal":null} | Payments   | 26335e5a-c409-4a51-bfb2-4b28028011e7 |              | 2019-01-07 18:36:00.25058  | 2019-01-07 18:36:00.25058  | df133d3f-8132-48ef-a976-63a1216274c8
```


Timo614:
Moving ahead with adjusting the reports to take refunds into account. I won't be able to do the additional task of making the sendgrid template as I lack access to staging's sendgrid account so will still have to make that a new issue once this PR is set.

May end up making a new branch / PR for the reports but depending upon how large the changes end up being may end up just merging it in here. Will hold off with its own branch if we end up reviewing it to not complicate things though.

Krakaw:
I'm going to sink my teeth into this in the morning and get it merged in.

Timo614:
I've rebased this and adjusted the reports. I ended up setting the `summary_event_report_core` related report SQLs back to not include refunds as they seem to be for settlement purposes checking out the reports controller.

I adjusted the settlements reports as well but I think we may need another report that is able to adjust numbers after a week settles. We may end up refunding outside of the settlement window so I'd think we'd need a report which then itemizes the refunds from the period after the orders were made. Not sure what would be best there.

api/src/controllers/orders.rs
```diff
@@ -32,6 +38,232 @@ pub fn show(
...
+                .create_payment_processor(&payment.provider)?;
+
+            refund_data = match payment.external_reference {
+                Some(ref external_reference) => Some(
+                    client
+                        .partial_refund(external_reference, amount_to_refund)?
+                        .to_json()?,
+                ),
+                None => {
+                    return application::internal_server_error(&format!(
```
mikethetike: Might need to account for box office external payments here

api/tests/functional/users.rs
```diff
@@ -494,7 +494,9 @@ fn current_user_organization_owner() {
             "hold:read",
             "hold:write",
             "order:make-external-payment",
+            "order:details-read",
```
mikethetike: What is the difference between order:read and order:details-read?

db/src/models/order_items.rs
```diff
@@ -52,6 +49,38 @@ impl OrderItem {
...
+        refund_fees: bool,
+        conn: &PgConnection,
+    ) -> Result<u32, DatabaseError> {
+        if self.refunded_quantity == self.quantity {
+            return DatabaseError::business_process_error(
+                "Order item refund failed as requested refund quantity exceeds remaining quantity",
+            );
+        }
+
+        self.refunded_quantity += 1;
```
mikethetike: We should probably check that the order is in a valid status here

mikethetike: I see it is used later, perhaps mark it as pub(crate) so that it is not used by mistake in the Api

db/src/models/payments.rs
```diff
@@ -49,6 +49,24 @@ impl Payment {
         }
     }
 
+    pub fn log_refund(
```
mikethetike: I think we should also add a negative payment so that we can sum all the payments for an order and it balances

db/src/models/ticket_instances.rs
```diff
@@ -43,6 +43,48 @@ impl TicketInstance {
             .to_db_error(ErrorCode::QueryError, "Unable to load ticket")
     }
 
+    pub fn release(&self, conn: &PgConnection) -> Result<(), DatabaseError> {
+        let clear_uuid_value: Option<Uuid> = None;
+        let clear_datetime_value: Option<NaiveDateTime> = None;
+        let clear_string_value: Option<String> = None;
+
+        diesel::update(self)
```
mikethetike: Rather use/alter the current release query, there are intentionally sql locks and filters to ensure that the ticket is not released at the same time as someone buys/updates it

db/src/models/ticket_instances.rs
```diff
@@ -410,6 +452,20 @@ impl TicketInstance {
         Ok(())
     }
 
+    pub fn set_wallet(&self, wallet: &Wallet, conn: &PgConnection) -> Result<(), DatabaseError> {
+        diesel::update(ticket_instances::table.filter(ticket_instances::id.eq(self.id)))
```
mikethetike: I assume this is for an Admin to return the ticket to their wallet. Might be worth making the method do something more specific or make a note to developers that the transfer mechanism should be used in most cases

db/src/queries/reports/reports_transaction_details.sql
```diff
@@ -1,29 +1,21 @@
-SELECT *,
-    CAST ((quantity * unit_price_in_cents) + company_fee_in_cents + client_fee_in_cents AS BIGINT) AS gross
- FROM (
-SELECT e.name                                             AS event_name,
-       CASE
-         WHEN oi.ticket_type_id IS NULL THEN
-           'Per Event Fees'
-         ELSE
-           tt.name END                                    AS ticket_name,
```
mikethetike: Possible bad merge here?

mikethetike:
> Should refunding still take it back to the organization's wallet

I think so

With regards to the failures and rollbacks on the transactions, I think as much as possible, we should return the state of the DB and blockchain to the state it was before the API call was made. Unfortunately, the refund may have already been processed in Stripe, so we must keep those details in the DB, or have some audit of it. Maybe we can separate it into two steps: Create a refund and then process the refund? Just a suggestion. Let's hear some comments and other suggestions before going ahead

api/src/controllers/orders.rs
Timo614: This is only for the credit card ones. For the box office ones was thinking they'd use the amount refunded returned by the API call to then credit the user in person.

api/tests/functional/users.rs
Timo614: `OrderRead` is used by all users to look at their own orders so everybody has it not just organization members.

We could remove all the scopes granted at the `User` level and could reuse that `OrderRead` name in place of `OrderDetailsRead` though as guest accounts do not go through the has_scope_for_organization logic so if someone is logged in they're at the User level.

db/src/models/order_items.rs
Timo614: Sounds good, will do.

db/src/models/payments.rs
Timo614: Sounds good, will add one as well.

db/src/models/ticket_instances.rs
Timo614: Will do.

db/src/models/ticket_instances.rs
Timo614: Will do.

db/src/queries/reports/reports_transaction_details.sql
Timo614: I adjusted the logic a bit to refactor / cleanup some of it. Most of the `CASE` statements here could be switched to `COALESCE` as they were just branching on whether the field was null and using a different value if so. I think it should be good. I also moved the gross calculation to the inner query and removed the outer query to make it clearer.

db/src/models/ticket_instances.rs
Timo614: Adjusting the SQL used by the release tickets logic to include the additional fields I was clearing here (the transfer and redeem related ones namely).

The other method `release_tickets` takes an `order_item_id` and a `quantity` so opted to just modify this `release` method to call the same SQL but pass in an optional `ticket_instance_id` to lock it down to the specific `TicketInstance`.

Timo614:
May be worth iterating it after the fact with a new issue if we want to split it apart there are probably a lot of gotchas there in terms of a background process that retries as we wouldn't want it rerunning transferring tickets back to organization and back over and over again due to the blockchain fees if the refund kept failing. Not sure what would be ideal there. 

From Zach's comments in the associated issue I'm not even sure if having partial payments makes sense (he made it sound like there would only be one method) so we may want to just do away with all the logic to support partial payments in favor of just one payment for the full cost of the order.

if we were to remove partial payments the logic for refunds would be simplified as we wouldn't need to calculate how much of the refund to give per payment method and then the only step that could fail and not be rolled back would be the refund itself with Stripe but that happens last so if that were to fail everything else could be rolled back without issue if there was only one method. 

We could / may want to figure out a different way to handle the refund (store credit? perhaps far off) if their credit card wasn't working for the refund for whatever reason. I'd imagine in most cases that would just be a Stripe blip and work on subsequent attempts though as if they managed a charge I'd imagine they can handle a refund back to it.

For now going to wrap up these changes by adding some transactions to the refund flow (which I should be able to push up in the next hour or so) to rollback elements of the refund if it fails but still log an attempted refund for the account. I think between that and the errors it would generate if it fails we'll have enough to debug from and can always iterate and add more later if we see it's insufficient in practice once we have some flows using it.

api/tests/functional/users.rs
mikethetike: Maybe let's rename `OrderRead` to `OrderReadOwn` and `OrderDetailsRead` to `OrderRead`?

db/src/queries/reports/reports_transaction_details.sql
mikethetike: Ok cool. Just wanted to highlight it

mikethetike:
I'm happy to get this in and do the other cases as separate issues

api/tests/functional/users.rs
Timo614: Sounds good, will make that change.

PR #848, opened 2019-01-07
Add localized event time

Krakaw:
Adds a `localized_times` key to event responses. If the `Venue` has a `timezone` value it will format the `event_start`, `event_end` and `door_time` fields to the timezone for the venue.

### DIFF
--- a/db/src/models/events.rs
+++ b/db/src/models/events.rs
@@ -1,5 +1,7 @@
 use chrono::prelude::*;
 use chrono::Utc;
+use chrono_tz::Tz;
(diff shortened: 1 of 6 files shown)


db/src/models/events.rs
```diff
@@ -1,5 +1,7 @@
 use chrono::prelude::*;
 use chrono::Utc;
+use chrono_tz::Tz;
+//use chrono_tz::{Tz, UTC as UtcTz};
```
Timo614: Not a huge deal but may be worth nuking this line since it's not needed.

PR #849, opened 2019-01-07
Added more specific errors to ticket transfers

### LINKED ISSUE #837: Ticket transfer: Invalid email address / phone number should respond with error
## Background

Even sending an empty `email_or_phone` to the ticket transfer endpoint results in a `200` response.

## What should happen instead

An HTTP error response code with a validation error JSON payload.

## Why

On the one hand, the mobile app could validate the email address or phone number using its own code. On the other hand, if the API and mobile app ever diverge, and the set of rules in the API becomes stricter, we'll end up with a mobile app that says "Tickets transferred!" when that's not true and the transfer was discarded.

### LINKED ISSUE #847: Ticket Transfers - Transfer tickets behaving odd "Business process error"

Krakaw:
Closes #837 
Closes #847 

Return the error cause to the client for 7000 errors
Do some simple validation on send ticket destinations and abort if there
are problems


api/src/controllers/tickets.rs
```diff
@@ -136,6 +137,8 @@ pub fn send_via_email_or_phone(
             .unwrap_or(604_800) as u32,
         connection,
     )?;
+    let re = Regex::new(r"[^0-9\+]+").unwrap();
+    let numbers_only = re.replace_all(&send_tickets_request.email_or_phone, "");
```
Timo614: If someone ends up putting letters in for their phone number it may be worth just triggering the error condition below instead of stripping them out. I'd think it'd be a pretty rare case either way though so probably not worth adjusting. Just thinking if someone were to put in bad data there it's unlikely they understood it was for their phone. Either way unlikely to happen in practice.

PR #861, opened 2019-01-10
Added transfer_pending bool to ticket display

### LINKED ISSUE #856: Add pending transfer info to tickets index response data
For mobile, in order to be able to show ticket transfers in a separate tab, we need the API response to tell us that the ticket is pending transfer.

(It would probably also make sense for the API to prevent transferring the same ticket multiple times, since doing so eventually leads to the "Business process error" from #847. That said, if we know a ticket has a pending transfer, then the mobile app won't let the user transfer it again.)

Krakaw:
Closes #856 
One thing that I would like to confirm, when the ticket is transferred to the receiver, the `transfer_key` and `transfer_expiry_date` were not being reset, allowing someone else to attempt to redeem it. During the redeem phase I now set those back to null so the new owner can transfer.

Also added a domain event for transfer started and transfer completed.

### DIFF
--- a/db/src/models/ticket_instances.rs
+++ b/db/src/models/ticket_instances.rs
@@ -820,7 +820,7 @@ impl From<DisplayTicketIntermediary> for DisplayTicket {
             ticket_type_id: ticket_intermediary.ticket_type_id,
             ticket_type_name: ticket_intermediary.name.clone(),
             status: ticket_intermediary.status.clone(),
-            pending_transfer: ticket_intermediary.pending_transfer.clone(),
+            pending_transfer: ticket_intermediary.pending_transfer,
             redeem_key,
         }
     }
--- a/db/src/models/ticket_instances.rs
+++ b/db/src/models/ticket_instances.rs
@@ -80,8 +80,8 @@ impl TicketInstance {
                 sql::<Bool>(
                     "CAST(CASE WHEN
                             ticket_instances.transfer_key IS NULL
-                            AND (ticket_instances.transfer_expiry_date IS NULL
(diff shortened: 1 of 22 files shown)


db/src/models/ticket_instances.rs
```diff
@@ -764,6 +820,7 @@ impl From<DisplayTicketIntermediary> for DisplayTicket {
             ticket_type_id: ticket_intermediary.ticket_type_id,
             ticket_type_name: ticket_intermediary.name.clone(),
             status: ticket_intermediary.status.clone(),
+            pending_transfer: ticket_intermediary.pending_transfer.clone(),
```
mikethetike: Looking at it, I don't think this needs a .clone(), but might be different via the compiler

db/src/models/ticket_instances.rs
```diff
@@ -77,6 +77,18 @@ impl TicketInstance {
                 ticket_instances::status,
                 ticket_instances::redeem_key,
                 events::redeem_date,
+                sql::<Bool>(
+                    "CAST(CASE WHEN
+                            ticket_instances.transfer_key IS NULL
+                            AND (ticket_instances.transfer_expiry_date IS NULL
```
Timo614: Should this be `OR`? If there's a `transfer_key` set and it has expired it should still be `pending_transfer` false I'd think since they can't use it to transfer then so would need to initiate another transfer.

Krakaw: Good spot.

db/src/models/ticket_instances.rs
Krakaw: You are correct.

PR #862, opened 2019-01-10
Add getter for redemption codes

mikethetike:
Should solve https://github.com/revelrylabs/bn-mobile-react/issues/410

Create a new route: `/redemption_codes/{code}` to list the details of a hold/comp for the front end to display


api/src/controllers/cart.rs
```diff
@@ -74,25 +71,6 @@ pub fn update_cart(
         })
         .collect();
 
-    if redemption_code.is_some() {
```
Timo614: Should we remove the additional `redemption_code` field from the `UpdateCartRequest` struct since we removed this and it only uses the individual `UpdateOrderItem` redemption code fields now?

From what I gather this logic used to use that redemption code and if it wasn't already added by one of the `UpdateOrderItem` elements it would add its max per order value to the cart? Was that used to add comps or something before?

mikethetike: Good point

mikethetike: When we were coming up to the release last year, we added a simple way of adding comps/holds to a cart, which was just adding a redemption code to the update_cart call. It would then add the ticket types that were available, but it will probably lead to confusion in future, so now you have to get the ticket types via the new api call and add them individually with the redemption code on the order_items

PR #863, opened 2019-01-10
Add tracking keys to public event responses

### LINKED ISSUE #859: Add tracking keys to event responses

Krakaw:
Closes #859 
This PR adds a `tracking_keys` field to event responses so that individual event requests can be tracked using the organization's tracking keys (if they have been added)

### DIFF
--- a/api/src/controllers/events.rs
+++ b/api/src/controllers/events.rs
@@ -247,7 +247,7 @@ pub struct EventParameters {
 }
 
 pub fn show(
-    (state, connection, parameters, query, user, http_request): (
+    (state, connection, parameters, query, user): (
         State<AppState>,
         Connection,
         Path<PathParameters>,
--- a/api/src/controllers/events.rs
+++ b/api/src/controllers/events.rs
@@ -84,7 +84,12 @@ pub fn checkins(
 }
 
 pub fn index(
-    (connection, query, auth_user): (Connection, Query<SearchParameters>, OptionalUser),
+    (state, connection, query, auth_user): (
+        State<AppState>,
+        Connection,
+        Query<SearchParameters>,
+        OptionalUser,
+    ),
 ) -> Result<HttpResponse, BigNeonError> {
     let connection = connection.get();
     let query = query.into_inner();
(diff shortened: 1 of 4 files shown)


api/src/controllers/events.rs
```diff
@@ -183,11 +199,19 @@ pub fn index(
     let results = events.into_iter().fold(Vec::new(), |mut results, event| {
         let venue = event.venue_id.and_then(|v| Some(venue_map[&v].clone()));
         let localized_times = event.get_all_localized_times(&venue);
+        let organization_id = event.organization_id;
+        let tracking_keys = tracking_keys_for_orgs
+            .get(&organization_id)
+            .map(|v| v)
```
Timo614: Think we can remove the map here since it's just mapping it to the same value.

api/src/controllers/events.rs
```diff
@@ -183,11 +199,19 @@ pub fn index(
...
+            .unwrap_or(&TrackingKeys {
+                ..Default::default()
+            })
+            .clone();
         results.push(EventVenueEntry {
             venue,
             id: event.id,
             name: event.name,
-            organization_id: event.organization_id,
+            organization_id: organization_id.clone(),
```
Timo614: Not a huge deal either way but I think you could probably remove the `clone()` here and just use `organization_id` since it'll fall back on the `Copy` trait for `Uuid`. I may be mistaken though but I think that should work.

api/src/controllers/events.rs
```diff
@@ -308,6 +334,18 @@ pub fn show(
         }
     }
 
+    let tracking_keys = Organization::tracking_keys_for_ids(
+        vec![organization.id],
+        &state.config.api_keys_encryption_key,
+        connection,
+    )?
+    .get(&organization.id)
+    .map(|v| v)
```
Timo614: Same thing with the map here I think we can remove it.

api/src/controllers/events.rs
Krakaw: Done.

api/src/controllers/events.rs
Krakaw: Yup, thanks

api/src/controllers/events.rs
Krakaw: 👍

Issue #879, opened 2019-01-16
Published events that have not hit their publish date, should not show up in the events search


Krakaw:
If you publish an event with a publish date in the future, it shows up as an event you can buy tickets to.

PR #883, opened 2019-01-16
Filter future published events from event search

Timo614:
### Closes Issues:
Closes: #879 

### Description:

This branch simply filters out published events that have not hit their publish date from the events search. Logged in organization admins still see these events but normal users and those logged out no longer see them in their search results.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Did you want me to do make any modifications here? Wasn't sure from the review above if you were checking on something. Or are you wondering about the generated SQL from the diesel filters?

It scopes it at the individual filter level as if it was in parenthesis from my testing so the `OR` clauses don't leak to the other filters. Will see if I can find some way to output the generated SQL and post it here (I believe there's some way).

mikethetike:
Ok cool, checked it myself as well and it's correct. 

Just wanted to make sure as diesel does not always put the brackets in the correct place:

For example the lines around 873 are wrong:
image

Without the `sql("(")....`, the following is generated:
image

as opposed to:
image

Issue #880, opened 2019-01-15
Event List - Events should not be considered a past event until the event End Time, not the event Start Time


zmelamed:
Event moved to past prematurely. It seems that the event time (rather than date) triggers this function.

An event was moved to 'past' mode as soon as the event show time took place. This has sales, check-in, reporting, onsite event management implications.

I also set an extended event end time (next day, 24 hrs after original event time)

In both scenarios, I saw:

-Event moved to 'Past' in event listing
-Event not available for sale via box office
-Tickets not able to be checked 
 
Date (rather than time) should designate 'past' status

PR #884, opened 2019-01-16
Use event end for event past related logic

Timo614:
### Closes Issues:
Closes: #880

### Description:

This branch attempts to fix the associated issue by using `event_end` in place of `event_start` for logic that's trying to limit data based on things being in the past. I updated a few different locations I found while searching so I think I have them all covered. Let me know if there was an endpoint I missed though and can adjust it further.

Questions:
- Should we make `event_start` and `event_end` required at the time of publishing? Several of the methods seem to expect them there. Event search previously only required `event_start` with a value but now with this change only those with both `event_start` and `event_end` appear.

### Migrations
 * `db/migrations/20190117145558_set_event_end_for_events`

### Environment Variables
 * No change


Timo614:
@Krakaw `event_start` isn't required at the moment either. I think some of that is for the purposes of making the events prior to publishing them. I guess we could add a validation that checks if the event isn't a draft status and if so makes sure both of them are set (should cover Offline events then if someone is updating those).

Will add the migration you mentioned for the ones with event start now:
`UPDATE events SET event_end = event_start + interval '1 day' WHERE event_end IS NULL and event_start IS NOT NULL;`

Will file another issue for making them required when not a draft if that sounds good.

Issue #881, opened 2019-01-15
Event Management - Ticket Type Deletion - Deleting a ticket type should remove unsold tickets from the total event capacity, reports and all dashboard views.


zmelamed:
Right now deleting a ticket type marks it as cancelled, but the total qty still shows up in the total capacity stats, reports and the ticket type inventory levels card still shows up in the various dashboard views

- If a ticket type is deleted, and no tickets had yet been sold from that ticket type, it should be completely hidden from the interfaces and reports
- If a ticket type is deleted, but tickets had been sold, the reports should only show the sold inventory stats, and the unsold ticket counts should not be included anywhere

Timo614:
So if we're cancelling these we want to `nullify` the tickets and on the blockchain similar to when we lower the capacity of a ticket type right? I noticed when I took a quick look that we only nullify available tickets. 

How do we want the logic to behave in the event some of these tickets are in the cart at the time of cancelling? They would become available again as soon as the reservation ended so they could end up back in inventory again.

For the partial payments logic where a partial payment can reserve tickets for up to a day would we want to trigger a refund on any partially paid orders if we plan on removing them above? Wondering how the cart would show this to the user.

One idea could be to mark the tickets as nullified and then have the cart fail to checkout if it's attempted. Perhaps having some additional text in the cart show logic to include a flag that the order cannot be purchased with some information as to why and the individual order items that need to be remove marked? The same sort of issue occurs if codes are no longer in their active period (we have start and end dates for discount codes) and the user wants to checkout. Perhaps we should have another issue to add some sort of invalidate cart messaging and flags to let the user know (thinking similar to how Amazon lets you know if a price has changed since you last added it to the cart but in this case that you can no longer purchase a given item).

Timo614:
![screen shot 2019-01-17 at 4 18 29 pm](https://user-images.githubusercontent.com/1319304/51349500-94b59480-1a73-11e9-8aed-cd8776fbd161.png)

I adjusted the logic to nullify the available tickets. This graph shows an example of what's left for an actively sold one with holds. I had the logic recursively release the hold's tickets and set their effective quantity to the unsold count prior to nullifying the remaining inventory.

49 of the tickets were reserved and the 30 held were all already purchased (should we make an issue for that? double counting them in the sold and held figures held only checks if they have a `hold_id` and doesn't exclude ones in the `Purchased` or `Redeemed` state.

Timo614:
[...]

PR #888, opened 2019-01-18
Adjust cancel behavior around remaining tickets

Timo614:
### Closes Issues:
Closes: #881

### Description:

This branch fixes the associated issue by marking unused tickets as nullified similar to the behavior that occurs when you update a `ticket_type` to have less quantity than it currently has. I adjusted the reporting logic to exclude the ticket types with no non nullified tickets as well.

Ticket types with no sold orders fall off the dashboard page once cancelled (aside from the details where they are marked as cancelled but do not appear in the listing, capacity, or reports). Ones that had sold tickets have only their unsold marked as `Nullified` and the other ones still appear in the reports and dashboard numbers.

May be worth having someone try the branch with their copy of bn-web and confirm I didn't miss any other places this may appear. I spent a little while looking so I think I covered all of the areas in the tooling this shows up though.

Open questions (mentioned in associated issue but persisting here):
- How do we want the logic to behave in the event some of these tickets are in the cart at the time of cancelling? They would become available again as soon as the reservation ended so they could end up back in inventory again. Right now I'm not doing anything with them (hence they appear in the `Open` number below).
- For the partial payments logic where a partial payment can reserve tickets for up to a day would we want to trigger a refund on any partially paid orders if we plan on removing them above? Wondering how the cart would show this to the user.

I noticed during this branch / maybe worth adjusting in other issues but:
- Holds can have their set quantity lowered even below the number of units currently reserved in the cart
- Holds do not currently respect the `end_at` value they have set
- Hold purchases appear in both the `Sold` and `Held` values in the dashboard which may be misleading since I'd think the sold should only be in the former once they have been purchased
- The revenue numbers appear off by $0.10 in the photo below

screen shot 2019-01-17 at 4 30 09 pm

You can see the latter here. There are 90 tickets left after 30 held and 60 regular. The 30 held are all purchased already. 11 of the regular are in a sold state and 49 are in a reserved state. The reserved state are shown as `Open` here. It looks a bit confusing on first glance without calculating those out.

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
@zmelamed please have a look at @Timo614 's comments regarding the holds

api/src/controllers/ticket_types.rs
```diff
@@ -167,17 +168,38 @@ pub fn index(
...
     let ticket_type = TicketType::find(path.ticket_type_id, connection)?;
-
     ticket_type.cancel(connection)?;
 
+    // Reduce holds to quantity sold
+    for hold in Hold::find_by_ticket_type(ticket_type.id, connection)?
+        .iter()
+        .filter(|h| h.parent_hold_id.is_none())
+    {
+        hold.remove_available_quantity(connection)?;
```
mikethetike: A hold can only belong to one ticket type, so we should delete the hold once it's empty

db/src/models/ticket_types.rs
```diff
@@ -269,6 +269,21 @@ impl TicketType {
         Ok(valid_ticket_count as u32)
     }
 
+    pub fn valid_unsold_ticket_count(&self, conn: &PgConnection) -> Result<u32, DatabaseError> {
+        let valid_unsold_ticket_count: i64 = ticket_instances::table
+            .inner_join(assets::table)
+            .filter(assets::ticket_type_id.eq(self.id))
+            .filter(ticket_instances::status.ne(TicketInstanceStatus::Nullified))
```
mikethetike: The next statement makes this line redundant. Did you intend a different column?

api/src/controllers/ticket_types.rs
```diff
@@ -167,17 +168,38 @@ pub fn index(
...
+
+    let valid_unsold_ticket_count = ticket_type.valid_unsold_ticket_count(connection)?;
+    nullify_tickets(
+        state,
+        organization,
+        ticket_type,
+        valid_unsold_ticket_count,
+        connection,
+    )?;
+
```
mikethetike: Should we not add a check to see that all of the ticket instances are now nullified?

api/src/controllers/ticket_types.rs
Timo614: Sounds good, will do that if there are no sold tickets for the hold.

api/src/controllers/ticket_types.rs
Timo614: We kind of do that already via the `nullify_tickets` method.

https://github.com/big-neon/bn-api/pull/888/files#diff-3cdd11b8c61ac0e1afaa3fa5020e1339R377

If the quantity we requested to be nullified does not match the quantity that was actually nullified the logic errors out reverting the change. We'd still want some to be in the `Purchased` and `Redeemed` states so the logic would be just confirming that the new `valid_unsold_ticket_count` was 0 so it effectively would work the same way as that check.

db/src/models/ticket_types.rs
Timo614: Nope, was a mistake as I iterated on this. Will remove the ne nullified line here.

Timo614:
> I think for tickets that are currently reserved in someone's cart, they should be cancelled. We must then add a check to see that all of the items in the cart 1. are still reserved (if we don't do this check already) and 2. have not been cancelled. If either are not true, the checkout should fail.

Sounds good for this card I'll modify the logic to also nullify reserved tickets and will make a note to make a follow up issue to check if a ticket is still reserved when in the cart and checking out. We may want to also modify the display cart logic to signal if there's a problem so a user can be alerted prior to payment so will make a note in that follow up issue for that.

api/src/controllers/ticket_types.rs
mikethetike: Ok cool

PR #895, opened 2019-01-18
#866 Globee Payments

### LINKED ISSUE #866: Globee Payments

mikethetike:
### Closes Issues:
Closes #866
 
### Description:
Adds a payment method for Globee. The main difference with Stripe is that it will return a url for the front end to redirect the customer to. Globee then sends IPN's to the api to inform it of status changes. 
We queue those IPNs as domain actions and then trigger (still in progress) the rest of the payment flow in the background.

### Migrations
 * 20190118094738_make_domain_action_main_table_optional
 * 20190121140415_add_checkout_url_to_order

### Environment Variables
 * GLOBEE_API_KEY
 * GLOBEE_BASE_URL
 * IPN_BASE_URL= (e.g.) "https://localhost:8088"


api/src/controllers/ipns.rs
```diff
@@ -0,0 +1,34 @@
...
+use errors::BigNeonError;
+use globee::GlobeeIpnRequest;
+use log::Level::Debug;
+
+pub fn globee(
+    (data, conn): (Json<GlobeeIpnRequest>, Connection),
+) -> Result<HttpResponse, BigNeonError> {
+    let data = data.into_inner();
+    jlog!(Debug, "Globee IPN received", { "data": &data });
+    let action = DomainAction::create(
```
sdbondi: Do we need to validate somehow that this request comes from Globee? Something like a message signature if Globee support that or a configurable shared secret which is put in the IPN URL?

api/src/controllers/cart.rs
```diff
@@ -460,97 +486,138 @@ fn checkout_payment_processor(
...
+
+                if save_payment_method {
+                    info!("CART: User has requested to save the payment method");
+                    match auth_user
+                        .user
+                        .payment_method(provider_name.to_string(), connection)
+                        .optional()?
+                    {
+                        Some(payment_method) => {
+                            let behavior = match client.behavior() {
```
Timo614: May be worth moving this check up to https://github.com/big-neon/bn-api/pull/895/files#diff-244164a04d7c56cd1efa02ca851b66edR527 as it's duplicated in both match arms (I think from the looks of it).

api/src/controllers/cart.rs
```diff
@@ -460,97 +486,138 @@ fn checkout_payment_processor(
...
+    payment_processor: &PaymentProcessor,
+) -> Result<HttpResponse, BigNeonError> {
+    let connection = conn.get();
     info!("CART: Auth'ing to payment provider");
     let auth_result = client.auth(
         &token,
         req.amount,
         currency,
-        "Tickets from Bigneon",
+        &format!("Big Neon"),
```
Timo614: Not a huge deal more just wondering, why the format here in place of just `"Big Neon"`? Wondering if you were using string interpolation and opted not to in the end?

api/src/controllers/ipns.rs
```diff
@@ -0,0 +1,32 @@
...
+    DomainAction::create(
+        None,
+        DomainActionTypes::PaymentProviderIPN,
+        None,
+        json!(data),
+        None,
+        None,
+        Utc::now().naive_utc(),
+        (Utc::now().naive_utc())
+            .checked_add_signed(Duration::days(30))
```
Timo614: Trying to wrap my head around the logic and just started looking it over so may be mistaken but does this give up to 30 days to process the payment? 

Wondering how we'll handle this from a show perspective as users purchasing day off may process beyond that of the concert so we'd either have to give them access to the event or refund them if they do not process in a timely manner. Wonder if we need some configurable cutoff date for Globee payments where we estimate the processing time and if there isn't enough time we don't allow the payment method to be used.  Not sure if people would be buying these day of via this method though. 

May be a use case for having store credit where the user makes a payment to just sit in their store credit wallet in advance so they don't end up waiting for a payment.

api/src/payments/globee/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+        PaymentProcessorBehavior::RedirectToPaymentPage(Box::new(GlobeePaymentProcessorBehavior {
+            client: GlobeeClient::new(self.api_key.clone(), self.base_url.clone()),
+        }))
+    }
+
+    fn refund(&self, _auth_token: &str) -> Result<ChargeAuthResult, PaymentProcessorError> {
+        unimplemented!()
+    }
+
+    fn partial_refund(
```
Timo614: Is this temporary pending some future logic to add store credit or reverse the crypto currency payment? Wondering if it's worth somehow flagging that the provider does not support refunds if not so the frontend knows to gray out that functionality.

api/src/controllers/ipns.rs
mikethetike: I've added some comments to make what happens more clear

api/src/payments/globee/mod.rs
mikethetike: Good point. This will currently break the frontend

Issue #877, opened 2019-01-16
Event Details - Displayed price ranges inconsistent related to listed tickets


zmelamed:
Shows $5-$15 but the ticket price on the event detail page is $10
<img width="1317" alt="big_neon_-_the_new_standard_in_event_ticketing" src="https://user-images.githubusercontent.com/456953/51226495-90b73480-1904-11e9-9149-571e7b0bf4c5.png">
<img width="1435" alt="vibe_vessel_tickets_on_big_neon" src="https://user-images.githubusercontent.com/456953/51226526-a4629b00-1904-11e9-943f-0a2a9c60bd5a.png">

Krakaw:
@zmelamed Do you want it to only show what ticket prices are only available as of right now, not what the total possible price range is?

PR #906, opened 2019-01-23
Adjust displayed pricing range for events

Timo614:
### Closes Issues:
Closes: #877

### Description:
This branch adjusts the associated issue by removing pricing caches in favor of calculating them at the time they are called. Previously ticket types would update the cache but it didn't take into account what pricing was active at that time. This logic just checks against the current ticket type pricing values and figures out the min and max pricing. I've added a test for this as well which demonstrates it calculating with no box pricing available (returns `null`), only pricing in the future (returns `null`), one price (returns same value for min and max), with box office pricing (when enabled), and with multiple pricing (shows the min and max as expected). 

screen shot 2019-01-22 at 8 08 12 pm
screen shot 2019-01-22 at 8 08 19 pm

### Migrations
 * 20190122212554_remove_pricing_cache_related_fields

### Environment Variables
 * No change


db/src/models/events.rs
```diff
@@ -295,6 +240,38 @@ impl Event {
...
+        conn: &PgConnection,
+    ) -> Result<(Option<i64>, Option<i64>), DatabaseError> {
+        let ticket_types = TicketType::find_by_event_id(self.id, conn)?;
+
+        let mut has_prices = false;
+        let mut min_ticket_price: i64 = std::i64::MAX;
+        let mut max_ticket_price: i64 = 0;
+        for ticket_type in ticket_types {
+            if let Some(ticket_pricing) = ticket_type
+                .current_ticket_pricing(box_office_pricing, conn)
```
sdbondi: `n+1` db queries here (m events * n ticket_types * k ticket_pricing for `GET /events`) - perhaps the approach could be doing aggregations in sql (`MIN` & ` MAX`) and (if it works) putting the results in the cache columns

Timo614: Sounds good re the n+1 query. I'll switch it to handle that as part of the event search SQL.

Timo614: I've adjusted the logic to remove the n+1 query. Now the logic performs one SQL query to fetch all of the ticket types active at the current date and then figures out the min / max for them.

I opted to not return it from the `Event#search` logic but to perform the additional query in the events controller to avoid needing to return a custom struct for the search results / and to avoid to extra convoluted logic that would be required to do it in one go.

sdbondi: Awesome - looks good

PR #907, opened 2019-01-23
Reuse Spotify access token

### LINKED ISSUE #583: Add caching to spotify auth key
The spotify endpoints create a new auth token on every request, this auth key needs to be cached in memory

sdbondi:
### Related Issues:
Closes #583 

### Description:

This PR keeps the Spotify access token in memory and reuses it until it expires (more accurately, a minute before). It does this by creating a static instance (+- a singleton) of the Spotify client which is shared between callers. Thread safety is achieved by using a `RwLock` on the access token which allows concurrent read access and single write access (Mutex). Access tokens are shared per process and therefore, each running instance will have their own access token.

Added `ApplicationErrorType::ServerConfigError` which should be used if for e.g. there is a missing environment variable. `WebError` will match this to a `500`. (If I'm using `ApplicationError` in an unintended way then we can remove this enum key as it's not strictly necessary for this PR).

### Migrations
 * No Migrations

### Environment Variables
 * No change


sdbondi:
> Looked through the code and it looks good to me, that `lazy_static` crate looks to be pretty useful.

Thanks 👍  - it was already a dep used in libraries we use so guess it's not a huge deal to make it a BigNeon dependency.

Issue #901, opened 2019-01-22
Refunded tickets are not being released back into the pool of available tickets


Krakaw:
@zmelamed please confirm if the following is correct.

If a ticket has been refunded before an event has taken place, it should be released back into either, the hold it came from, or the general pool of available tickets.
If the event is complete, the same can happen as it will show as unsold inventory.
If an event is cancelled and a refund is processed, it should be cancelled and not sent back to available inventory.

Timo614:
I'm seeing them get released locally when testing. The postman test flow has a case where it does this as well. If you remove the test steps after the refunds can see the associated tickets go back into the available inventory. 

Were these redeemed tickets? The only ones that aren't returned are the redeemed tickets as we don't want them to go back into the inventory and get redeemed again. Hold reserved tickets should be released as well like the ones in the postman test as there's nothing in the query logic to limit to just those outside of holds (may be worth us adding a test scenario though for it).

Will take this one and:
1. Add a test flow for refunding hold tickets
2. Adjust the refund logic to set the type to `Nullified` for tickets if the event is in the cancelled status when refunded

Timo614:
@Krakaw I confirmed that after refunding a ticket purchased via a hold I can add it back to my cart and it's set as reserved and linked to the new order item. Perhaps it's some bug with some other logic I'm not seeing though. I'll push up the additional tests I added here and the logic for setting the status to `Nullified` if it's a cancelled ticket type. 

Can you let me know if you still see the issue? Maybe something I'm missing or some flaw in the response for some API method that's making it look like it's not released when it is. I have been using smaller amounts via the API to test it but wondering if there's some bug showing up in some other way.

PR #924, opened 2019-01-24
Modify ticket release behavior when refunded

Timo614:
### Closes Issues:
Closes: #901

### Description:
I wasn't able to reproduce the issue mentioned in the associated issue so left a note there asking Keith if he could see if it's happening still (wondering if it's some bug with an API giving faulty numbers and not with the release behavior as I'm able to add those). 

This branch:
- Sets the refunded ticket to `Nullified` if the associated ticket type is `Cancelled` otherwise `Available` as it normally does
- Adds additional test logic with a functional test for refunding a held ticket (and then confirming it can be added to the cart), and some unit tests for new methods added to support the logic in the last bullet point

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
It looks like you have this solved, what I saw was a combination of the ticket numbers in the reporting being off and some of the refunded tickets going back into reserved status.

PR #929, opened 2019-01-25
Make venue timezone required

Timo614:
### Closes Issues:
Closes: #890

### Description:
This PR closes the associated issue by making timezones required on venues. For the purposes of making existing venues work with this migration I set an initial timezone of Los Angeles for any that lacked one. If we'd rather switch it to a different timezone for the migration let me know and can adjust.

It may be worth adding a validator in the future to confirm that it's a valid timezone but felt out of scope here since it was just to make it a required field for the purposes of this issue so only added a basic length validator that confirms it is at least one character at minimum in size.

### Migrations
 * 20190125172528_make_venue_timezone_required - sets existing venues lacking the timezone to Los Angeles and then makes the timezone column not null and adds a constraint confirming at least one character exists. 

### Environment Variables
 * No change


Timo614:
Sounds good, I had meant to run those but slipped my mind, adjusted them so should be set now once travis finishes.

PR #939, opened 2019-01-28
#927 Fee schedule constraints

### LINKED ISSUE #927: Fee schedule range constraints
#### Problem
If a ticket price falls into a range where no fee_schedule_range has been defined, the entire event errors out.

#### Proposed solution
Create a constraint that all values from 0 - greatest `min_price_in_cents` for the `fee_schedule_range` are covered.

Example:
Create a range from 0 - 1000 which has a 100 fee.
Create a range from 1500 - 2000 Which has a 150 fee, this will fail as there is nothing defined between 1000 - 1500.

The same is true for creating a fee from 1 - 1000, the range between 0 - 1 does not exists therefore it must fail

mikethetike:
### Closes Issues:
Closes #927 

### Description:
If the ticket price falls below the min fee schedule rank, use 0

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/tests/unit/models/user_display_ticket_type.rs
```diff
@@ -240,4 +241,36 @@ fn from_ticket_type() {
...
+            event.issuer_wallet(conn).unwrap().id,
+            None,
+            10,
+            0,
+            conn,
+        )
+        .unwrap();
+    let ticket_type = event.ticket_types(true, None, conn).unwrap().remove(0);
+
+    println!("{:?}", ticket_type);
```
Timo614: Left from debugging? Not a huge deal either way just saw in passing.

PR #941, opened 2019-01-29
Added migration status check to /status

### LINKED ISSUE #926: Add migration checker to the status endpoint
Compile in the latest migration required and compare it to the migrations that have been run. Return a 500 if the migrations for the deployment have not been run.

sdbondi:
### Related Issues:
Closes #926 

### Description:

`GET /status` not returns a `500` if there are any missing migrations, otherwise `200`. This is to prevent health checks from passing when a new version is deployed, which will prevent traffic being routed to the new instances until migrations have been run.

This is implemented by 'compiling in' the migration directory names (using a `proc-macro`) and comparing that to what has been run. Diesel does not currently publicly expose what is required to do something like this, but will [in the near future](https://github.com/diesel-rs/diesel/pull/1846), at which time the proc-macro can be removed.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/embed_dirs_derive/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
mikethetike: I think this crate should live one folder up

api/src/controllers/status.rs
```diff
@@ -0,0 +1,41 @@
...
+use db::Connection;
+use diesel::PgConnection;
+use errors::*;
+use log::Level::*;
+
+static mut IS_OK: bool = false;
+
+pub fn check(connection: (Connection)) -> Result<HttpResponse, BigNeonError> {
+    if unsafe { IS_OK } {
+        return Ok(HttpResponse::Ok().body("OK"));
```
mikethetike:
```suggestion
        return Ok(HttpResponse::Ok().finish());
```

db/src/utils/migration.rs
```diff
@@ -2,7 +2,7 @@ use diesel::{migration::RunMigrationsError, PgConnection};
 use diesel_migrations::MigrationConnection;
 
 #[derive(EmbedFileList)]
-#[embed_dir = "./db/migrations"]
+#[embed_dir = "./migrations"]
```
mikethetike: I think it would be better to use a path relative to the current file in the same way that the `include_str!` macro does it

sdbondi: Not too sure how `include_str!` does it - https://github.com/rust-lang/rust/blob/167ceff01ec2f01f677fa6351646255d3dacbb98/src/libstd/macros.rs#L757 - https://doc.rust-lang.org/std/macro.file.html could be an option (has to be used in calling file) but perhaps let's make this consistent later as this matches what diesel does with `embed_migrations!`

Issue #943, opened 2019-01-29
Events can be lost on admin page


Krakaw:
If you set an event start and an event end over a few days, the event shows up in neither upcoming or past events, effectively hiding the event from view until it completes.

Timo614:
Should we introduce another filter for ongoing events? Or could rename upcoming and just include the current and future events there.

Timo614:
For now going to add it to the upcoming response -- if it's worth breaking out into a separate response for just current events let me know but I figure given it's the admin section probably ideal to list the current along with the upcoming so the events in process and starting soon can be seen in one place.

PR #948, opened 2019-01-29
Include ongoing events in upcoming admin event list

Timo614:
### Closes Issues:
Closes: #943

### Description:
Fixes the associated issue by including ongoing events in the upcoming events list. Left a note in the other card but let me know if we want to make a new section for ongoing events or if this is the direction we want. I figure there won't be many ongoing events (aside from multi day festivals) really active at a given time so the upcoming list seems the most appropriate as it shows what's happening at a glance.

Improved the tests for the method in use here to show the behavior for upcoming and past events fetched.

Master:
screen shot 2019-01-29 at 3 37 41 pm

This branch:
screen shot 2019-01-29 at 3 30 49 pm

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/queries/find_all_events_for_organization.sql
```diff
@@ -53,7 +53,7 @@ FROM events e
 WHERE e.organization_id = $1
   AND CASE
         WHEN $2 IS NULL THEN TRUE -- All events
-        WHEN $2 THEN e.event_start >= now() -- upcoming
+        WHEN $2 THEN e.event_start >= now() OR e.event_end > now() -- upcoming
```
Timo614: I guess these checks could be simplified further by removing the `e.event_start >= now()` as the `event_end` condition should always cover that scenario. Will quickly do that.

Timo614: Actually will keep the check as it was -- we don't require the event_end date during creation so otherwise this may hide events until one is set.

mikethetike: Agree

db/tests/unit/events.rs
```diff
@@ -503,6 +484,82 @@ fn find_individuals() {
...
+    let organization = project.create_organization().finish();
+    let past_event = project
+        .create_event()
+        .with_name("Event1".into())
+        .with_event_start(
+            NaiveDateTime::parse_from_str("2014-03-05 12:00:00.000", "%Y-%m-%d %H:%M:%S%.f")
+                .unwrap(),
+        )
+        .with_event_end(
+            NaiveDateTime::parse_from_str("2014-03-06 12:00:00.000", "%Y-%m-%d %H:%M:%S%.f")
```
mikethetike: I've created a `times` module in `bigneon_db::dev` to help clean up these dates

PR #949, opened 2019-01-30
Verify Globee IPNs

mikethetike:
### Description:
Verify Globee IPNs when they come in.

I also removed the amount that is sent during checkout, instead calculating it from the order.

I also fixed the capacity and available fields on methods that return the TicketType so that they do not include `Nullified` tickets

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/ticket_types.rs
```diff
@@ -336,7 +341,13 @@ pub fn update(
...
     }
 
-    Ok(HttpResponse::Ok().finish())
+    let result = AdminDisplayTicketType::from_ticket_type(
+        &(TicketType::find(path.ticket_type_id, connection)?),
+        &FeeSchedule::find(fee_schedule_id, connection)?,
+        connection,
+    )?;
+
+    Ok(HttpResponse::Ok().json(json!(result)))
```
sdbondi: Is there a difference between `.json(json!(result))` and `.json(&result)` since `AdminDisplayTicketType` is serialisable?

PR #956, opened 2019-01-31
HTTP Caching on events

sdbondi:
### Related Issues:
None

### Description:

Cache-Control and ETag headers. Currently only implemented on `GET /events` or `GET /events/:id` but can easily be added where needed. When you want to generate an etag for a struct you can use the `#[derive(ToETag, Serialize)]` proc macro (serde `Serialize` is required to be implemented on the struct).

Current behaviour for cached events:

- Events will be cached for 60 seconds on proxies/CDN's only (`s-maxage`) unless the user is logged in. 
- Once the cache expires (after 60 seconds) the browser/cdn will re-request the content. If it hasn't changed (etag hash matches), the server will respond with with an _empty_ HTTP 304 (NotModified) 
 [details here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match#Specifications) and the CDN/browser will use their cache.  Otherwise, a normal 200 is returned.

This can be tweaked as necessary.

### Migrations
 * No Migrations

### Environment Variables
 * No change


sdbondi:
Removed caching headers from events endpoint so that it's safe to merge this - caching headers can be added later as needed

db/src/models/paging.rs
```diff
@@ -96,6 +98,34 @@ impl<T> Payload<T> {
...
+        let etag_strs = self
+            .data
+            .iter()
+            .map(|d| format!("{}", d.to_etag()))
+            .collect::<Vec<String>>()
+            .join("");
+
+        let tag_str = self
+            .paging
+            .tags
```
Timo614: Taking a look at this if we don't have the tags logic correctly populating all of the query string values this would likely use the same etag across multiple requests right? Is that why we shut it off for now? Since only a portion of the parameters end up as tags on the paging struct? Makes sense if so just confirming we'd need to fix the tag storage logic to take into account all of the struct's fields before we enable this.

https://github.com/big-neon/bn-api/blob/master/api/src/controllers/events.rs#L20..L38

https://github.com/big-neon/bn-api/blob/master/api/src/controllers/events.rs#L42..L61

http/caching_derive/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    let name_str = format!("{}", name);
+    let convert_to_etag_code = quote! {
+    {
+        let mut s = format!("{}{}", #name_str, json!(self));
+        ::bigneon_http::caching::etag_hash(&s)
+    }
+    };
+
+    let gen = quote! {
+    use bigneon_http::caching::{ETag, ToETag, EntityTag};
```
Timo614: Is there some benefit to having the `use` declaration here as opposed to the top of the file? We've kept them up there mostly (aside from some ones where we import ones likely to cause namespace collisions like schema related ones in the db models). Since these should be unique just wondering if there's a benefit down here versus at the top for consistency. Not a huge deal either way just more curious.

db/src/models/paging.rs
sdbondi: The `etag_strs` above takes all the values into account - as it takes the `SHA1(json)` of each struct (implementation dependent but the `ToETag` derive trait implements it this way.)

Now that I think about it. we may not even need to take the tags into account, since all the returned records are taken into account. If you search for "abc" and then "abcd" returns the same records then it would be fine for the etag to come out the same - the content hasn't changed.

http/caching_derive/src/lib.rs
sdbondi: We are "injecting code" here into the implementation - so this `use` statement doesn't live here it lives in the same file as the struct using the `#[derive(ToETag)]` proc/derive macro - notice `bigneon_http` is not a dependency of `caching_derive` (dev-dependency because it's used in tests)

sdbondi: So as if we are using bigneon_http::... in e.g. the events controller

sdbondi:
@Timo614 I'm removed it from `/events` because I wanted my PR to be merged with no risk :) Once this goes through, we can easily add http caching to endpoints in new PRs and each one can be evaluated - caching is tricky 🥇

db/src/models/paging.rs
sdbondi: Though the `"tags": { .... }` object would have changed - I assume the frontend uses it and needs it to be up to date

PR #966, opened 2019-02-01
Return redeem_key if event start within 24 hours

sdbondi:
### Closes Issues:
Closes #960 

### Description:

- If there's no event_start date the redeem key is not returned
- If it's more than 24 hours before the event start date, the redeem key is not returned

### Migrations
 * No Migrations

### Environment Variables
 * No change

### DIFF
--- a/db/src/models/ticket_instances.rs
+++ b/db/src/models/ticket_instances.rs
@@ -645,8 +645,9 @@ impl TicketInstance {
             .first::<RedeemableTicket>(conn)
             .to_db_error(ErrorCode::QueryError, "Unable to load ticket")?;
 
-        if ticket_data.redeem_date.is_some()
-            && ticket_data.redeem_date.unwrap() > Utc::now().naive_utc()
+        // Remove redeem key from results unless it's 24 hours before event start
+        if ticket_data.event_start.is_none()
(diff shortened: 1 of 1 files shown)


db/src/models/ticket_instances.rs
```diff
@@ -645,8 +645,9 @@ impl TicketInstance {
             .first::<RedeemableTicket>(conn)
             .to_db_error(ErrorCode::QueryError, "Unable to load ticket")?;
 
-        if ticket_data.redeem_date.is_some()
-            && ticket_data.redeem_date.unwrap() > Utc::now().naive_utc()
+        // Remove redeem key from results unless it's 24 hours before event start
+        if ticket_data.event_start.is_none()
+            || ticket_data.event_start.unwrap() - Duration::hours(24) > Utc::now().naive_utc()
```
Krakaw: The redeem_date could be set before 24 hours for some reason, so this would just be a temporary catch to make sure the redeem key is available at least 24 hours before

sdbondi: Ah thanks - Changed so that it uses either the redeem date, or 24 hours before the event start whichever is earliest - no redeem_date = 24 hours before the event

Issue #935, opened 2019-01-28
Add whether refunds are supported to API call  for orders so that the frontend can disable the refund button


mikethetike:

mikethetike:
The Globee payment processor struct does not support refunds at the moment, so the frontend should disable the refund button for orders that have used it.

PR #980, opened 2019-02-04
Added refund_supported field to the /events/{id}/guests endpoint

philipr-za:
### Closes Issues:
Closes https://github.com/big-neon/bn-api/issues/935

### Description:
Added an extra field to the /events/{id}/guests endpoint that indicated whether the payment provider used to pay for the ticket in the guest list allows for refunds or not. This can be used by the front end. 

This was done in this endpoint because this output is used to populate the Refund page in the frontend.

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
Code is OK

db/src/queries/retrieve_guest_list.sql
```diff
@@ -31,6 +32,7 @@ FROM ticket_instances ti
        INNER JOIN users u ON coalesce(o.on_behalf_of_user_id, w.user_id) = u.id
        INNER JOIN events e ON t2.event_id = e.id
        INNER JOIN venues v ON e.venue_id = v.id
+       LEFT JOIN payments p ON o.id = p.order_id
```
Timo614: I think this may cause dupe rows to be returned for orders with multiple tickets where some have been refunded as the refund logic adds a negative payment (so if you were to add all the payments together it zeroes out).

Since we only allow one payment per order may be safe to just have it distinct the records as they'd always match.

philipr-za: Eish, you are right. I will try a different approach.

Krakaw: I've used this approach for this problem:
```
LEFT JOIN (SELECT order_id, ARRAY_TO_STRING(ARRAY_AGG(DISTINCT p.payment_method), ', ') AS payment_method FROM payments p GROUP BY p.payment_method, p.order_id) AS p on orders.id = p.order_id
```

philipr-za: Ahh nice, my SQL-fu is not that great so I resorted to Rust post-processing

api/src/payments/payment_processor.rs
```diff
@@ -72,3 +72,12 @@ pub struct RedirectInfo {
     pub id: String,
     pub expires_at: NaiveDateTime,
 }
+
+pub fn is_refund_supported(provider: String) -> bool {
```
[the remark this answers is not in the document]
philipr-za: ServiceLocator is the perfect spot for it, moved it there.

db/src/models/events.rs
```diff
@@ -1232,3 +1268,9 @@ pub struct DayStats {
     pub revenue_in_cents: i64,
     pub ticket_sales: i64,
 }
+
+#[derive(Clone, Debug, Serialize)]
+pub struct GuestPaymentProviders {
```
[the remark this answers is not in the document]
philipr-za: Agreed, done

PR #981, opened 2019-02-04
Add additional domain events for order actions

### LINKED ISSUE #931: Add DomainEvents for Orders
As an order changes need to log these events:

1. Created
2. Completed (With ticket_instance_ids stored)
3. Updated

Timo614:
### Closes Issues:
Closes #931

### Description:
Branch simply adds some additional domain events for order updates and pipes the current user id down into the methods that trigger them. I've added some additional test logic to show the events getting created where we had existing tests. 

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/orders.rs
```diff
@@ -1243,6 +1296,20 @@ impl Order {
             for item in &order_items {
                 TicketInstance::mark_as_purchased(item, self.user_id, conn)?;
             }
+
+            let ticket_ids: Vec<Uuid> = TicketInstance::find_for_order(self.id, conn)?
```
Timo614: Actually thinking this over -- going to make a quick edit in the morning to just have a custom method return the ticket ids as the order may be really large and there's no point putting these in memory just to fetch out their ids. Shouldn't take too long so will adjust first thing.

PR #991, opened 2019-02-07
Globee fixes with new redirect urls

mikethetike:
### Closes Issues:
Closes big-neon/bn-web#1072

### Description:
This adds a GET method that will be called on Success and Cancel from Globee's payment page. This allows us to update the status of the order back to draft if cancelled. A Nonce (perhaps should be called secret?) is also added to the url to prevent this being called manually, but it doesn't remove this possibility entirely

### Migrations
 * `20190206204219_add_url_nonce_to_payments`

### Environment Variables
 * `IPN_BASE_URL` to be renamed to `API_BASE_URL`


db/src/models/orders.rs
```diff
@@ -1336,7 +1364,24 @@ impl Order {
...
+                Utc::now().naive_utc(),
+                // Technically this IPN must be processed and should never expire
+                (Utc::now().naive_utc())
+                    .checked_add_signed(Duration::days(3))
+                    .unwrap(),
+                3,
+            )
+            .commit(conn)?;
+        };
+        jlog!(Debug, "Order was checked for completion by was short", {"required_amount": total_required, "total_paid": total_paid, "order_id": self.id});
```
Krakaw:
```suggestion
        jlog!(Debug, "Order was checked for completion but was short", {"required_amount": total_required, "total_paid": total_paid, "order_id": self.id});
```

db/migrations/20190206204219_add_url_nonce_to_payments/down.sql
```diff
@@ -0,0 +1,2 @@
+alter table payments
```
Krakaw:
```suggestion
ALTER TABLE payments
```

db/migrations/20190206204219_add_url_nonce_to_payments/down.sql
```diff
@@ -0,0 +1,2 @@
...
+    drop column url_nonce
```
Krakaw:
```suggestion
    DROP COLUMN url_nonce;
```

db/migrations/20190206204219_add_url_nonce_to_payments/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table payments
```
Krakaw:
```suggestion
ALTER TABLE payments
```

db/migrations/20190206204219_add_url_nonce_to_payments/up.sql
```diff
@@ -0,0 +1,2 @@
...
+    add url_nonce text
```
Krakaw:
```suggestion
    ADD url_nonce TEXT;
```

Issue #989, opened 2019-02-06
Reporting - Org Level - Reconciliation Report


philipr-za:
Blocks https://github.com/big-neon/bn-web/issues/767

API implementation of report described in:
https://github.com/big-neon/bn-web/issues/767

PR #995, opened 2019-02-07
Add Reconciliation reports API implementation

philipr-za:
### Closes Issues:
Closes https://github.com/big-neon/bn-api/issues/989

### Description:
Implemented the Reconciliation Summary and Detailed report options for the /reports/{id}

For the Detailed report it was required that the organization_id be added to the fee_schedules table so that old fee_schedules can be associated with the organization that they were created for.

### Migrations
20190211125037_add_organization_id_to_fee_schedules

### Environment Variables
 * No change


mikethetike:
Note: We'll have to manually fix the data in production after this is deployed. I'd suggest we just delete and recreate the staging environments when we deploy it there

PR #1009, opened 2019-02-09
Updated fees calculation

Krakaw:
### Closes Issues:
Fixes the fees calculation error when there are no per event fees and the the calc was error'ing on a null value. Had to COALESCE all the fee row values.
### Migrations
 * 20190209073214_fix_fees_in_sales_report @sdbondi @Krakaw @mikethetike 

### Environment Variables
 * No change


mikethetike:
Looks good to me

PR #1015, opened 2019-02-11
Allow the creation of private events 

### LINKED ISSUE #982: Add private boolean to events
Either a boolean or a code that allows you to publish an event but it doesn't show up on the public index endpoints. You must have the event_id or the code (if we go that route) to view the event.

SWvheerden:
### Closes Issues:
fixes #982 
### Description:
This hides events that are private and allows the user to add give an access code to view the event
### Migrations
add private code to events table

### Environment Variables
 * No change


Cargo.lock
```diff
@@ -1,3 +1,5 @@
+# This file is automatically @generated by Cargo.
```
Krakaw: Don't commit this file.
Use
```bash
git update-index --assume-unchanged ./Cargo.lock 
```

db/src/models/events.rs
```diff
@@ -397,12 +399,26 @@ impl Event {
...
-        )
+    pub fn find(
+        id: Uuid,
+        user: Option<User>,
+        private_code: Option<String>,
+        conn: &PgConnection,
+    ) -> Result<Event, DatabaseError> {
+        let mut query = events::table.find(id).into_boxed();
+        match user {
+            Some(user) => {if !user.get_global_scopes().contains(&Scopes::OrgAdmin) {
```
Krakaw: This can be global scopes, but also organization scopes. Basically anyone that is part of the org.

api/src/controllers/events.rs
```diff
@@ -544,7 +561,7 @@ pub fn show_from_organizations(
 ) -> Result<WebPayload<EventSummaryResult>, BigNeonError> {
     let conn = connection.get();
     let org = Organization::find(path.id, conn)?;
-    user.requires_scope_for_organization(Scopes::RedeemTicket, &org, conn)?;
+    user.requires_scope_for_organization(Scopes::CanViewPrivate, &org, conn)?;
```
mikethetike: I don't think this is the correct scope. Perhaps `EventWrite || RedeemTicket` (For scanning or editing)

api/tests/functional/base/events.rs
```diff
@@ -104,7 +104,7 @@ pub fn show_box_office_pricing(role: Roles, should_test_succeed: bool) {
 
     if should_test_succeed {
         let event_expected_json =
-            expected_show_json(event, organization, venue, true, None, None, conn);
+            expected_show_json(event, organization, venue, true, None, None, conn, 1);
```
mikethetike: Is this change related to this PR? Seems to have snuck in

api/tests/functional/base/events.rs
```diff
@@ -690,6 +690,7 @@ pub fn expected_show_json(
     redemption_code: Option<String>,
     filter_ticket_type_ids: Option<Vec<Uuid>>,
     connection: &PgConnection,
+    intrested_users: u32,
```
mikethetike: spelling

db/src/models/events.rs
```diff
@@ -938,6 +941,8 @@ impl Event {
         match user {
             Some(user) => {
                 // Admin results include all drafts across organizations
+                println!("we have user: {:?}", user);
```
mikethetike: Debug code?

db/src/models/scopes.rs
```diff
@@ -14,6 +14,7 @@ pub enum Scopes {
     CodeWrite,
     CompRead,
     CompWrite,
+    CanViewPrivate,
```
mikethetike: We can probably infer this from the other scopes. If a user can edit events in an org, they can see private events. If they need to scan, they can see private events

api/tests/functional/base/events.rs
SWvheerden: Yes it is, I needed to change the function to allow the passing in the number of interested users.  This is so I can reuse the function.

api/src/controllers/events.rs
SWvheerden: Purposely changed the scope. Boxoffice does not have the right redeemticket.

db/src/models/scopes.rs
SWvheerden: This is the only shared right between boxoffice, doorman, org member.

db/src/models/events.rs
```diff
@@ -950,7 +957,7 @@ impl Event {
                                 .le(dsl::now.nullable())
                                 .or(events::status.ne(EventStatus::Published))
                                 .or(organization_users::user_id.eq(user.id)),
-                        );
```
sdbondi: `;` should be here

mikethetike:
@SWvheerden Just need to address the comments by @Krakaw and then this can go in

db/src/models/events.rs
```diff
@@ -170,6 +171,7 @@ pub struct EventEditableAttributes {
     #[serde(default, deserialize_with = "double_option_deserialize_unless_blank")]
     pub override_status: Option<Option<EventOverrideStatus>>,
     pub event_end: Option<NaiveDateTime>,
+    pub private_access_code: Option<String>,
```
mikethetike: This probably needs to be a double option to allow setting it to null. See `external_url`

SWvheerden:
@Krakaw 

On your comments,
All is done except the limits, should there be a limit? and if, what should it be

Krakaw:
> @Krakaw
> 
> On your comments,
> All is done except the limits, should there be a limit? and if, what should it be

Make it 6 chars [a-z0-9]{6} if you're lowercasing it to compare.

Issue #985, opened 2019-02-06
Downgrade stripe errors 'do not honor', 'incorrect cvc' and 'insufficient funds' to validation errors


mikethetike:
The main aim of this issue is to remove these from the error logs. The front end will need to handle these as 422 validation errors similar to other API calls.

@zmelamed, please confirm priority of this

PR #1021, opened 2019-02-12
Handle some stripe errors as validation errors

Timo614:
### Closes Issues:
Closes: #985

### Description:
This branch closes the associated issue by treating some Stripe errors as validation errors instead of internal errors. I adjusted the logic so the `StripeError` persists the `error_code` and the logic to convert that into a `PaymentProcessorError` displays some error validation text in the event those responses are returned. The status code 422 is returned in those cases with some additional validation text in the response body.

Note: It's not a complete fix for the desires of 985, however, as the current master's logic tends to log errors across multiple locations:
- https://github.com/actix/actix-web/blob/f383f618b537d6912ba9e2e7e27402b7bba964e1/src/pipeline.rs#L558 Active-web pipeline files a warning with the StripeError regardless of anything we do
- https://github.com/big-neon/bn-api/blob/master/api/src/middleware/big_neon_logger.rs#L48 BigNeon's `big_neon_logger` files an `Error` regardless if it finds an `error` on the response 

Can be seen via Stripe's test cards (which allow for some error flows):
https://stripe.com/docs/testing#cards

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
I think we would need to add some code to `controllers::cart::auth_then_complete()` to handle the result and return an Ok(HttpResponse) to avoid the big_neon_logger.

Happy for this to go in as it is now though

PR #1037, opened 2019-02-14
Promoter event access

mikethetike:
### Description:
Some extra routes to make inviting a user as a promoter to an event easier to use on the front end. Also added a more descriptive "OrgReadEvents" scopes that allows a user to get events for an organization. (Previously used Scopes::RedeemTicket)

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/scopes.rs
```diff
@@ -33,6 +33,7 @@ pub enum Scopes {
     OrgFans,
     OrgFinancialReports,
     OrgRead,
+    OrgReadEvents,
```
SWvheerden: This is probably going to clash with canviewprivate?
We should probably only keep one

PR #1038, opened 2019-02-15
Changed the dashboard results to face value

Krakaw:
### Closes Issues:
Closes big-neon/bn-web#1104
### Description:
Returns only ticket face values for dashboard
### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
> Not sure how this change fixes the problem, but the code itself looks fine

The dashboard only shows the face value of the tickets. So I've only included the item_type = 'Tickets' rows from the order_items table. That way no fees are making their way to the dashboard.

PR #1044, opened 2019-02-18
Delete promoter

mikethetike:
### Closes Issues:
Related to big-neon/bn-web#1117

### Description:
Add endpoints for deleting users/promoters from an event

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/events.rs
```diff
@@ -1053,3 +1053,45 @@ pub fn users(
...
     Ok(WebPayload::new(StatusCode::OK, payload))
 }
+
+#[derive(Deserialize)]
+pub struct EventUserPathParams {
+    id: Uuid,
+    user_id: Uuid,
+}
+
+pub fn remove_user(
```
Krakaw: Do we need a test for this?

Issue #1029, opened 2019-02-13
When a ticket_type has it's remaining available allocation in a hold it should only be shown to the client after adding an access code


Krakaw:
If you create a ticket type called "Media Passes" with a quantity of 100 and a price of $0. 
Then create a hold with 100 quantity. They should only show up to people after they have entered a code to "unlock" them.

Timo614:
Should we close this in favor of #1059 and using private for these ticket types? It's possible people will mistakenly set them to public still so there could be some value it's just more explicit with that approach. Wondering if I should close my PR for this (should be set just not sure if the other configuration options may accomplish the same thing).

Krakaw:
Closed in #1059

Krakaw:
Closed in #1059

PR #1050, opened 2019-02-19
[DO NOT MERGE] Hide ticket type when entirely allocated to hold

Timo614:
### Closes Issues:
Closes: #1029 

### Description:
This logic modifies the display logic for an event not include ticket types that take up the entire allocated set.

I also adjusted the logic that determines the price range to not include the price for the ticket types that are entirely allocated by holds (as those aren't visible so wouldn't make sense to see 25-50 and reach the page and only see 50 if the 25 are entirely in hold inventory).

Note: It would be possible to hide these ticket types now with some recent changes pushed up but would have to be set on each hold only ticket type so this issue seems worth doing. Holds tickets are not included in the quantity returned as available so a hold only ticket type is immediately in the sold out state and can trigger the configurable sold out state logic to be hidden if sold out:
https://github.com/big-neon/bn-api/pull/1047
https://github.com/big-neon/bn-api/pull/1048

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Not sure if this is still worth it with #1059 -- it does prevent accidental data leakages as it catches any ticket type setup for a holds only setup but anyone making a hold only ticket type is probably aware they want it to be private so it may just be added clutter to the logic as it's less explicit why it's not returned with this approach.

Timo614:
Confirming via associated issue if we still want this behavior or if the private ticket type approach would be more ideal for this use case. Will update the PR once I hear back and close if it's not needed.

Timo614:
Going to close this PR re message in associated issue to use #1059 for this feature. Will reopen if we have a need though.

Issue #990, opened 2019-02-06
Payments - Add additional meta data to stripe transactions


zmelamed:
Add as many of the following as meta data to orders in stripe as possible:
Event Name
Event ID
Event Date/Time
Venue Name
Customer Name
Number of Tickets
Face Value
Fees

Add this info retro-actively as well if possible

PR #1051, opened 2019-02-20
Stripe additional metadata

Timo614:
### Closes Issues:
Closes: #990

### Description:
This branch attempts to fix the associated issue by adding some additional metadata to Stripe charge. To facilitate this I just added a new metadata method on the `Order` model which can be updated in the future if we need to store different data here.

We should be able to reuse the script we're using here for the old entries to trigger a sitewide resync of metadata. I ended up having to make it its own CLI in the API project as this logic didn't really belong in the DB crate. I figure using this approach we can add different tasks we need to do on the API side over time as needed. The script should work fine as it has a little sleep in it to prevent us from banging away at Stripe. We will likely need to adjust the script as the size of payment records grows in the future but for the time being should be fine if we need to iterate on metadata and update it.

Wasn't sure if we could update metadata with Globee (the issue was for Stripe and I have little experience with Globee). For now I have the logic only update Stripe charges when running the update script.

Let me know if we should rename any of the fields. I went with plural names for the `event_names`, etc. as we theoretically could support a card with multiple events but I could go with singular if we prefer that. I know it's unlikely to happen so not sure the best call there so seemed safer to set the expectation it could be a comma separated list.

Example in Stripe:
screen shot 2019-02-19 at 9 08 22 pm

### Migrations
 * No Migrations

### Environment Variables
 * One change is to how you run the server locally. Previously we only had the one binary file so it defaulted to using that for cargo run. You need to specify it now, however like: `cargo run --bin server -- -t false`. There is a experimental cargo feature to allow the selection of a default binary so once that makes it into the release we should be able to revert back to the easier API but for now will need to specify.


Timo614:
I've added a script for updating the existing data. I couldn't find any way via Stripe to bulk update a set of properties across multiple charges so it iterates over each charge with a small sleep. It should work fine for now but someone with access will need to run it after it's live as I lack staging's credentials.

I did run it locally for testing after updating the event name used by my charge (debug logging so it shows the per charge):

```
../target/debug/api-cli sync-purchase-metadata
 {"level":"INFO","time":"[2019-02-20][18:15:17]","target":"api_cli","message":"Loading environment"}
 {"level":"INFO","time":"[2019-02-20][18:15:17]","target":"none","message":"Environment loaded"}
 {"level":"INFO","time":"[2019-02-20][18:15:17]","target":"api_cli","message":"Syncing purchase metadata"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:17]","target":"hyper::client::connect::dns","message":"resolving host=\"api.stripe.com\""}
 {"level":"DEBUG","time":"[2019-02-20][18:15:17]","target":"hyper::client::connect::http","message":"connecting to 54.187.208.163:443"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:17]","target":"hyper::proto::h1::io","message":"flushed 279 bytes"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:17]","target":"hyper::proto::h1::io","message":"flushed 392 bytes"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:18]","target":"hyper::proto::h1::io","message":"read 3007 bytes"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:18]","target":"hyper::proto::h1::io","message":"parsed 14 headers"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:18]","target":"hyper::proto::h1::conn","message":"incoming body is content-length (2388 bytes)"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:18]","target":"hyper::proto::h1::conn","message":"incoming body completed"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:18]","target":"hyper::client::pool","message":"pooling idle connection for \"https://api.stripe.com\""}
 {"level":"DEBUG","time":"[2019-02-20][18:15:18]","target":"reqwest::async_impl::response","message":"Response: '200 OK' for https://api.stripe.com/v1/charges/ch_1E5kHRJIAfNqtxWrL2NYjCBp"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:18]","target":"tokio_reactor","message":"dropping I/O source: 0"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:19]","target":"hyper::client::connect::dns","message":"resolving host=\"api.stripe.com\""}
 {"level":"DEBUG","time":"[2019-02-20][18:15:19]","target":"hyper::client::connect::http","message":"connecting to 54.187.208.163:443"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:19]","target":"hyper::proto::h1::io","message":"flushed 279 bytes"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:19]","target":"hyper::proto::h1::io","message":"flushed 390 bytes"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:20]","target":"hyper::proto::h1::io","message":"read 3036 bytes"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:20]","target":"hyper::proto::h1::io","message":"parsed 14 headers"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:20]","target":"hyper::proto::h1::conn","message":"incoming body is content-length (2417 bytes)"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:20]","target":"hyper::proto::h1::conn","message":"incoming body completed"}
 {"level":"DEBUG","time":"[2019-02-20][18:15:20]","target":"hyper::client::pool","message":"pooling idle connection for \"https://api.stripe.com\""}
 {"level":"DEBUG","time":"[2019-02-20][18:15:20]","target":"reqwest::async_impl::response","message":"Response: '200 OK' for https://api.stripe.com/v1/charges/ch_1E5kHbJIAfNqtxWrjJK0atYu"}
```

Old:
screen shot 2019-02-20 at 1 10 54 pm

With new event name:
screen shot 2019-02-20 at 1 15 55 pm

PR #1059, opened 2019-02-20
Add is private to ticket types

mikethetike:
### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
> Change looks good to me but we may want to also make a modification to the following file:
> [bn-api/db/src/models/events.rs](https://github.com/big-neon/bn-api/blob/953bbf6b196ef985892762aaf65110c3b93e085f/db/src/models/events.rs#L275)
> 
> Line 275 in [953bbf6](/big-neon/bn-api/commit/953bbf6b196ef985892762aaf65110c3b93e085f)
> 
>  pub fn ticket_pricing_range_by_events( 
> Just to filter out the private ticket types otherwise a private one with a lower/higher price could adjust the range we show to consumers.

Good point!

PR #1078, opened 2019-02-27
Add natterbot feedback

leet4tari:
### Description:
 Adds Drone CI notification 
### Migrations
 * No Migrations

### Environment Variables
 * No change

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -259,11 +259,6 @@ steps:
 
     - name: notify
       image: plugins/slack
-      depends_on:
-        - compile
-        - other-tests
-        - api-tests
-        - integration-tests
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -257,6 +257,21 @@ steps:
 #      when:
 #          status: failure
 
+    - name: notify
+      image: plugins/slack
+      depends_on:
+        - compile
```
sdbondi: The tests already depend on compile so can remove this line

leet4tari: Test is failing and notify step depends on tests, should we just remove the depends and add test status in the template?

PR #1084, opened 2019-02-28
Add organization specific Globee api key override

mikethetike:
### Migrations
 * 20190228145342_add_globee_api_key

### Environment Variables
 * No change

### DIFF
--- a/api/src/bin/api-cli.rs
+++ b/api/src/bin/api-cli.rs
@@ -51,10 +51,6 @@ fn sync_purchase_metadata(database: Database, service_locator: ServiceLocator) {
     let mut i = 0;
     let mut exit = false;
 
-    let stripe = service_locator
-        .create_payment_processor(PaymentProviders::Stripe)
-        .expect("Expected Stripe processor");
-
     loop {
         if exit {
             break;
@@ -74,6 +70,12 @@ fn sync_purchase_metadata(database: Database, service_locator: ServiceLocator) {
         }
 
         for (payment, order) in payments {
+            let organizations = order.organizations(connection.get()).unwrap();
+
+            let stripe = service_locator
+                .create_payment_processor(PaymentProviders::Stripe, &organizations[0])
+                .expect("Expected Stripe processor");
+
             if let Some(external_reference) = payment.external_reference {
                 let purchase_metadata = order
                     .purchase_metadata(connection.get())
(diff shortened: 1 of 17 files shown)


api/src/bin/api-cli.rs
```diff
@@ -74,6 +70,12 @@ fn sync_purchase_metadata(database: Database, service_locator: ServiceLocator) {
         }
 
         for (payment, order) in payments {
+            let organizations = order.organizations(connection.get()).unwrap();
+
+            let stripe = service_locator
+                .create_payment_processor(PaymentProviders::Stripe, &organizations[0])
```
Timo614: Should we make the organization argument optional and cache this? Would allow us to avoid having to make the organizations call on each order and since it isn't used for this payment provider could just cache that above the loop. Could then just error out in that create_payment_processor call if it's missing for Globee or something.

api/src/controllers/cart.rs
```diff
@@ -521,8 +521,19 @@ fn checkout_payment_processor(
             "Could not complete this cart because it is not in the correct status",
         );
     }
+    // Can only have one event at a time because there are potentially different
+    // payment gateway settings
+    let mut events = order.events(connection)?;
+    if events.len() != 1 {
```
Timo614: May be worth doing this check where we update the cart to avoid the cart getting into the odd state.

api/src/controllers/orders.rs
```diff
@@ -269,9 +269,15 @@ pub fn refund(
             let amount_to_refund = cmp::min(refund_due - amount_refunded, remaining_balance as u32);
             let mut refund_data = None;
             if payment.payment_method == PaymentMethods::CreditCard {
+
+                let mut organizations = order.organizations(connection)?;
+                if organizations.len() != 1 {
```
Timo614: https://github.com/big-neon/bn-api/blob/d55c7938904f44c87c6d817cbc17998f2f525a9c/db/src/queries/order_details.sql#L9 we may want to modify the listing logic to avoid it flagging these items as refundable. Could probably just count distinct event ids on the order_items table there and use that to mark it as un-refundable.

api/src/controllers/cart.rs
mikethetike: Good point, the UI flow at the moment makes it pretty hard to get into this state, but it's probably worth updating the API as well

mikethetike: I've created issue #1087 for this

api/src/controllers/orders.rs
mikethetike: I think we need to move to the point where all orders only have one event on them so that we don't need to do these checks everywhere. The main problem is that different organizations (and potentially events) have different payments rules and refunds become really difficult.

Issue #1054, opened 2019-02-20
Holds without codes


Krakaw:
Allow holds to have a nullable code, if the code is null, they cannot be added to a cart, but they can be split into a sub hold that does have a code.

Jasonvdb:
@Krakaw holds created from adding a name to a list need a code at least. When checking out a ticket for that hold, the only way to get the discount/comp is to pass in that code.

Krakaw:
Yes, you would not be able to access the tickets in the hold without a code.

Jasonvdb:
Related to https://github.com/big-neon/bn-api/issues/1070

Timo614:
@Krakaw @Jasonvdb is this intention here to ever sell the tickets that are part of the hold in this state? 

Is this just a means of setting aside a bunch of tickets that will be split off later into their own holds for redemption? Wondering if #1070 is needed or if the scenario it's setup for isn't something we need to support.

What case would we want to allow a box office person to purchase from a hold but wouldn't want to set a redemption code on it? I'd think we'd just split off x number of tickets for the box office if we ran into a situation like that and they'd have a code.

Jasonvdb:
@Krakaw can probably answer better. But I was imagining a scenario where names are added to a hold. Like a promotor gives a few comps to the media for the general access ticket type. He doesn't want anyone redeeming these online so he doesn't give a redemption code. Now the box office agent can't checkout those tickets when the media guys arrive because we are always assuming there'll be a code for box office.

Jasonvdb:
I guess one solution could be to make up a unique redemption code if one isn't supplied. 
What's happening right now in bn-web is tickets are being added to carts using a blank redemption code. Which then adds normal tickets and doesn't apply the discount/comp. So to the box office user, that behaviour is wrong.

mikethetike:
We could also create a hidden redemption code for the hold and not show it on the front end. In my mind that is the same as passing through a hold's id. Doing it this way will mean that we don't have to change any code

Timo614:
I pushed up what I had in progress here to make the column null: https://github.com/big-neon/bn-api/compare/master...Timo614:nullable-redemption-code-for-holds?expand=1 still needed to wrap it up and didn't include 1070 as I was waiting first on the feedback above.

[...]

Issue #480, opened 2018-10-30
Comp check-in at box office


Timo614:
> A manager of a "comp" hold adds a name to a list, and assigns it 4
   tickets, then they say to this person "Hey, your name is on the list, just
   mention your name at the box office or door and they'll check you in" --
   The order is created at the time of the ticket being "transfered to" (or
   claimed by) or, in the case that the box office person/doorman doesn't
   transfer it first and just immediately "checks in" those tickets from a
   comp list it should trigger that order as well... so the order trigger is
   either a checkin event or a transfer to a users account.

https://github.com/big-neon/bn-api/issues/422#issuecomment-432774151

Timo614:
This may not require any API work. Probably can just have the door person app logic trigger an order when checking the list and going through the normal comp order flow. Will leave open but I think we may be set.

PR #1086, opened 2019-03-01
Nullable redemption code for holds

Timo614:
### Closes Issues:
Closes: #1054 

### Description:
Closes the associated issue by introducing the capability to have holds that have no redemption code set. This updates the create and update endpoints to allow null for the redemption code and any associated code related to redemption codes that expected not null previously.

Tooling with codeless hold (went through and confirmed both the split and comps list logic worked for these):
screen shot 2019-02-28 at 7 02 26 pm

### Migrations
 * db/migrations/20190222143417_holds_nullable_redemption_code

### Environment Variables
 * No change


mikethetike:
I'm worried that we won't be able to sell these holds via Box Office. Is there a way that we can get this working?

Timo614:
@mikethetike mentioned via https://github.com/big-neon/bn-api/issues/1054#issuecomment-468469918

> With #480 we should be able to handle that case that @Jasonvdb mentioned as it should introduce a means to lookup a user by name, find their comp, and make their purchase. Redemption code would still be used we just would prevent the user from ever finding out.

So the redemption code is still used in a normal purchase by the box office agent it's just looked up via whatever we're going to provide in #480 to allow that functionality. As long as the redemption code is never sent to the end users it's private from them and it's still used in the redemption process so worth having set for that flow.

There is some value in having the nullable redemption code hold in the event an organization wanted to pre-allocate their entire ticket inventory in the system and then ration of their hold inventory via splits to other segments.

The alternative is we make the code `null` and we allow checkout via hold id but we still would need to lookup their hold via their name at the box office we'd just then use the id instead of the redemption code to make the purchase (complicating checkout but not by much).

mikethetike:
Fair enough, just wanted to check that was the intention

PR #1098, opened 2019-03-04
Add hold domain events

mikethetike:
### Description:
Adds more domain events for tracking changes to holds. Also changes Hold.delete from an actual SQL delete to changing the status to HoldStatus::Deleted

### Migrations
 * 20190304141623_add_status_to_holds

### Environment Variables
 * No change


db/src/models/holds.rs
```diff
@@ -299,52 +320,92 @@ impl Hold {
     /// or from the parent hold if `parent_hold_id` is not `None`. Likewise, if the
     /// quantity is lower, it will release the reserved tickets back to either the
     /// main pool or the parent hold.
-    pub fn set_quantity(&self, quantity: u32, conn: &PgConnection) -> Result<(), DatabaseError> {
+    pub fn set_quantity(
+        &self,
+        user_id: Option<Uuid>,
```
Timo614: More of an aside but I wonder if it's worth us refactoring all of the db calls to take like a request_context field (RequestContext struct) or something along those lines in place of the user_id and conn fields.

As we've started adding domain events for a lot of methods I've noticed we've had to pipe the user_id through a lot of random code. Thinking it'd keep the signatures cleaner if we did that. Could implement into for the `PgConnection` into the `RequestContext` for the basic calls so endpoints without user context won't need to be modified too heavily (ones with obviously will need something like `RequestContext::from_user(user, conn)`

Definitely not something we should address in this PR just starting a conversation about that.

mikethetike: Yeah I like it

PR #1102, opened 2019-03-05
Limits the maximum number of ticket instances per ticket_type

### LINKED ISSUE #1096: Create new ENV and per org field that limits the maximum number of ticket instances per ticket_type
[ ] Create a `MAX_INSTANCES_PER_TICKET_TYPE` env variable default to 10000
[ ] Add max_tickets_per_ticket_type to the `organizations` table default to MAX_INSTANCES_PER_TICKET_TYPE
[ ] Only system admins can edit max_tickets_per_ticket_type.
[ ] When creating or updating a ticket_type quantity it cannot go above max_tickets_per_ticket_type.

neonknight64:
- Added MAX_INSTANCES_PER_TICKET_TYPE env variable with 10000 default
- Added max_tickets_per_ticket_type to the organizations table with default 
- Modified organization create  and update so only system admins can edit max_tickets_per_ticket_type.
- Added checks in ticket_type create and update to check that the requested capacity is less than max_tickets_per_ticket_type for that organization.
- Added organisation functional test update_tests_with_max_tickets_per_ticket_type
- Added ticket_types functional test create_with_out_of_bounds_ticket_capacity

Closes #1096


api/src/config.rs
```diff
@@ -285,6 +293,7 @@ impl Config {
             api_keys_encryption_key,
             jwt_expiry_time,
             branch_io_branch_key,
+            max_instances_per_ticket_type,
```
Timo614: What is this configuration option used for? Guessing it may have been from an earlier iteration and unused now?

api/src/controllers/organizations.rs
```diff
@@ -179,8 +181,14 @@ pub fn update(
 ) -> Result<HttpResponse, BigNeonError> {
     let conn = connection.get();
     let mut organization = Organization::find(parameters.id, conn)?;
-    user.requires_scope_for_organization(Scopes::OrgWrite, &organization, conn)?;
     let organization_update = organization_parameters.into_inner();
+
+    if organization_update.max_tickets_per_ticket_type.is_some() {
```
Timo614: I wonder if it's worth checking to see if any ticket types would be below this new value for their capacity and then either returning an error or setting their capacity to the new maximum. Otherwise the next time they are updated for any reason they will fail due to https://github.com/big-neon/bn-api/pull/1102/files#diff-3cdd11b8c61ac0e1afaa3fa5020e1339R92 .

api/tests/functional/base/organizations.rs
```diff
@@ -256,6 +258,56 @@ pub fn update(role: Roles, should_succeed: bool) {
     assert_eq!(updated_organization.cc_fee_percent, 5.5);
 }
 
+pub fn update_with_max_tickets_per_ticket_type(role: Roles, should_succeed: bool) {
```
Timo614: May be worth just adding this field to the other update instead of a whole new test since it's a new field so probably not enough different to warrant checking against every role.

api/src/config.rs
neonknight64: This is a new mechanism to hard limit the number of tickets per ticket type. The default max_instances_per_ticket_type is specified as an Env variable, a system admin can adjust it and make it higher per organisation.

api/tests/functional/base/organizations.rs
neonknight64: Placing it all in one test function would have been great, but only a system admin is able to change the max_tickets_per_ticket_type thus the permissions required by the update function changes when this attribute is provided compared to the normal update. This test function tests two things: was the roles and permissions correctly enforced for this "special" attribute and if the attribute was correctly updated in the db.

api/src/controllers/organizations.rs
neonknight64: Cool, I will look at how I can implement this..

api/src/controllers/ticket_types.rs
```diff
@@ -249,6 +257,13 @@ pub fn update(
     jlog!(Debug, "Updating ticket type", {"ticket_type_id": path.ticket_type_id, "event_id":event.id, "request": &data});
     let ticket_type = TicketType::find(path.ticket_type_id, connection)?;
     if let Some(requested_capacity) = data.capacity {
+        //Check that requested ticket capacity is less than max_tickets_per_ticket_type
+        if requested_capacity as i64 > organization.max_tickets_per_ticket_type {
+            return application::internal_server_error(
```
mikethetike: It's better to return a validation error. An internal server error will appear in the logs as a error rather than a warning

api/tests/functional/base/organizations.rs
Timo614: Ah makes sense.

api/src/config.rs
neonknight64: Good catch, I made changes so when a new org is created and the max_instances_per_ticket_type is none then the default env value will be used.

api/src/controllers/ticket_types.rs
neonknight64: Changed to application::unprocessable error

api/src/controllers/organizations.rs
neonknight64: The db migration should add the default max_tickets_per_ticket_type to each existing org. Currently, I don't think there are any events with more tickets than this and all future events will have this constraint. Might not be worth it.

Timo614: I had a typo above I meant to write:

> I wonder if it's worth checking to see if any ticket types would be below above this new value for their capacity and then either returning an error or setting their capacity to the new maximum.

I was thinking more along the lines of:

- Joe creates organization with default limit 10,000
- Life goes on and admins use the tooling
- Ticket types are created under the limit but near it 8,500
- Joe realizes he set it too high so he goes to lower it setting it to 8,000
- We now have ticket types that are above capacity

Wondering if that's an acceptable state to be in (being over that set capacity or if it may be business rule based and something we want to enforce even on existing if we change it). 

I believe given the code if situation above happens a user can still edit a ticket type without errors so long as they don't include a new capacity field in their request. If the frontend does end up passing the existing value with the update it would block saving until the user modified it to a lower threshold.

mikethetike: The main intention of this issue is to prevent a user creating a large amount of ticket instance rows in the database, making our DB unnecessarily large. If there are already that many tickets in the DB, we can't really go back anyway

PR #1103, opened 2019-03-05
Add push notification and broadcast domain actions

mikethetike:
### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/utils/communication/mod.rs
```diff
@@ -134,21 +126,19 @@ impl Communication {
                     true => Either::A(future::ok(())), //Disable communication system when block_external_comms is true,
                     _ => {
                         let destination_addresses = communication.destinations.get();
-                        let source_address =
-                            communication.source.as_ref().unwrap().get_first().unwrap();
```
Timo614: Seemed a bit DRYer to have this stored here and reused below. Any reason to copy it to each use? Not the end of the world either way just curious.

PR #1110, opened 2019-03-06
Lock cart down to one event

Timo614:
### Closes Issues:
Closes: #1087

### Description:
Closes the associated issue by adding a validation to prevent users from adding more than one event's items to the cart at any given time. We prevent checkout in this state due to different organization needs for checkout customization so this just further locks it down that way.

If we plan on making this permanent we may want to add an event_id to the cart and update a lot of the logic in place which assumes multiple events. I assumed we're still on the fence about the longterm state of how we want the cart to function one versus many events so I left logic like a method called `events()` in place it just will always only return 1 event now. If we plan on making this permanent let me know and I'll spend some additional time refactoring logic.

I ended up having to do quite a bit of test pruning for this issue as we had a few for methods like order#organizations, order#events, and so on that expected multiple. If we do decide to later swing back and allow multiple events per cart we should revert some of the test modifications here.

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
Looks good. Might need to hold off merging this for a few days though

PR #1115, opened 2019-03-09
Admin functions

Krakaw:
### Closes Issues:
Added admin functions for ticket counts as well as stuck pending transactions.
These are only meant for admins of the system.
### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/admin.rs
```diff
@@ -0,0 +1,27 @@
...
+use bigneon_db::models::{DomainAction, Report, Scopes};
+use db::Connection;
+use errors::*;
+
+pub fn admin_ticket_count(
+    (connection, user): (Connection, AuthUser),
+) -> Result<HttpResponse, BigNeonError> {
+    let connection = connection.get();
+    //Check if they have org admin permissions
+    user.requires_scope(Scopes::OrgAdmin)?;
```
mikethetike: Perhaps a specific scope, e.g. `SystemAdmin`?

PR #1124, opened 2019-03-12
Transfer on blockchain direct ticket transfers

mikethetike:
### Description:
Adds code to transfer the tickets on the blockchain if they are transferred directly. 

Also adds a push notification for users if they receive tickets via email and they already exist

Note: This PR was originally for the push notification only, and has since been expanded to include the block chain transfer

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/tickets.rs
```diff
@@ -137,12 +140,33 @@ pub fn send_via_email_or_phone(
...
                 auth_user.id(),
                 &send_tickets_request.ticket_ids,
                 user.id,
                 connection,
             )?;
+
+            let receiver_wallet = user.default_wallet(connection)?;
+
+            // TODO change blockchain client to do transfers from multiple wallets at once
+            for (sender_wallet_id, tickets) in
```
Krakaw: Would it not make it slightly DRY'er if we moved this into the `transfer_tickets_on_blockchain` and the removed the `sender_wallet` arg. That way if we have a single `receiver_wallet` we can call a single function that will find all of the senders automatically (I guess not that is should be more than 1) and do the transfer in 1 place?

mikethetike: I looked into it and it's actually a bit haphazard how we get the wallets that tickets come from. We should actually send a single transaction to the block chain with the wallets/addresses/accounts that the tickets are in and also do the verification in `TicketInstances::received_ticket_transfer` 
I've created #1128 to address this in future

Issue #1112, opened 2019-03-08
Epic - Implement unit tests in `/db` for sections that are not covered


mikethetike:
Using the coverage script in `/db/scripts` there are a number of files with less than 75% coverage. 

This issue should be used as a parent issue for creating smaller issues:

Some of the files that need covering:

  |   |   |   |  
-- | -- | -- | -- | --
Filename |  |   |  Line Coverage  | Functions
artists.rs |   |   | 67.6 % | 115 / 170 | 13.9 % | 22 / 158  
assets.rs |   |   | 68.4 % | 26 / 38 | 20.0 % | 6 / 30  
broadcasts.rs |   |   | 0.0 % | 0 / 83 | 0.0 % | 0 / 94  
codes.rs |   |   | 82.6 % | 147 / 178 | 23.5 % | 31 / 132  
domain_actions.rs |   |   | 26.2 % | 32 / 122 | 10.0 % | 8 / 80  
domain_events.rs |   |   | 71.9 % | 41 / 57 | 24.0 % | 12 / 50
enums.rs |   |   | 84.5 % | 49 / 58 | 10.0 % | 126 / 1261
event_artists.rs |   |   | 80.6 % | 58 / 72 | 12.1 % | 7 / 58
event_interest.rs |   |   | 86.7 % | 65 / 75 | 23.5 % | 8 / 34
events.rs |   |   | 69.0 % | 531 / 770 | 13.9 % | 70 / 504
external_logins.rs |   |   | 71.4 % | 30 / 42 | 14.3 % | 6 / 42
fans.rs |   |   | 0.0 % | 0 / 13 | 0.0 % | 0 / 2
fee_schedule_ranges.rs |   |   | 22.6 % | 7 / 31 | 8.3 % | 3 / 36
fee_schedules.rs |   |   | 84.7 % | 72 / 85 | 23.3 % | 7 / 30
for_display.rs |   |   | 93.0 % | 53 / 57 | 66.7 % | 10 / 15
history_item.rs |   |   | 0.0 % | 0 / 6 | 0.0 % | 0 / 14
holds.rs |   |   | 72.7 % | 192 / 264 | 22.7 % | 34 / 150
mod.rs |   |   | 100.0 % | 64 / 64 | 80.0 % | 8 / 10
order_items.rs |   |   | 84.2 % | 245 / 291 | 20.3 % | 32 / 158
orders.rs |   |   | 79.0 % | 774 / 980 | 28.0 % | 65 / 232
[...]

PR #1131, opened 2019-03-14
Added tests for domain actions model

sdbondi:
### Related Issues:
Ref #1112 

### Description:
Domain actions model coverage 80.3 %
`scheduled_at` for `NewDomainAction` now correctly updates `expires_at` and `blocked_until`

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/domain_actions.rs
```diff
@@ -256,7 +256,7 @@ impl NewDomainAction {
     }
 
     pub fn schedule_at(&mut self, schedule_at: NaiveDateTime) {
-        let diff = self.scheduled_at - schedule_at;
+        let diff = schedule_at - self.scheduled_at;
```
Krakaw: Do we always know which is larger?
```suggestion
        let diff = schedule_at - self.scheduled_at;
        let diff = diff.abs();
```

mikethetike: No, this is intentional. You may bring the scheduled time forward

Issue #1132, opened 2019-03-15
Replace Tari client errors with a more generic description


Krakaw:

Timo614:
Closing per #1142

PR #1142, opened 2019-03-19
Use a generic error for Tari client errors

Timo614:
### Closes Issues:
#1132

### Description:
Fixes the associated issue by returning a generic error message in place of more specific Tari client messaging when one occurs.

Was: `Internal error: Problem with the Tari client`
Now: `Internal error`

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/errors/web_error.rs
```diff
@@ -190,7 +190,7 @@ impl ConvertToWebError for SerdeError {
 impl ConvertToWebError for TariError {
     fn to_response(&self) -> HttpResponse {
         error!("Tari error: {}", self);
-        internal_error("Internal error: Problem with the Tari client")
+        internal_error("Internal error")
```
Krakaw: Sorry, I should have specified that it needs to be a friendlier error that doesn't say Tari error. Maybe something like
```suggestion
        internal_error("There was an error during the ticket transfer process")
```

Issue #1133, opened 2019-03-15
Add pagination and search to the Transaction Details report


Krakaw:
connects to big-neon/bigneon#54
required by big-neon/bn-api#1205

Issue #972, opened 2019-02-02
Function tests for reports


Krakaw:
Either in bn-api or in postman, but run through the purchasing and refunding and validate that the results we are getting in the reports are correct.

PR #1143, opened 2019-03-19
Search and pagination support for transaction details report:

Timo614:
### Closes Issues:
Closes: #1133
Relates to: #972 (adds functional tests for transaction details report)

### Description:
This PR fixes the associated issue by adding support for a query parameter and pagination. I've added unit and functional tests for this report as well. 

The branch checks the query against: `Should be able to search by name, email, order number, event name`
See https://github.com/big-neon/bn-web/issues/1207 https://github.com/big-neon/bn-web/issues/1205

Note: previously the transaction details report associated the user information with the `user_id` of the order regardless of if `on_behalf_of_user_id` was set or not. In this branch I changed the logic to use the on behalf of user where possible for these records. Let me know if we want me to revert this change but I can't think of a good reason why we'd want to show the box office user's details there over the user they bought it for so thinking it's the right call.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Jasonvdb:
@mikethetike can we get this ready to merge? bn-web almost done

Timo614:
Looks like the failing test here is that socket hang up error we occasionally run into during the integration tests. Should pass them if we restart the test build.

PR #1150, opened 2019-03-22
Added artist response to events index

Krakaw:
### Closes Issues:
Related to revelrylabs/bn-mobile-react#641
### Description:
Added the artists list to the search results
### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/events.rs
```diff
@@ -1139,7 +1140,7 @@ fn event_venues_from_events(
     let mut results: Vec<EventVenueEntry> = Vec::new();
     for event in events.into_iter() {
         let venue = event.venue_id.and_then(|v| Some(venue_map[&v].clone()));
-
+        let artists = EventArtist::find_all_from_event(event.id, connection)?;
```
mikethetike: Is this method event called anywhere from the front end? I thought it was replaced by the general event search with a venue_id?

Krakaw: It's called from other controller methods in events.rs.

PR #1153, opened 2019-03-25
Upgrade diesel to fix locking/open transactions on panic

### LINKED ISSUE #1114: Postgres locking issues
Related to #1113

mikethetike:
### Closes Issues:
Closes #1114 

### Description:
Upgrades diesel to a fork that does not add connections back to the connection pool if there is still an open transaction. The diesel PR is here: https://github.com/diesel-rs/diesel/pull/2020
When/If that PR gets merged, we can link back to the crates.io version, but until then we'll link to my fork

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
In your fork of diesel what happens to the connection that has the uncommited transaction with regards to returning it to the pool. 
Let's say you create a pool of 1
```rust
let pool = Pool::builder()
            .max_size(1);
```
and that one connection lands up with an uncommited transaction, does it destroy and recreate the connection and add it back to the pool so another process can use it?

mikethetike:
Yeah it checks if the connection is broken (having open transactions count as being broken) and then will not add it back to the pool. Then it will create a new one if it is below the min

PR #1162, opened 2019-03-27
Add unit test for payments in db

SWvheerden:
### Description:
created unit test for the payment models in db
### Migrations
 * No Migrations

### Environment Variables
 * No change


SWvheerden:
Just note that currently this PR points to PR #1161 
It needs to point to master before merging, this is just to avoid merge conflicts with the other one

PR #1178, opened 2019-04-04
Fix bug found while updating discount codes

philipr-za:
### Description:
Fixed the handling of updating a discount code and changing the type of discount it is (percentage vs in cents). 

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/codes.rs
```diff
@@ -370,7 +372,16 @@ impl Code {
         update_attrs: UpdateCodeAttributes,
         conn: &PgConnection,
     ) -> Result<Code, DatabaseError> {
+
+        let mut update_attrs= update_attrs;
+
+        if update_attrs.discount_in_cents.is_some() || update_attrs.discount_as_percentage.is_some() {
```
Krakaw: What happens if both are set to a value, or both are set to 0?

philipr-za: then there are no updates to the discount type, just needed to make sure that if one is updated then both need to be set properly.

Krakaw: Ahh right I see the constraint in the DB that'll catch that.

philipr-za: and the validator that is called right after that if statement, that is what fails if only one of the options is set.

PR #1181, opened 2019-04-04
Add availability to requests for discount codes

philipr-za:
### Description:
Adds an availability field to requests for discount codes for events.

### Migrations
 * No Migrations

### Environment Variables
 * No change


philipr-za:
Will incorporate this into a later PR.

Issue #1193, opened 2019-04-09
Promo codes - cart limit modifications


Timo614:
Connected to big-neon/bigneon#84

This issue will track the bn-api related modifications to address the problems mentioned in the associated bigneon issue.

PR #1194, opened 2019-04-09
Cart limit modifications

Timo614:
### Closes Issues:
Closes: #1193 
Related: big-neon/bigneon/issues/84

### Description:
This branch fixes the associated issue by making some modifications to the way limit checks are done in checkout in a few ways.

Previously we treated hold and code limits are overrides of the ticket type limits in the backend. For example: with the old behavior you could have a ticket type with a limit of 1 and a hold with a limit of 5 and manage to purchase 5 tickets for that hold. With the new behavior a limit of 1 would be enforced (https://github.com/big-neon/bigneon/issues/84#issuecomment-481256627).

Primary changes:
- Ticket type and hold / code limits are both checked now (with the former being done across all uses in holds / codes and outside)
- Adjusted the code limit check from a validation to one that happens as part of the update_quantities flow similar to that of the holds logic (it was a relic of my initial codes implementation 5 months ago and dated compared to how we handle the cart now)
- Adjusted the API logic to return the minimum of the hold/code and ticket type limit so that the frontend properly restricts quantity

### Examples

Ticket type 3 limit
ticket-type-3-limit

Ticket type 3 limit code 4 limit
ticket-type-3-limit-code-4-limit

Ticket type 5 limit
ticket-type-5-limit

Ticket type 5 limit code 4 limit
ticket-type-5-limit-code-4-limit

Ticket type no limit
ticket-type-no-limit

Ticket type no limit code 4 limit
ticket-type-no-limit-code-4-limit

### Migrations
 * 20190409173024_codes_remove_max_tickets_constraint - migration removes a db constraint and function in favor of the update_quantities approach for limit checking

### Environment Variables
 * No change


Timo614:
@zmelamed 

This should be mostly here. I was thinking things over further and was wondering:
- The intent is for the limit behavior (appending the limit not overriding it) to be like this for both codes and holds right? Or did you want holds to work like overrides for the limit? I made the modification here across both but was thinking perhaps we intended them to be separate in that regards. Can modify the behavior if so.
- Should we make a special case for comps? If you were to have a ticket type with a limit of 5, purchased 5 normally, and then went to activate your comp you'd be blocked due to that limit. May be a bit of an edge case so perhaps out of scope for the time being.

mikethetike:
@Timo614 RE: SQL functions: We've moved to having SQL functions in a separate file to the migrations, with the idea that every deploy we recreate the functions to
1. Force a recompile in case there are any errors
2. Make it easy to find the latest version
3. To track changes to functions via source control

Ideally I wanted to have one file per SQL function, but we didn't find a way to run them all easily, so we decided to put them in one file until we can work that out

Timo614:
@mikethetike the SQL function isn't in use any more once I removed the validation and the db constraint so I had removed it from that file.

The migration just removes the constraint from the table (in favor of the update_quantities limit check logic) and the function with a "down" to add it back just to reflect the reverse of its removal. I can remove adding it back from the "down" if we want just did that for completeness sake to reflect the initial migration which had added it. Not sure if we would want to keep it in the functions.sql though as it may get out of date with nothing using it.

mikethetike:
Perfect, sorry I didn't check it properly

PR #1198, opened 2019-04-10
Fix refunds on discounted items

philipr-za:
### Closes Issues:
Closes https://github.com/big-neon/bn-api/issues/1188

## Description:
Refunding items did not take into account if discounts were applied.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
As an aside, how were the EventFees handled on an order refund?

philipr-za:
Checks if there are any remaining order_items for that order_id where `refunded_quantity` < `quantity` (excluding EventFees and now Discounts) and if there are none then adds the EventFee to the `amount_to_be_refunded`.

PR #1200, opened 2019-04-11
Refactor SoldOutBehavior to Visibility

mikethetike:
### Closes Issues:
Child of big-neon/bigneon#88

## Release Details:
Replace "When sold out" and "private" on ticket types with a Visibility field with 
* Always - Ticket type is shown as "On sale soon", <Price>, "Sales ended", "Unavailable" (when only holds tickets are available" and "Sold out."
* Only when available - Shown unless sold out or there is no ticket pricing for the dates
* Hidden (Promo code/holds only) - Hidden except for when a promo code is entered

See issue for more detail 

### Migrations
 * 20190411120000_add_ticket_type_visibility

### Environment Variables
 * No change


db/src/models/ticket_types.rs
```diff
@@ -37,10 +37,9 @@ pub struct TicketType {
     updated_at: NaiveDateTime,
     pub price_in_cents: i64,
     pub cancelled_at: Option<NaiveDateTime>,
-    pub sold_out_behavior: SoldOutBehavior,
-    pub is_private: bool,
     pub parent_id: Option<Uuid>,
     pub rank: i32,
+    pub visibility: TicketTypeVisibility,
```
Timo614: Putting this here since it's more of a general question (not tied to this line of code but rather the file).

Is it worth us adding domain events for calls to `TicketType#update` that change the visibility? Just thinking since it affects the sales so heavily although I doubt in most cases people will modify this once it goes live so perhaps it's just noise.

db/src/test/builders/event_builder.rs
```diff
@@ -153,13 +173,23 @@ impl<'a> EventBuilder<'a> {
...
 
         if self.with_tickets {
-            let early_bird_start = NaiveDateTime::from(Utc::now().naive_utc() - Duration::days(2));
-            let early_bird_end = NaiveDateTime::from(Utc::now().naive_utc() - Duration::days(1));
-            let standard_start = NaiveDateTime::from(Utc::now().naive_utc() - Duration::days(1));
-            let standard_end = NaiveDateTime::from(Utc::now().naive_utc() + Duration::days(2));
-
-            let event_start = NaiveDateTime::from(Utc::now().naive_utc() + Duration::days(2));
-            let event_end = NaiveDateTime::from(Utc::now().naive_utc() + Duration::days(4));
+            let early_bird_start = NaiveDateTime::from(dates::now().add_days(-2).finish());
```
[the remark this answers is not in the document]
mikethetike: Good point

db/src/models/ticket_types.rs
mikethetike: I think there are domain events for TicketTypeUpdated. Were you thinking about maybe adding TicketTypeVisibilityUpdated specific events?

Timo614: That's true, I think the TicketTypeUpdated should cover it since we store all the attributes as part of the domain event record.

PR #1202, opened 2019-04-11
Fix start date on Tiered Pricing

mikethetike:
### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/ticket_types.rs
```diff
@@ -340,18 +340,18 @@ impl TicketType {
...
-                .set((
-                    ticket_types::start_date.eq(self.start_date),
-                    ticket_types::updated_at.eq(dsl::now),
-                ))
-                .execute(conn)
-                .to_db_error(
-                    ErrorCode::UpdateError,
-                    "Could not update start date on ticket type",
-                )?;
+            //            self.start_date = Some(times::now());
```
Timo614: Should we still have the logic bring the start date forward if they have a current start date of the future?

Otherwise we end up saying below that the sales have started but I don't think will end up starting until the existing date is reached.

db/src/models/ticket_types.rs
```diff
@@ -340,18 +340,18 @@ impl TicketType {
...
+            //                    ticket_types::updated_at.eq(dsl::now),
+            //                ))
+            //                .execute(conn)
+            //                .to_db_error(
+            //                    ErrorCode::UpdateError,
+            //                    "Could not update start date on ticket type",
+            //                )?;
 
             DomainEvent::create(
                 DomainEventTypes::TicketTypeSalesStarted,
```
Timo614: May be worth nuking the line below:
`Some(json!({"old_start_date": old_start_date, "new_start_date": self.start_date })),`

Since it should be always be the same here.

db/src/models/ticket_types.rs
Timo614: I think the only caller of this method is https://github.com/big-neon/bn-api/blob/a3f46ff42e62597ec67aeb64a2b2e1897084b3b3/db/src/models/ticket_types.rs#L318 which appears to only run this method for children without start dates so this shouldn't lead to any issues with changed behavior.

mikethetike: The TicketPricing row needs to have the correct start date (while the TicketType start_date is null) for a couple of scenarios:
1. The pricing is initially set to the Parent's end_date so that it automatically enables as active when the parent ends
2. Having the TicketType start_date and the parent_id set makes it difficult for the UI to know what to display to the user, and also to know what to pass back to the API on edits

By having the cart UI allow for a null ticket type start_date, these cases are handled. 

I'll remove the comments though, it's not temporary. I was being overly cautious.

PR #1204, opened 2019-04-12
Add report Sales by organization

mikethetike:
### Description:

Report 1 (For every customer in an Org):

customer name
customer email
total number of events they've purchased tix to
total number of tickets purchased
gross dollars spent
total number of tickets transfered

I've also split the SQL functions into individual files that are combined in `build.rs` when `db` is built.
### Migrations
 * No Migrations

### Environment Variables
 * No change


philipr-za:
Man, I should leave SQL to the pro's. Even in review.

db/functions/report_ticket_sales_by_organization.sql
```diff
@@ -0,0 +1,36 @@
...
+AS $$
+BEGIN
+    return query SELECT u.id,
+                        u.first_name,
+                        u.last_name,
+                        u.email,
+                        count(DISTINCT e.id)                                                               AS events,
+                        sum(CASE WHEN oi.item_type = 'Tickets' THEN 1 ELSE 0 END)                          AS tickets,
+                        sum(CASE WHEN oi.item_type = 'Tickets' AND ti.wallet_id != w.id THEN 1 ELSE 0 END) AS transferred,
+                        sum(oi.unit_price_in_cents * oi.quantity)                                          AS gross_spent
```
Krakaw: Because there will be a ticket_instance row that will always link back to this order_item row we don't need to multiply by the quantity
```suggestion
                        sum(oi.unit_price_in_cents)                                          AS gross_spent
```

db/functions/report_ticket_sales_by_organization.sql
```diff
@@ -0,0 +1,36 @@
...
+                 FROM orders o
+                          INNER JOIN order_items oi ON o.id = oi.order_id
+                          INNER JOIN events e ON e.id = oi.event_id
+                          INNER JOIN users u
+                          INNER JOIN wallets w ON u.id = w.user_id
+                                     ON o.user_id = u.id
+                          LEFT JOIN ticket_instances ti ON oi.id = ti.order_item_id
```
Krakaw:
```suggestion
                          LEFT JOIN ticket_instances ti ON oi.id = ti.order_item_id
                          LEFT JOIN refunded_tickets rt ON rt.order_item_id = oi.id AND ti.id = rt.ticket_instance_id
```

db/functions/report_ticket_sales_by_organization.sql
```diff
@@ -0,0 +1,36 @@
...
+
+                 WHERE e.organization_id = p_organization_id
+                   AND o.status = 'Paid'
+                   AND o.paid_at >= p_start_date
+                   AND o.paid_at < p_end_date
```
Krakaw:
```suggestion
                   AND o.paid_at < p_end_date
                   AND rt.id IS NULL
```

PR #1207, opened 2019-04-12
Fix bug with promo codes with the same redemption code

philipr-za:
### Closes Issues:
Closes https://github.com/big-neon/bigneon/issues/110
### Description:
Sometimes if two promo codes or holds have the same redemption code trouble was caused.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/holds.rs
```diff
@@ -456,16 +456,29 @@ impl Hold {
...
     ) -> Result<Hold, DatabaseError> {
-        holds::table
-            .filter(holds::redemption_code.eq(redemption_code.to_uppercase()))
-            .filter(holds::deleted_at.is_null())
-            .first(conn)
-            .to_db_error(
-                ErrorCode::QueryError,
-                "Could not load hold with that redeem key",
-            )
+        if let Some(e) = event_id {
```
Krakaw: Could this not be shortened to 
```rust
let mut query = holds::table
            .filter(holds::redemption_code.eq(redemption_code.to_uppercase()))
            .filter(holds::deleted_at.is_null())
            .into_boxed();
        if let Some(e) = event_id {
            query = query.filter(holds::event_id.eq(e))
        }

        query.first(conn).to_db_error(
            ErrorCode::QueryError,
            "Could not load hold with that redeem key",
        )
```

philipr-za: yes, will do

PR #1211, opened 2019-04-16
Add handling of credit card fees to orders and organization admin

philipr-za:
### Closes Issues:
Closes https://github.com/big-neon/bigneon/issues/64

### Description:
* Added accepting and updating the cc_fee_percentage field in organizations.
* Added adding a OrderItemType::CreditCardFee order item to an order if the organization's cc fee is non-zero.
* Handled the new OrderItemType in calculating totals and refunds.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/order_items.rs
```diff
@@ -195,6 +196,7 @@ impl OrderItem {
         if self.item_type == OrderItemTypes::PerUnitFees
```
Timo614: Not a huge deal but may be worth us updating this:
`if self.item_type != `OrderItemTypes ::Tickets`

db/src/models/orders.rs
```diff
@@ -1134,6 +1140,25 @@ impl Order {
                     new_event_fee.commit(conn)?;
                     per_event_fees_included.insert(event_id, true);
                 }
+                // Credit card fees are set per organization
+                let org = Organization::find(event.organization_id, conn)?;
```
Timo614: Just confirming, we add this regardless of the payment type? If the organization supports Globee and Stripe they have the fee across any orders?

philipr-za: Good question. @Krakaw what do you think about this?

db/src/models/order_items.rs
philipr-za: Yes, that is indeed much neater.

db/src/models/orders.rs
mikethetike: @zmelamed

db/src/models/orders.rs
```diff
@@ -1134,6 +1140,25 @@ impl Order {
...
+                if org.cc_fee_percent > 0f32 {
+                    let cc_fee = (self.calculate_total(conn)? as f32
+                        * (org.cc_fee_percent / 100f32))
```
mikethetike: I wonder if it's not better to store the `cc_fee_percent` as a 3-5 fixed point i32 in the db rather than a float.
e.g. `350` for `3.5%`

philipr-za: Happy, to do that. I was just using was just using what was already in the db. Though you will not still have to deal with floating point math when you apply the percentage?

mikethetike: I didn't realise the field already existed. Let's work with it as is

db/src/models/orders.rs
zmelamed: Thanks for bringing this up, I'm clarifying with our business team when/if we should expect agreements to have separate payment processor fees per service provider or if the singular fee for all payment methods for which we collect and settle the funds is all we'll need at least for the foreseeable future.

If separate fees by payment methods are needed in the future we will break out a separate issue and prioritize it accordingly. 

I know for the contracts we currently have in place a single fee is satisfactory, and I know that having multiple fees per payment method makes this a much larger consumer facing UX project that will impact the order of operations within the checkout flow so you should not go down that path at this time.

PR #1219, opened 2019-04-23
Add campaign tracking data

mikethetike:
### Closes Issues:
Connected to big-neon/bigneon#49

### Description:
Adds fields for tracking campaign data. The frontend must extract URL parameters and add `tracking_data` to the cart methods  

### Migrations
 * 20190418121122_add_tracking_data_to_order

### Environment Variables
 * No change


db/migrations/20190418121122_add_tracking_data_to_order/down.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE orders
+    DROP tracking_data;
+ALTER TABLE orders
```
Krakaw: Purely a style thing, can this not be done in a single `ALTER TABLE orders`?

db/migrations/20190418121122_add_tracking_data_to_order/up.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE orders
```
Krakaw: Same as above, purely a style thing, can this not be done in a single `ALTER TABLE orders`?

db/migrations/20190418121122_add_tracking_data_to_order/down.sql
mikethetike: You can do that?

Krakaw: `ALTER TABLE table DROP COLUMN col1, DROP COLUMN col2;`

PR #1230, opened 2019-04-29
Add sales_start_time to GET /events/{id}

philipr-za:
### Closes Issues:
Connected to https://github.com/big-neon/bigneon/issues/117

### Description:
Added a field to the GET /events/{id} endpoint that reports the earliest sale start time of all the ticket types. This will be used to set the starting time on the Transaction Detailed Report date filter.

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/events.rs
```diff
@@ -270,8 +271,14 @@ pub fn show(
         connection,
     )?;
     let mut display_ticket_types = Vec::new();
+    let mut sales_start_date = Some(times::infinity());
     for ticket_type in ticket_types {
```
Krakaw: Are these ticket_types populated with _all_ available tickets even if they are not available for public consumption?

philipr-za: It did take into account Access Codes but I saw that I needed to move the IF statement below the check for visibility via holds.

PR #1234, opened 2019-04-30
Allow box office refunds

philipr-za:
### Closes Issues:
Closes https://github.com/big-neon/bigneon/issues/116
### Description:
Allow for refunds of tickets sold via the box office. In the end was a simple change to the function that checks if a ticket has been transferred before refunding. Now the check will include checking if the ticket was bought on behalf of another user (the way Box Office does) and let the refund go forward.

All the rest of the refund logic is appropriate. Actual funds are only paid if its a Credit Card payment so the refund is logged and the emails are sent as per usual.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/ticket_instances.rs
```diff
@@ -103,7 +104,7 @@ impl TicketInstance {
             )
             .filter(ticket_instances::id.eq(self.id))
             .select(
-                sql::<Bool>("case when order_items.id is null then true else orders.user_id <> wallets.user_id end")
+                sql::<Bool>("case when order_items.id is null then true else (orders.user_id <> wallets.user_id AND (orders.on_behalf_of_user_id IS NULL OR wallets.user_id <> orders.on_behalf_of_user_id)) end")
```
Krakaw: Just some case cleanup, otherwise good find.
```suggestion
                sql::<Bool>("CASE WHEN order_items.id IS NULL THEN TRUE ELSE (orders.user_id <> wallets.user_id AND (orders.on_behalf_of_user_id IS NULL OR wallets.user_id <> orders.on_behalf_of_user_id)) END")
```

Issue #1222, opened 2019-04-25
[Needs Investigation] Updating user profile triggers unique key constraint


Krakaw:
We just need to check that if a user is updating their user profile, and they are _not_ changing their email address but have posted it along with the rest of their profile data, we must not trigger the unique key constraint.

![image](https://user-images.githubusercontent.com/41575888/56706454-546e4e00-6715-11e9-90d7-5a452392c223.png)

Timo614:
I tried to reproduce this but I can update my user keeping my email address the same. I did however notice that if I try to switch my user email to one in user it triggers the database error which in turn returns the `users_email_key` causing the "Duplicate record exists" error on the frontend.

We could handle this better by having the code do a validation check instead to see if the email already exists and in that case just return it on the email field like we do for other validations.

Will quickly patch a validation for that to make the experience nicer.

My guess is the user in question already has another user account and tried to update this other account to use the same email the other is using.

PR #1244, opened 2019-05-02
Improve error messaging around email already being in use for profile update

Timo614:
### Closes Issues:
Closes: #1222 

### Description:
I couldn't reproduce the associated issue but I did notice I ran into it if I tried to change my user's email to one that was already in use by another in the system. This pull request just improves the user experience on this flow to make the error more informative and prevent the index violation from firing. Previously it was a benign situation (unless I'm missing some other issue here) it just clogged the logs a bit with junk.

Note: we may eventually want to put a captcha on the profile update page as someone could scan the system for what emails have accounts by just trying to change their email and looking for ones that fail. May also be worth having some sort of email confirmation flow where the user has to approve the email.

Screen Shot 2019-05-02 at 1 25 05 PM

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/users.rs
```diff
@@ -506,16 +508,58 @@ impl User {
...
 
+    fn email_unique(
+        id: Uuid,
+        email: String,
+        conn: &PgConnection,
+    ) -> Result<Result<(), ValidationError>, DatabaseError> {
+        let email_in_use = select(exists(
+            users::table
+                .filter(users::id.ne(id))
+                .filter(users::email.eq(email)),
```
mikethetike: We should probably to a `to_lower` or equivalent here

Timo614: Oops, definitely should. Thanks for catching that, will quickly address and adjust the tests to add that case.

Timo614: It may actually be fine since the logic in the update method lowercases the email attribute prior to the validation (from the find_by_email it looks like we always lowercase it prior to entry so that should be sufficient). 

I'm going to add an explicit to_lowercase to it though so if we ever call this method elsewhere we make sure the validation always is properly checking that value.

PR #1248, opened 2019-05-06
Adjust transfer expired error messaging

Timo614:
### Closes Issues:
Closes: #1247 

### Description:
Fixes the associated issue by checking if the transfer is expired and if so returns, `{"error": "Business process error: The transfer has expired."}`

Let me know if we want to adjust the error. Previously it used to say, `{"error": "Business process error: These tickets have already been transferred."}` because the logic confirmed the total number of transferred tickets equalled the expected and in the case of the expired transfers it did not.

## Release Details:
Added better message for when a transfer fails due to expiring
### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/ticket_instances.rs
```diff
@@ -955,13 +955,31 @@ impl TicketInstance {
                 Some("ECDSA Signature is not valid".to_string()),
             ));
         }
+        let transfer_is_expired: bool = select(exists(
```
Timo614: This is currently required because each transfers record is tied back to the ticket instance in a denormalized way. It's not ideal and is an artifact of the transfers model having been designed around the existing implementation. I filed #1252 to improve the data model there. Once it's set we can simple fetch the transfer record by itself and check an `is_expired` method or the likes.

Sorry about that, I should have thought more about the data model at the time I was adjusting the existing system.

mikethetike: I think put this message as comments in the actual code so we can pick it up later.

api/tests/functional/tickets.rs
```diff
@@ -597,3 +599,93 @@ fn receive_ticket_transfer() {
...
+        auth_user2.clone(),
+        request.extract_state(),
+    ))
+    .into();
+
+    assert_eq!(response.status(), StatusCode::UNPROCESSABLE_ENTITY);
+    let body = support::unwrap_body_to_string(&response).unwrap();
+    assert_eq!(
+        body,
+        json!({"error": "Business process error: The transfer has expired."}).to_string()
```
mikethetike: Is there a way to remove `Business process error: ` from the error message. The front end displays this message directly to the user

Timo614: Yeah, this is just something we've historically done via the error mapping logic (https://github.com/big-neon/bn-api/blob/master/api/src/errors/web_error.rs#L225). Will remove that so those errors aren't prefixed with it.

PR #1254, opened 2019-05-07
Update refund amounts to use signed integers to prevent possible underflows

mikethetike:
### Description:
I renamed/extracted some methods to make this code read a bit easier. This is obviously subjective so feel free to let me know if I've made it more complicated.

The main thing I noticed was that we had restricted the amounts to use u32. I think this could have been my doing, but I think in future all monetary amounts should be signed (i64) so that we don't run into an underflow and accidentally send an email with a refund amount around 43 million.

## Release Details:
Removed potential underflow on refunds. This situation is not possible due to other constraints in the system but is an extra safeguard against future bugs

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
Just waiting for a review by @Timo614

PR #1255, opened 2019-05-07
[HIGH RISK] Add locking of orders for IPNs

mikethetike:
### Description:
This PR adds locking in two places: 
1. When an IPN is received
2. When the browser is redirected on success (from the payment provider). 

In the IPN case, if the order can't be locked because another process has it, it will fail and retry via the scheduler process. 

In the case of the redirect, the process tries 5 times to get the lock. This is to prevent the user from dropping out of the sales process due to an error

## Release Details:
Added better concurrent handling of IPNS coming in at the same time

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
@mikethetike I'm actually noticing one oddity in this branch though. For some reason after doing an IPN I am unable to `control + c` out of the process as it seems stuck. Seems like the backend workers may be getting into a stuck state.

Confirming now (and will edit here) if it happens if I don't hit the IPN callback at all (just booting, hitting the web frontend a bit, and trying to `control + c` it).

mikethetike:
Thanks @Timo614. It does sound like what we are getting in production, so perhaps the source of the error is something different. 

This query should return 0 rows. Please try it on your side:
```sql
SELECT blocked_locks.pid     AS blocked_pid,
       blocked_activity.usename  AS blocked_user,
       blocking_locks.pid     AS blocking_pid,
       blocking_activity.usename AS blocking_user,
       blocked_activity.query    AS blocked_statement,
       blocking_activity.query   AS current_statement_in_blocking_process,
       (select relname from pg_class where oid = blocking_locks.relation) as rel,
       blocking_locks.*
  FROM  pg_catalog.pg_locks         blocked_locks
         JOIN pg_catalog.pg_stat_activity blocked_activity  ON blocked_activity.pid = blocked_locks.pid
         JOIN pg_catalog.pg_locks         blocking_locks
         ON blocking_locks.locktype = blocked_locks.locktype
           AND blocking_locks.DATABASE IS NOT DISTINCT FROM blocked_locks.DATABASE
           AND blocking_locks.relation IS NOT DISTINCT FROM blocked_locks.relation
           AND blocking_locks.page IS NOT DISTINCT FROM blocked_locks.page
           AND blocking_locks.tuple IS NOT DISTINCT FROM blocked_locks.tuple
           AND blocking_locks.virtualxid IS NOT DISTINCT FROM blocked_locks.virtualxid
           AND blocking_locks.transactionid IS NOT DISTINCT FROM blocked_locks.transactionid
           AND blocking_locks.classid IS NOT DISTINCT FROM blocked_locks.classid
           AND blocking_locks.objid IS NOT DISTINCT FROM blocked_locks.objid
           AND blocking_locks.objsubid IS NOT DISTINCT FROM blocked_locks.objsubid
           AND blocking_locks.pid != blocked_locks.pid

         JOIN pg_catalog.pg_stat_activity blocking_activity ON blocking_activity.pid = blocking_locks.pid
  WHERE NOT blocked_locks.GRANTED;
```

Timo614:
Getting back:

```
       76371 | postgres     |        76367 | postgres      | UPDATE "orders" SET "version" = $1, "updated_at" = CURRENT_TIMESTAMP WHERE "orders"."id" = $2 AND "orders"."version" = $3 | SELECT "orders"."id", "orders"."user_id", "orders"."status", "orders"."order_type", "orders"."order_date", "orders"."expires_at", "orders"."version", "orders"."note", "orders"."on_behalf_of_user_id", "orders"."created_at", "orders"."updated_at", "orders"."paid_at", "orders"."box_office_pricing", "orders"."checkout_url", "orders"."checkout_url_expires", "orders"."create_user_agent", "orders"."purchase_user_agent", "orders"."external_payment_type", "orders"."tracking_data", "orders"."source", "orders"."campaign", "orders"."medium", "orders"."term", "orders"."content", "orders"."platform" FROM "orders" WHERE "orders"."id" = $1 |     | transactionid |          |          |      |       |            |        849021 |         |       |          | 5/46               | 76367 | ExclusiveLock | t       | f
```

Via that query you posted above.

mikethetike:
I added a quick/temporary fix for this. Basically moved the call to "execute" on the action. What was happening was that the DomainAction would run `perform_job` on each action from the database, but not call the `commit` or `rollback` until it had read all the actions from the database. Now, by moving the execute call out of the `find_actions` method, each job is run and the transaction completed before moving onto the next.

I think that our logic and attempt to use futures here is overzealous. We should really just be reading an action, running `perform_job` and completing it, each in its own thread.

api/src/domain_events/domain_action_monitor.rs
```diff
@@ -213,8 +212,8 @@ impl DomainActionMonitor {
             let command = command.unwrap();
 
             per_action_connection.begin_transaction()?;
-            let f = command.execute(action, per_action_connection);
-            result.push(f);
+            // let f = command.execute(action, per_action_connection);
```
Timo614: Not a huge deal but may be worth just removing the line if we're unlikely to revert it back / can always use the commit history if we need it again.

PR #1269, opened 2019-05-13
Fix transfer issue with TicketInstance#find_for_display

Timo614:
### Description:
Similar to #1266 / #1267 this fixes the other for the logic to display an individual ticket for display.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
It may be worth DRYing up the logic in question if we can since it uses a lot of the same joining and select logic just with different filters as the other method. 

I tried unsuccessfully by using diesel's `BoxedSelectStatement` but the joining makes the type pretty crazy so not sure how feasible that is to use. Perhaps we could adjust it to use some direct SQL that we put some logic in place for checking if the filters are null and adjusting depending upon that (going around diesel). Will leave this PR as this for now so we can hotfix as needed similar to the other PR we pushed up earlier.

PR #1270, opened 2019-05-14
Tiered ticket type edit bug

### LINKED ISSUE #1268: Tiered ticket type goes off sale after editing event
Steps to reproduce:
1. Create Ticket type A with sales ending in the future
2. Create Ticket type B, set "Sales start after sales end for Ticket A"
3. Buy all the tickets in ticket type A
4. Sales should be enabled for ticket type B (can add to cart)
5. Edit event + save

Expected: 
Sales for ticket type B remain enabled

Actual:
Ticket type B says "On sale soon"

https://cl.ly/213d0f992c03

mikethetike:
### Closes Issues:
Closes #1268 

### Description:
This PR fixes the above bug, but also should make editing ticket types more resilient in that it checks whether sales should start immediately for tier ticket types when editing

## Release Details:
Fix bug where editing a tiered ticket type would take it off sale

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/cart.rs
```diff
@@ -104,14 +104,6 @@ pub fn update_cart(
     cart.set_user_agent(request_info.user_agent.clone(), false, connection)?;
     cart.set_tracking_data(json.tracking_data.clone(), Some(user.id()), connection)?;
 
-    // Beware there could be multiple carts that meet this condition
```
mikethetike: I moved this logic into the `orders` struct so that it can be tested in the `db` crate

db/src/models/orders.rs
```diff
@@ -1089,6 +1089,13 @@ impl Order {
         }
         self.update_fees(conn)?;
         self.validate_record(conn)?;
+        // Beware there could be multiple orders that meet this condition
+        for (ticket_type_id, remaining) in self.ticket_types(conn)? {
+            if remaining == 0 {
+                TicketType::find(ticket_type_id, conn)?
+                    .check_for_sold_out_triggers(Some(current_user_id), conn)?;
```
Timo614: What happens when these are no longer reserved because the tickets expire? We have both ticket types on in tandem? Or does the original ticket type no longer appear after?

Just wondering about the behavior we'd see with a highly active event where people are scooping up tickets as soon as reservations end. Just thinking we'd end up hitting the `check_for_sold_out_triggers` logic quite a bit in that scenario and probably end up logging a lot of sold out domain events if that's the case (if I'm following the logic correctly, I'm somewhat new to the tiered ticket type logic).

That may not be a bad thing though as we were sold out multiple times in that case. Since we run this for every cart update I'd think we also could run into spamming the domain events if someone were to add another ticket type that isn't sold out after they have already added the sold out one. So they could keep editing their cart and fill the logs with the sold out domain events (not that it would do them any good but just pointing it out). Wonder if it's worth only looking at ticket types that are new or changed in the cart. So perhaps doing a quantity check at the beginning and any that now have increased should run through this, etc. just to avoid that scenario (although it seems unlikely since we drop people off in the cart pay flow after adding tickets).

mikethetike: I share your concerns regarding the spamming of the domain events. Let me see if I can add a pre-check or something.

I don't think it's a massive problem to have both ticket types on sale if some tickets get released back into the first ticket type (e.g. refunds, expired orders). The important thing is that the second tier (usually more expensive) is released so that if there is a huge on sale event, the tickets are quickly available for the next batch of users.

mikethetike:
I've disabled the SoldOut domain event for now. Let's revisit it again later

Issue #1272, opened 2019-05-14
Ticket Transfers - A ticket transfer should not expire automatically


Timo614:
connected to big-neon/bigneon#154

Context: A ticket transfer link should not expire until the transferrer or an admin manually cancels the transaction. Many people wait until the last minute to accept there transfer, thus transfer links expire causing confusion. The expiry is unnecessary.

Anziskavds:
Reviewed and tickets no longer expires automatically. I waited 40 min after transferring a ticket and I still received the transferred ticket.

Anziskavds:
Tested on Prod. Transferred tickets and waited a while before logging in to see if the tickets was transferred successfully or if it automatically expired. Tickets transferred successfully and didn't expire.

PR #1263, opened 2019-05-10
Transfers cancel support

Timo614:
### Closes Issues:
Closes: #1260
Closes: #1252
connected to big-neon/bigneon#145 
Requires #1282

### Description:
This branch fixes the associated issues by adjusting the transfers data model to be better structured and then builds out support for the endpoints to list and cancel transfers.

Changes:
- Adds a new GET `/transfers` endpoint which can list transfers either with the user as the source or destination
- Adjustments to the data model
- Adds new GET `/orders/{id}/transfers` endpoint which lists the transfers for a given order
- Adds new DELETE `/transfers/{id}` which can be used to cancel a transfer (used delete as the verb as we use that across all other cancelling actions in the repo)

Note:  
- Incoming transfers will all be completed because they either were direct transfers to an existing user or accepted on user create via the link to send via email or phone.
- Wasn't able to test SMS functionality as I don't have access but confirmed the logic in the logs shows it sends similarly with block external comms set.

##### Fetching transfers for user 

```
curl http://localhost:9000/transfers --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJkMGUwMGFlOC0xOWFlLTRkNjQtOWMyNS01NzM3MGQwYzg0NGYiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTU3NzkwNzkzfQ.32ZACrw99rS03MvN5RsagOhreALiL-PEZsn0LminNiw"
[...]

PR #1273, opened 2019-05-14
Transfer: remove expiration date

Timo614:
### Closes Issues:
Closes: #1272 

### Description:
This branch removes the idea of a transfer expiration. Transfers have 3 states they can be in: Pending, Completed, and Cancelled. If a transfer has not been cancelled or completed it will no longer expire.

This branch is built off of #1263 given it significantly altered the transfers data model. It will need to be rebased once 1263 is merged into master but I'll keep it up to date by rebasing it over that branch (which will be rebased over master routinely as conflicts arise). 

Note: tests will fail here due to the same reason as #1263 -- we need to have the ENV vars added to the build server (with random data since block_external_comms = true) but don't have access. Will rerun the tests once we're set here in the associated PR.

## Release Details:

The migration here cancels any pending expired transfers. It does this because we allow tickets to be transferred multiple times so long as they were expired so it's possible removing the expiration date could activate already expired transfers otherwise.

### Migrations
 * 20190514145813_remove_transfer_expiration - removes the transfer expiration. Any pending transfers set to expire are updated here to be Cancelled so we don't violate the validation we have in place to confirm only one pending transfer is active for a ticket instance at a time.

### Environment Variables
 * No change


mikethetike:
Added DNM. See related #1263

Timo614:
I'll rebase this over master today and remove the commits from the other transfers branch which are now merged in.

PR #1275, opened 2019-05-15
Better DB locking on Carts

mikethetike:
## Release Details:
Fixed a bug where the entire assets table was locked instead of just the ticket_instance row when adding tickets to cart

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
It may be worth adjusting the bulk orders test to be partially done via external payments. I tried making a larger dataset with https://github.com/big-neon/bn-web/issues/1441 using the mocha test and it takes quite a bit of time with all the Stripe calls. May be worth limiting it in terms of external services. Doesn't represent the true makeup of the site users but could always run some SQL after the fact to emulate the credit card step (updating the payment types, etc and pretending they were done via Stripe if we needed for local testing).

Maybe some mix of the two? Like mostly external but some stripe orders in there?

integration-tests/mocha/helpers/user.js
```diff
@@ -1,26 +1,28 @@
...
-let requestBody = `{
+const registerAndLogin = async function (saveVarName) {
+    let email  ="user" + (new Date()).getTime() + Math.floor((Math.random() * 1000000)) + "@tari.com";
+    pm.environment.set("last_email", email);
+    console.log(email);
+    const response = await post('/users', `{
 	"first_name":"User",
 	"last_name":"Surname",
-	"email":"{{last_email}}",
+	"email":"` + email + `",
```
Krakaw:
```suggestion
	"email":"${email}",
```

integration-tests/mocha/test/21 Bulk orders/Create bulk orders.js
```diff
@@ -0,0 +1,118 @@
...
+                console.log(user_token);
+                user_tokens[i] = user_token;
+                return cart.replace(ticket_type.id, null, 1, user_token);
+            }).then(()=>  {return stripe.getToken();}).then((stripe_token)=>{
+
+                let requestBody = `{
+	"method": {
+		"type" : "Card",
+		"provider": "Stripe",
+		"token" : "` + stripe_token + `",
```
Krakaw:
```suggestion
		"token" : "${stripe_token}",
```

integration-tests/mocha/helpers/events.js
```diff
@@ -27,10 +27,10 @@ const create = async function (saveVarName) {
 }
 
 
-const createTickets = async function (event, saveVarName) {
+const createTickets = async function (event, saveVarName, quantity) {
     let requestBody = `{
 	"name":"Default_Pricing_{{$timestamp}}",
-	"capacity": 100,
+	"capacity": ` + (quantity || 100) + `,
```
Krakaw:
```suggestion
	"capacity": ${quantity || 100},
```

integration-tests/mocha/helpers/cart.js
```diff
@@ -3,15 +3,17 @@ const pm = require('../test/pm');
 const expect = require('chai').expect;
 
 
-const replace = async function (ticket_type_id, saveVarName) {
+const replace = async function (ticket_type_id, saveVarName, quantity, token) {
     let requestBody = `{
 "items": [{
 "ticket_type_id": "` + (ticket_type_id || "{{last_ticket_type_id}}") + `",
-"quantity":20
+"quantity":` + (quantity || 20) + `
```
Krakaw:
```suggestion
"quantity":${quantity || 20}
```

db/src/queries/add_tickets_to_hold.sql
```diff
@@ -6,7 +6,7 @@ WITH r AS (SELECT t.id
              AND coalesce($4, 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11') =
```
Krakaw: Do we need these dummy guid's?

db/src/queries/reserve_tickets.sql
```diff
@@ -5,7 +5,7 @@ WITH r AS (SELECT t.id
              AND a.ticket_type_id = $3
              AND coalesce($4, 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11') =
```
Krakaw: Do we need these dummy guid's?

integration-tests/mocha/test/pm.js
```diff
@@ -7,9 +7,9 @@ const DEFAULTS = {
 const pm = {};
```
Krakaw:
```suggestion
const pm = {};
const debug=require('debug');var log = debug('bn-api');
```

PR #1293, opened 2019-05-28
Add slug to event

mikethetike:
Related to big-neon/bn-web#1452

### Description:
Adds the ability to find by a friendly URL, e.g. `/events/big-event`

### Migrations
 * 20190527150600_create_slug_events

### Environment Variables
 * No change


db/src/models/events.rs
```diff
@@ -192,6 +210,21 @@ impl NewEvent {
     }
 }
 
+fn create_slug(name: &str) -> String {
+    // Unwrap should be treated as a compile time error
+    let whitespace = Regex::new(r#"\s+"#).unwrap();
```
Krakaw: I think we should also add something like this crate https://crates.io/crates/unidecode so that we use plain ascii for our slugs.
The other option is a trigger in Postgres and we use the `unaccent` extension
```sql
CREATE EXTENSION IF NOT EXISTS "unaccent";
SELECT unaccent(slug) AS slug
```

db/migrations/20190527150600_create_slug_events/up.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE events
+    ADD slug VARCHAR(255);
+
+CREATE UNIQUE INDEX index_events_slug ON events (slug);
+
+UPDATE events
+SET slug= substr(lower(replace(name, ' ', '-')), 255);
```
Krakaw: This is going to fail if there are any duplicate event names (which we'll have for sure in dev and staging) it is also taking the substr from 249 to end of string
What about something like:
```suggestion
SET slug = CONCAT(SUBSTR(LOWER(REPLACE(name, ' ', '-')), 0, 249), '-', SUBSTR(MD5(RANDOM()::TEXT), 0, 6));
```

db/src/models/events.rs
mikethetike: Added unidecode. I don't think it's necessary to add the postgres trigger at this time

Issue #1310, opened 2019-06-04
Event Audit Report: Face value calculation incorrect


Timo614:
See: https://github.com/big-neon/bn-web/issues/1417

It should be ticket fee inclusive it also currently is showing a price for comps etc which it shouldn't as it's the price the user paid for the tickets including their fees.

nine-tari:
Was merged a bit back fixes the issue which was causing bn-web to show an incorrect face and totals columns.

Anziskavds:
Tested on Beta and the calculations is correct

Anziskavds:
Tested on Prod and looks like the calculations are correct

PR #1311, opened 2019-06-04
Event audit report: Fix face calculation

Timo614:
### References Issues:
References: #1310

### Description:
Face should show the discount inclusive pricing shown to the user.

 ### Examples
Before:
Screen Shot 2019-06-10 at 1 21 16 PM
After:
Screen Shot 2019-06-10 at 1 30 01 PM

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Confirming why the promo discount codes is listed with 0 items in the result set so one moment prior to merging / testing this.

Issue #1313, opened 2019-06-05
Add pagination to /guests endpoint


Krakaw:

Tas4tari:
Has been released

PR #1314, opened 2019-06-05
Add pagination to guest list

Krakaw:
### References Issues:
Connected to #1313 
### Description:
This PR adds pagination to the guest list. I removed the retrieve_guest_list.sql file and replaced it with a native diesel query, which allowed us to use the pagination.
I have also added search on order_id as well as ticket_instance_id.
I've also pulled the guest list functionality into its own function so we could re-use it for ticket_instances.
My only real cause for concern is [here](https://github.com/big-neon/bn-api/pull/1314/files#diff-03229ac4991883008bd5f4d220342ffaR1172)
```sql
CASE WHEN ticket_instances.redeemed_by_user_id IS NOT NULL THEN (SELECT CONCAT(u2.first_name, ' ', u2.last_name) FROM users u2 WHERE u2.id = ticket_instances.redeemed_by_user_id) ELSE NULL END  AS redeemed_by
```
Doing an inline select because diesel doesn't let you join the same table twice (unless you jump through hoops I didn't have the time to jump through)

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/events.rs
```diff
@@ -853,16 +868,19 @@ pub fn guest_list(
 
     let query_string = query.clone().query;
     let changes_since = query.clone().changes_since;
-    let tickets = event.guest_list(query_string, &changes_since, conn)?;
+    let paging = query.clone().into();
+    let tickets_and_total = event.guest_list(query_string, &changes_since, Some(&paging), conn)?;
+    let tickets = tickets_and_total.0;
```
Krakaw: Unpack correctly

Issue #1330, opened 2019-06-12
Allow transfers to be queries without authentication


Jasonvdb:
Not 401s for this guy please:
`/transfers/transfer_key/{transfer_key}`

PR #1331, opened 2019-06-12
Allow anonymous access to view a transfer by key

mikethetike:
### References Issues:
References #1330 

### Description:
Allow anonymous access to the details of a transfer. The main problem is that we cannot show a user the details of the transfer before they sign up. The reasoning behind allowing this is that if a user has the transfer key, then they still need the signature to claim the transfer, so there is no data leaked

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Makes sense why we'd want this to be open. Pushed up a commit to adjust the test. Sorry about that, I should have realized the callers of this would be logged out since direct transfers are immediately processed and anyone else won't have an account.

Issue #1312, opened 2019-06-05
`/admin/orders` endpoint for order list for an event.


Jasonvdb:
Can pretty much work the same as current `/orders` endpoint but needs to accept `event_id` instead of just returning the current user's orders.

Required for big-neon/bn-web#1523
Design: https://share.goabstract.com/eb756c85-a087-4208-b3d3-ac4d5ca002a9?collectionLayerId=e6f5fc83-c651-48b7-be93-3db8123c34b3&mode=design

Jasonvdb:
Sorry some more requests:
- [ ] Needs to also support a search query (Order #, Name,Email, Ticket #)
- [ ] Filter by ticket type
- [ ] Filter by promo code

Jasonvdb:
Also need the user names (first+last) and the user email address.

Jasonvdb:
And the point of sale. I think that's just `Online | App | BoxOffice`.

Jasonvdb:
@mikethetike need the first name, last name, email, POS to each order

Anziskavds:
To be tested with order management

PR #1339, opened 2019-06-14
Admin orders endpoint

mikethetike:
### References Issues:
References  #1312 

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/orders.rs
```diff
@@ -683,6 +695,207 @@ impl Order {
...
+        paging: &PagingParameters,
+        conn: &PgConnection,
+    ) -> Result<(Vec<DisplayOrder>, i64), DatabaseError> {
+        let mut query = sql_query(
+            r#"
+        SELECT *, COUNT(*) OVER () as total FROM (
+        select distinct o.*
+        from orders o
+            inner join order_items oi on oi.order_id = o.id
+            inner join ticket_types tt on oi.ticket_type_id= tt.id
```
Krakaw: Will this pick up the fees rows because they won't have a `ticket_type_id`?

mikethetike: Probably will, but will be removed by the `distinct` on orders

Issue #1337, opened 2019-06-13
Engagement - Ticket Purchase Automated Drip (Email/Push)


Tas4tari:
Context: After purchasing tickets to an event a drip campaign should be initiated which includes content that gets them excited about the event, attempts to get them to get their friends to come, attempts to drive them to purchase tickets for their friends, informs them about the need to download the app, provides information about getting to the venue, thanks them and provides artist/venue content (upcoming shows) after the event

Day of Purchase: Order confirmation email with instructions to download the app, include buy your friend a ticket
T - 5d: Get excited and get your friends involved
T - 3d: Events this weekend, got your tickets?
T - 1d: Are you ready? Get pumped, get the app (possible upsells)
0d: Your tickets are ready for you
+1d: Good times? How about some more? Upcoming events at that venue
+5d: [Promo]
+14d: [Promo]

Timo614:
Should no promos be delivered if there are no upcoming events for the venue? What sort of copy were we thinking for the upcoming events? Guessing since it's sendgrid we'll need to provide that copy all together as there won't be a way to dynamically include multiple events as variables and it would be cumbersome to try to include whatever event details we needed for each as their own variables.

Do we need the ability to have multiple sendgrid accounts as part of this so that we can keep the marketing emails outside of the transactional account? Guessing we'll need to allow people to opt out of these emails as well.

Timo614:
Should we change the text for the push notifications? Doesn't give much info about what event is about to happen (used the text from the copy above that we had available).

I'm updating the PR now but will put this in blocked and set it to DNM pending the answers here.

zmelamed:
@Timo614 I'll get to this today, need to think through the different conditions and the appropriate copy. We'll definitely hold on the post event notifications for the time being until we have some strong calls to actions.

Timo614:
Sounds good :+1:

Tas4tari:
@Timo614 did Zach confirm the push notifications with you?

Timo614:
No, I don't have the copy to use here.

Tas4tari:
@ridley-tari @harrisonlemon would you be able to assist us with sourcing the text for the push notifications

harrisonlemon:
@Tas4tari Copy for drip campaigns (mobile & email) [here](https://docs.google.com/document/d/1QPC4GOE7xn70CXiMksUffq41LDjFAtW_jLLWcGhtFFs/edit)

[...]

PR #1343, opened 2019-06-18
WIP: Engagement drips

Timo614:
### References Issues:
References: #1337

### Description:

WIP at the moment as I still need to spend a little time testing it and the associated issue was missing some copy / has open questions related to sendgrid etc. we may need to address. I'll add some examples and adjust as soon as I hear back on the associated issue.

Note:
- Stored the generated template data for an event in the payload to reduce the number of queries needed per user since it would be the same across a given event
- If a pending transfer exists for a ticket the owner is not taken into account when compiling the list of user's to email (they may still get an email if they have some other ticket not pending a transfer).
- Not sure how to test if the pusher is working locally but the logic there emulates other uses we've had so should work.

## Release Details:

The script to regenerate drip events will need to be run `cargo run --bin api-cli regenerate-drip-events` normally the drip events will be generated at the time of publishing after the initial setup for existing events.

### Migrations
 * No Migrations

### Environment Variables
 * SENDGRID_TEMPLATE_BN_EVENT_UPCOMING_DRIP
 * SENDGRID_TEMPLATE_BN_EVENT_POST_PROMO_DRIP

I've broken apart the drips into two env variables / templates. It seems like the content of the emails changes dramatically with whether or not the event has passed. All of the normal event based variables are included in the template data.


Timo614:
Will rebase this quickly to fix the conflicts.

PR #1344, opened 2019-06-19
Hide event tickets when not published

mikethetike:
### References Issues:
References big-neon/bigneon#196

### Description:
Hides the ticket types from an event if it has not yet been published, or if the user has access to edit the event so that they can preview it

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Although wouldn't it make more sense to just lockdown the show action completely versus just limiting the ticket types? Is there ever a reason why someone should be able to access the show action if they don't have preview access and it's unpublished?

Krakaw:
See the comment by @harrisonlemon [here](https://github.com/big-neon/bigneon/issues/196#issuecomment-503581088) let's go with a full 404 on the event page.

Krakaw:
@mikethetike I pushed https://github.com/big-neon/bn-api/pull/1344/commits/cfcc75c645519fe281fdf89c882c2289f456e88a that will 404 the event entirely if it is unpublished and the user does not have permission

Issue #1338, opened 2019-06-13
Fees - Event Level Fees


Tas4tari:
Context: More robust fee capabilities are now a need with All Aboard / Espee and Midway. Minimum org & event-level fee requirements:

1. We require the ability to add an additional fee per ticket type.
2. The amount allowed within the additional fee should be capped at $2.

Anziskavds:
Tested Develop and Beta on both Web and Mobile web and is working as expected.

Anziskavds:
Tested on Prod and is working as expected

PR #1348, opened 2019-06-20
Event level fees

mikethetike:
### References Issues:
Related to big-neon/bigneon#187
Related to #1338

## Release Details:
Add the ability to set an additional fee per ticket types for an event. The maximum amount allowed is set on an organization level. The default is currently set at 0 and will need to be set manually to $2 for a specific organization as indicated in the issue.

### Migrations
 * 20190619122323_additional_fees_on_ticket_types

### Environment Variables
 * No change


mikethetike:
Yeah it might be worth having an extra field on the order_items for additional_fee to track it as that point in time, but my main concern was getting it out with keeping the reports in tact

mikethetike:
I've created #1352 to separate them in future

PR #1349, opened 2019-06-20
Add user data to admin/orders

mikethetike:
### References Issues:
Related to #1312

### Description:
Add user data to order returned

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
My thinking was to change it to use `user.id` but thinking about it there may be different versions of clients in the wild, so user_id should stick around

PR #1356, opened 2019-06-21
Adjust fan search

Timo614:
### References Issues:
https://github.com/big-neon/bigneon/issues/178

### Description:
It was noticed that the organization fans endpoint was not returning transferred users in the list of fans. On further inspection it looks like it doesn't show box office purchased for or event interested users as well.

In addition:
- events#search_fans was missing transferred ticket users as well
- organizations#has_fan only checked if the user had a purchase so any box office purchasers, event interested users, or transferred users failed the check there that's done for the profile showing
- organizations#get_profile_for_organization  only checked if the user had a purchase so any box office purchasers, event interested users, or transferred users failed the check there that's done for the profile showing
- Opted to use load_and_count_pages over the existing way we had it -- switched the logic to just return the DisplayFan and cut out the need for the intermediate struct

Note:
- I had the events#search_fans logic use the organizations#search_fans with a Option for the event_id to DRY things up
- The organizations#has_fan logic also uses the organizations#search_fans to keep them in line just passes in the user_id for the query

Future issues / out of scope:
- Added a new tickets_owned field to the profile response so we can list that on the profile if desired (could have altered ticket_sales but it would be misleading) / may be worth a followup bn-web issue once this merges to switch to that (see examples for why)
- The fans list currently orders by the last order but event interests, transfers, may cause a person to be added without a last order so they appear near the top. We could add another field that has the earliest date from them using their created at dates though. May be worth doing as a separate issue as I may be less available for the next week at least.
- The history page currently lists 0 tickets and 0 revenue when a refund occurs (https://user-images.githubusercontent.com/1319304/59950122-09d22f00-9443-11e9-8e65-01d1379dd859.png). We may want to pass back how many tickets were refunded and the refunded amount from the order. Also the order show page shows the line item for a purchase but with a refund it just shows $0 for the amount so we probably need a bn-web issue to include a line showing the refund negated the price of the fee and the items (https://user-images.githubusercontent.com/1319304/59950144-0b9bf280-9443-11e9-9c96-1de067a267ad.png).
- Ended up using some subqueries to accomplish the totaling of revenue as there are multiple ticket_instances records so it otherwise lead to some odd counting / someone with more advanced SQL knowledge may be able to optimize the query to remove that but at least with the tests in place now we have the ability to confirm no regressions from it if we do

Question:
- In the profile page the event count now includes all of the event interested, purchased, or transferred in the count -- should it?
- Are we planning on adjusting the profile page soon? May be worth listing the event interests there for an organization so it's not blank for those users. Or somehow signifying that's why they are a fan in some way with a label.

### Examples

With two users who just had transfers:
```
bigneon=# select u.email from transfers t join users u on t.destination_user_id = u.id where t.status = 'Completed';
              email               
----------------------------------
 user1561128360091319107@tari.com
 nine+testing@tari.com
(2 rows)
```

With one user who expressed interest but nothing else:
```
bigneon=# select u.email from event_interest ei join users u on ei.user_id = u.id;
         email          
------------------------
 nine+testing2@tari.com
(1 row)
```

Example mentioned in future issues section where history shows 0 because of refund (this user's $80 purchase was via the box office as well):
Screen Shot 2019-06-21 at 1 11 14 PM
Screen Shot 2019-06-21 at 12 31 34 PM

Example of tickets_owned vs ticket_sales (mentioned in questions above, we should probably make a bn-web issue to modify that to use the new field) -- this user transferred one of their tickets away so I'd imagine we want 9 there:
Screen Shot 2019-06-21 at 3 51 23 PM
Screen Shot 2019-06-21 at 3 51 48 PM

Same idea as last just in reverse, user who got the ticket shows 0 there since 0 sales but does own 1:
Screen Shot 2019-06-21 at 3 52 07 PM
Screen Shot 2019-06-21 at 3 52 02 PM

User who just expresses interest (we may want to surface the interest here on bn-web):
Screen Shot 2019-06-21 at 3 52 16 PM
Screen Shot 2019-06-21 at 3 52 34 PM

Example of user with multiple purchases:
Screen Shot 2019-06-21 at 3 53 22 PM
Screen Shot 2019-06-21 at 3 53 32 PM

In the branch 22 users were found:
Screen Shot 2019-06-21 at 11 06 19 AM

Outside of it on master it includes box office user and only showed 19:
Screen Shot 2019-06-21 at 11 55 43 AM
Screen Shot 2019-06-21 at 11 56 23 AM

On master box office user shows purchases on behalf of others in this branch it errors as it's not an organization fan (no self orders):
Screen Shot 2019-06-21 at 11 57 05 AM

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
If I have time next week I'll try to handle the API side to support the following things I mentioned in the description:

> - The fans list currently orders by the last order but event interests, transfers, may cause a person to be added without a last order so they appear near the top. We could add another field that has the earliest date from them using their created at dates though. May be worth doing as a separate issue as I may be less available for the next week at least.
> - The history page currently lists 0 tickets and 0 revenue when a refund occurs (https://user-images.githubusercontent.com/1319304/59950122-09d22f00-9443-11e9-8e65-01d1379dd859.png). We may want to pass back how many tickets were refunded and the refunded amount from the order. Also the order show page shows the line item for a purchase but with a refund it just shows $0 for the amount so we probably need a bn-web issue to include a line showing the refund negated the price of the fee and the items (https://user-images.githubusercontent.com/1319304/59950144-0b9bf280-9443-11e9-9c96-1de067a267ad.png).

Specifically adding a new field and order type to order by something like `organization_fan_associated_at` and to break apart the history purchase item so it shows the number of tickets refunded and the amount refunded separately so that can be shown. Will confirm that the order page uses an endpoint that provides enough info for the line items we need there (guessing it does already just needs some bn-web adjustments but will confirm then).

mikethetike:
Agree with the new issues. Especially the refund tickets one. That will be confusing until it's fixed.

mikethetike:
Let's keep an eye on the running time of this query. It shouldn't be too bad. If it starts showing up in the logs as taking more than 500ms, we can optimize

PR #1382, opened 2019-07-05
DNM Reduce DB calls on event::show

mikethetike:
### Description:
Reduces the 4 calls to Event::find, Org::find, Venue::find and FeeSchedule::find into a single call to the DB. 

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
I'm busy trying to reduce the number of calls in DisplayTicketType, but I see to be creating some regression issues

Issue #1388, opened 2019-07-09
Remove automatic refund of order level fees


mikethetike:
Currently if the last ticket order item in an order is refunded, any remaining fee items are refunded. The new interface will allow the user to select these items individual for refund

mikethetike:
Waiting for new refund screens to be merged in

PR #1389, opened 2019-07-09
Remove auto-refunding of order level fees

mikethetike:
### References Issues:
Related to #1388 

### Description:
Currently if the last ticket order item in an order is refunded, any remaining fee items are refunded. The new interface will allow the user to select these items individual for refund

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/orders.rs
```diff
@@ -377,34 +377,34 @@ impl Order {
...
-                and oi2.refunded_quantity <> oi2.quantity
-            )"))
-            .select(order_items::all_columns)
-            .load(conn)
-            .to_db_error(
-                ErrorCode::QueryError,
-                "Could not check if order only contains event fees",
-            )
-    }
+    //    fn find_orphaned_per_event_fees(
```
Timo614: May be worth just removing the commented out logic completely. Can always check the commit history if we need to resurrect it.

PR #1401, opened 2019-07-17
Changing notes path to work better for bn-api-node

Jasonvdb:
### Description:
Changed `/{main_table}/{id}/notes` to `/notes/{main_table}/{id}` so it makes more sense in the middleware.

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
Good idea

Issue #1405, opened 2019-07-17
Add free-form text for push notifications


Krakaw:
Only allow super users to send push notifications with custom text

Anziskavds:
As discussed with Keith we need to test this on Prod for the most accurate testing results. @Tas4tari

Anziskavds:
Tested successful on Prod.

PR #1416, opened 2019-07-22
Add custom broadcast type

Krakaw:
### References Issues:
Connected to #1405 

### Description:
Added a new `BroadcastType::Custom` that allows free-form text to be sent via push notifications.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/broadcasts.rs
```diff
@@ -161,6 +161,18 @@ impl Broadcast {
 
 impl NewBroadcast {
     pub fn commit(&self, connection: &PgConnection) -> Result<Broadcast, DatabaseError> {
+        if self.notification_type == BroadcastType::Custom
```
Timo614: It may be worth switching this to use a custom validator.

Something like:
`#[validate(custom = "broadcasts::message_present_when_custom_type")]`

And just running `validate()` on the self here.

Timo614: Actually scratch that, the validators don't have model context.

We've used a `validate_record` method on other models to get around that (passing in a connection but at the very least having model context). Not the end of the world either way though and it's only one field so I'm indifferent personally if we want to use business process here since it accomplishes the same end result.

Krakaw: Hehe yes, I was toying with that, and I landed up going the quick way. You are 100% correct in that it'll clean this up. I'll refactor and push up the change.

Anziskavds:
As discussed with Keith we need to test this on Prod for the most accurate testing results.
@Tas4tari

PR #1425, opened 2019-07-26
[High Risk] Fix Order details amount when refunded

mikethetike:
### References Issues:
Related to #1426

### Description:
Previously when a ticket was refunded, the line item would have a ticket_price_in_cents of 0. Now even if refunded, the original ticket price is shown

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/queries/order_details.sql
```diff
@@ -45,9 +44,10 @@ FROM (
                 wallet_owner.last_name             AS attendee_last_name,
                 tt.id                              AS ticket_type_id,
                 tt.name                            AS ticket_type_name,
-                coalesce(h.name, c.name)           AS code,
+                coalesce(h.redemption_code, c.redemption_code)           AS code,
                 coalesce(h.hold_type, c.code_type) AS code_type,
-                tfs.id                             AS pending_transfer_id
+                tfs.id                             AS pending_transfer_id,
+                dis.unit_price_in_cents            AS discount_price_in_cents
```
Timo614: May need something like:

```
                CASE
                    WHEN h.redemption_code IS NOT NULL AND h.hold_type = 'Comp' THEN
                      tp.price_in_cents
                    WHEN h.redemption_code IS NOT NULL THEN
                      LEAST(h.discount_in_cents, tp.price_in_cents)
                    WHEN doi.id IS NOT NULL THEN
                      0 - doi.unit_price_in_cents
                    ELSE NULL
                    END      
```

To account for the different discount types. Only the codes have discount line items so otherwise would return null for the holds related discounts.

mikethetike: Interesting. I think what we should do is change all hold and code discounts to create a new discount row. That will make reporting a lot more consistent

mikethetike: Pushed a commit that consolidates discount logic to always create a discount row

Timo614: Sounds good. There's no business reason to keep it right? Wasn't sure if it was discount inclusive for something like if they want face value to reflect $0 for comps etc?

Timo614: Do we need to do anything about the existing orders? It'll return `0` for those orders when this does go live since we're relying on there being a discount row now but this stuff predated it. Seems a bit risky to mess with existing order data to break it out for past uses.

mikethetike: Yeah for older orders it's going to show a reduced ticket price and a 0 discount. I think it's better to leave the old data as it is, and new orders from release date will have this structure and look better.

mikethetike:
Also related to #1428

PR #1445, opened 2019-08-01
Customer.io integration

Timo614:
## References Issues:
References: https://github.com/big-neon/bigneon/issues/220

### Description:
This branch attempts to take care of the drip related activities by linking with customer.io. I've created some zapier zaps which translate the payload format into the format accepted by customer.io which can be looked at for an example of how it works.

If you add a `domain_event_publisher`:
```
insert into domain_event_publishers (event_types, webhook_url, import_historic_events) values ('{UserCreated,TemporaryUserCreated}','https://...', true);

insert into domain_event_publishers (event_types, webhook_url, import_historic_events) values ('{OrderCompleted}','https://...', true);

insert into domain_event_publishers (event_types, webhook_url, import_historic_events) values ('{TransferTicketStarted,TransferTicketCancelled,TransferTicketCompleted}','https://...', true);
```

It'll have those event types (array) published to that webhook url. If import_historic_events is enabled all domain events are published otherwise it will only publish ones created after the publisher's created_at. It can also be restricted to organization_id (we'll want to test it a bit more though as it was out of scope) so only events of a given organization have their domain events published.

It is advisable to setup 3 webhooks like I did in zapier for the example as the formats are so different and you have to specify the fields the webhook will hit with zapier's functionality. 

Note:
- It's probably worth taking a closer review look at the domain event publishing logic -- I kept some of it close to the original frozen implementation so not sure if we've run into issues with the domain event processing that may need to be modified here as well / wasn't following the updates / issues we've had there in the past.
- I have the migration remove all of the domain actions associated with the types I've removed. I did this out of precaution in case we ever add a frontend to display domain actions in the future as they will be invalid no longer being part of the enum set. I didn't want to keep them in the set as we'd need some code then for the logic that matches on the type. I did keep the previous domain events. Let me know if we want to nix them as well though and I can adjust the migration / enum to remove them.
- I also had it remove some duplicate domain events we had for transfers (previously we fired the events for individual tickets plus the transfers so now it's just the transfers).
- A transfer added prior to our logic to add the domain events for transfers will not be included in the export. We've had those for some time now though so it's really a small subset.
- I used the CommunicationChannelType enum to handle sending out the webhooks. I could do something more specific but it felt like the cleanest approach and webhooks are a form of server -> server communication although it does feel like the original intent was to communicate with users. Let me know if you want me to change that. Does have the added bonus of blocking them from going out in environments where BLOCK_EXTERNAL_COMMS is enabled.
- I added the logging of whether a transfer is a direct transfer or not and added it to the transfers data so we can customize messaging for those instances and filter out of the initial messaging, acceptance messaging, etc.
- I've added a script to backpopulate the temporary user creations. It makes domain events from when it would have been initially added allowing those to be populated in customer.io correctly with a timestamp.

Future:
- I didn't add validations for the domain event types that are supported to prevent domain publishers from using ones that will error out. I figured we could iterate on it and this was already getting a bit large.

Questions:
- The word doc with example text shows a set of parameters not referenced in the customer.io integration. Do we want the complete set?

### Migrations
 * `migrations/20190724195057_domain_event_publishing_modifications`

I've added a script in the cli to backpopulate the temporary user records. We're importing all of the old records so this is needed so they appear correctly in the timeline. 

```
cargo run --bin api-cli backpopulate-temporary-user-data
```

Once that has been run and it updates the existing records we can retire the code. It's a one-time task.

### Environment Variables
 * No change


mikethetike:
Haven't gone through it all, but this looks fantastic

Issue #1442, opened 2019-07-31
Order Management: Search: All issues requiring API work related to Search


Tas4tari:
- Search Filter should be cleaner, more precise. User will rarely search by promo type specifically (see parameters for search criteria).
- Search Filter: Ability to filter by broader parameters (date, point of sale, method of payment, status, ticket type, etc).
- Refunded ticket id’s do not render in search results but are still present in order https://cl.ly/256666891cd5
- Email: Ticket Transfer: When searching for ticket transfer on pending or cancelled transfer does not render results https://cl.ly/ba8d54666f4a
- Phone Number: Ticket Transfer: When searching by phone number for ticket transfer recipient on accepted transfer, phone number does not render results https://cl.ly/80a1ceefb706
- Order Number: When searching via Order Number, no results are displayed. https://cl.ly/ba8d54666f4a
- Ticket Transfer: Pending: When searching by phone number for ticket transfer recipient on pending transfer, phone number does not render results https://cl.ly/af0a2af1a58c
- Ticket ID: When searching via ticket ID, no results are displayed. https://cl.ly/ba8d54666f4a
- Phone number - event sales - searching for event purchaser by phone number does not render results https://cl.ly/e9ae0d5689e0
-

mikethetike:
Can we clarify "Point of sale"

mikethetike:
@Anziskavds please add each of these cases in the videos to our test cases

PR #1446, opened 2019-08-02
#1442 Enhanced order search

mikethetike:
### References Issues:
Connected to #1442 

### Description:
Add extra search fields for admin/orders

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/orders.rs
```diff
@@ -721,13 +721,17 @@ impl Order {
...
         email: Option<&str>,
+        phone: Option<&str>,
         name: Option<&str>,
         ticket_type_id: Option<Uuid>,
         partial_promo_code: Option<&str>,
         box_office_sales: bool,
         online_sales: bool,
         web_sales: bool,
         app_sales: bool,
+        _payment_method: Option<PaymentMethods>,
```
Timo614: Is this leftover from an earlier iteration? Or out of scope and handling in a future branch?

db/src/models/orders.rs
```diff
@@ -813,6 +829,20 @@ impl Order {
...
+            bind_no = bind_no + 1;
+            query = query
+                .sql(format!(" and o.order_date >= ${} ", bind_no))
+                .bind::<sql_types::Timestamp, _>(date);
+        }
+
+        if let Some(date) = end_date {
+            bind_no = bind_no + 1;
+            query = query
+                .sql(format!(" and o.order_date >= ${} ", bind_no))
```
Timo614: I think this filter has a typo and order_date should be `<=`

mikethetike: Good catch

db/src/models/orders.rs
mikethetike: I was waiting on a response on the issue. I think we already have the fields we require

mikethetike:
I also reduced the 401's to log as Info instead of Warn. That could have been in a seperate PR, but this one will probably go in soon

PR #1447, opened 2019-08-06
Delete events

Timo614:
### Description:
I have an asana issue which describes a desire to delete events. These events can only be deleted if they have no sales and are not currently in any carts. This just ended up being a check to see if any order items are associated with the event in question. 

Opted to go the route of marking the event as 'deleted' via a `deleted_at` field. Avoids the need to remove a lot of linger associations and gives us the ability to undo a mistaken delete. Let me know if we want a hard delete here though and can adjust. I have the logic filter out these events in most instances (the Event#find method doesn't as some tests use it to check the state but we could always filter it there as well). In addition since we only allow deletion if there have been no sales we don't need to filter the order sales queries in the repo.

Note: I opted to make a new route `DELETE "/events/{id}/delete"` as we already had `DELETE "/events/{id}"` setup for cancelling and I don't want to be disruptive with the consumers of the API. Not sure how we want to do those in the future / if we want a breaking release to move cancel over to a PUT `/events/{id}/cancel` route and reserve the DELETE `/events/{id}` for delete uses.

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
I agree that we should change the current DELETE route to `/events/{}/cancel` and use `DELETE` as the actual delete

api/src/controllers/events.rs
```diff
@@ -264,6 +264,24 @@ pub fn show(
...
+        Some(ref user) => user.has_scope_for_organization_event(
+            Scopes::EventWrite,
+            &organization,
+            event.id,
+            connection,
+        )?,
+        None => false,
+    };
+
+    if (!is_preview_allowed
```
mikethetike: 👍

mikethetike:
I think let's add the above to a new PR

PR #1450, opened 2019-08-12
Add back drip functionality ahead of customer.io

Timo614:
### References Issues:
Related to https://github.com/big-neon/bigneon/issues/220

### Description:
Issue 220 had us remove the drip functionality in favor of customer.io. Customer.io allows control of flows such that we can handle a lot of what we handle now in their tooling and remove the need for developer resources for modifications, improvements, testing, etc.

Thing is customer.io can't even have the production information imported until the release goes out and the tooling will be new to those implementing the flows / they may be missing components (I noticed in 220 they didn't request accept link access for invites which I added but there could be others). 

To avoid some period where we lack drips I'm reverting the changes to remove this drip for now. I tested the normal drip functionality and confirmed I received one for 7 days, 1 day, a few hours before as it used to function.

Note: 
- I didn't resurrect the regenerate bn-api cli script as it won't be needed and this is just a temporary change.
- `delete from domain_actions where domain_action_type in ('ProcessTransferDrip', 'RegenerateDripActions');` we'll need to add this to a new migration once we revert this PR's commits to remove the automatic drip functionality once we're set on customer.io

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Not sure why the tests are failing here:

```
1) Bulk create orders
--
1561 | "before all" hook:
1562 |  
1563 | AssertionError: expected 429 to equal 200
1564 | + expected - actual
1565 |  
1566 | -429
1567 | +200
1568 |  
1569 | at Object.getToken (helpers/stripe.js:28:32)
```

Looks like it's dying due to some Stripe js call to get a token. None of that would have changed in this branch as it has nothing to do with the order flow (it's just some domain actions and associated methods).

Timo614:
Oh it looks like that's:

```
429 - Too Many Requests | Too many requests hit the API too quickly. We recommend an exponential backoff of your requests.
```

We may have hit some rate limit with that bulk order test.

PR #1451, opened 2019-08-12
Promoter multi event access

Timo614:
### Description:
Adds ability to have multiple events associated with a promoter. In the past we completely replaced the user's organization user entry when we added a new event for the user. This caused invites to completely remove access to old events when users in the tooling added a new event invite.

The logic in this branch changes this to add a new record whenever an event promoter is added. Instead of logging an array with the organization user there's a new table for `event_users` which includes details such as the role. At the time of authorization the role is used for these users even if they have an event of each.

To allow for the frontend to show the right permissions I've added a new field to the response of the user for the readonly events they have access to. This is used as the general organization scopes list may not be accurate for these event specific functions. We should have a follow up web issue restrict based on that (the normal permission system we have in place breaks down a bit on the per event level as it's all organization based).

Screen Shot 2019-08-13 at 4 10 56 PM

### Examples
User added as readonly promoter:
Screen Shot 2019-08-14 at 10 40 05 AM

Readonly promoter has readonly access to event:
Screen Shot 2019-08-14 at 10 41 03 AM

Readonly promoter event added to event ids:
Screen Shot 2019-08-14 at 10 41 23 AM

Second regular promoter access added and accepted:
Screen Shot 2019-08-14 at 10 42 18 AM

Each shown in response:
Screen Shot 2019-08-14 at 10 42 50 AM

Can see roles in the db:
Screen Shot 2019-08-14 at 10 44 37 AM

I confirmed if I try to edit the promoter read only event (even though it's allowed due to the frontend needing an update) it fails but the promoter one succeeds.

### Migrations
 * `20190808203414_promoter_multi_event_access`

### Environment Variables
 * No change


db/migrations/20190808203414_promoter_multi_event_access/up.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE event_promoters (
```
mikethetike: It's probably better to keep this generic as `event_users`

Timo614: Sounds good will make these modifications now.

Timo614:
I'm going to make one more addition to help with the frontend a bit -- will pass back the scopes per event as well as that fits more in line with how they determine authorization.

Timo614:
I made the suggested change to make it more generic and adjusted the readonly boolean to be a role field instead which is used during the event permission checking.

Also as a final adjustment I just added the scopes per event to the response to make it easier for the frontend to determine if they will be able to authenticate a call here:

Screen Shot 2019-08-14 at 11 24 33 AM

PR #1458, opened 2019-08-14
Add device tracking for customer.io

Timo614:
### References Issues:
Relates to https://github.com/big-neon/bigneon/issues/220

### Description:
In the original customer.io issue I had thought that we didn't keep track of device tokens on the backend so thought we'd need a followup frontend issue for that. I talked to Keith and he mentioned how we did store these in the table `push_notification_tokens`. 

I've gone ahead and added a new domain event which is fired when these are created and has webhook support for sending out to customer.io. I also created a migration which backpopulates the old data (no need for a script since it was pretty simple) and I added a zapier zap to test adding these using the domain event publisher. It properly sent the data and I hooked that all up.

```
insert into domain_event_publishers (event_types, webhook_url, import_historic_events) values ('{PushNotificationTokenCreated}','https://...', true);
```

Note: The customer.io integration doc talked about storing when a device was last used and keeping it up to date but we don't currently update that anywhere (we have a field for it just not currently setup). Seemed out of scope here to handle fixing that somehow (not sure if we can infer when last used from request data?) but when we do we can fire a domain event and publish in a similar manner to keep it up to date on the customer.io side.

### Migrations
 * 20190814181155_devices-webhook-support

### Environment Variables
 * No change


mikethetike:
The `token_source` at the moment is always `expo`. All the rows in the production db have this as the source

Timo614:
Hmm they had wanted me to use this API: https://customer.io/docs/api/#apitrackcustomersdevices_update

But it appears that is limited to just platforms 'ios' and 'android' so I guess we may need to store these as events or customer data. I guess the former makes the most sense.

Timo614:
I didn't have to modify anything here just had to modify the zapier zap but it's now setup to log these additional devices as events.

Screen Shot 2019-08-15 at 10 43 42 AM

I'll make a bug card about us not having a means to keep the device token last used data up to data (we have a field but never update it) but aside from that this should be set. There's not much else I can do on a device side as we're limited by their API to 'ios' and 'android'.

Timo614:
I will make one more modification though -- I did notice that you can delete your push notification tokens.. if someone adds a future publisher later on it may try to fetch it based on the id and fail if it doesn't exist in the system any longer. I'll make it optional for loading there.

PR #1459, opened 2019-08-15
Sort before grouping

mikethetike:
### Description:
Adds a sort before any places we call `group_by` because it only groups as long as the key is consecutive.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/domain_event_publishers.rs
```diff
@@ -144,7 +144,8 @@ impl DomainEventPublisher {
                 domain_events::created_at,
             ))
             .distinct()
-            .order_by(domain_events::created_at)
+            .order_by(domain_event_publishers::id.asc())
```
Timo614: Before this was setup to order by created_at prior to the publisher id so we'd always return the earlier domain events. I guess in most cases though we'll have few domain events in need of publishing so wouldn't matter.

If we had it the old way (created at, domain publisher id), then as we're importing a new domain publisher for the first time it would front load all of the old events to publish ahead of any of the new events. With this new approach it will depend on the ids of the domain publishers (whichever is randomly assigned the earlier) so there's less control about the order things are processed.

One neat feature of the old approach is we could always make it search for `domain_events::created_at.desc()` as the first dimension and it would process the new domain events prior to the historic import ones. The downside of that feature though is that not all systems may like getting events out of order (customer.io is fine as we're passing a timestamp and it's okay with getting events for customers it hasn't seen yet).

mikethetike: The problem arises actually a few lines later at 158. I had always assumed that `itertools` `group_by()` would group over the whole collection, but that is not the case. It only groups when the key is the same for consecutive rows. So in this case, line 156-158 would return many different groups if the domain_event_publisher_ids where alternating by created_at.

This became apparent in this situation: https://github.com/big-neon/bn-web/issues/1717#issuecomment-521179463

See the comment at ticket_instances line 322

db/src/models/ticket_instances.rs
```diff
@@ -319,6 +319,7 @@ impl TicketInstance {
             ))
             .order_by(events::event_start.asc())
             .then_order_by(events::name.asc())
+            .then_order_by(events::id.asc())
```
mikethetike: So in this case the tickets were being returned in order of event_name, but in the issue above, there were multiple events with the same name. The group_by at line 327 would return multiple groups alternating between the two event ids with the same name.

PR #1461, opened 2019-08-16
Event search query improvements

Timo614:
### References Issues:
Asana issue

### Description:
```
Event listings are not searchable by:
City
State - full spelling
State - abbreviation
```

The issue was about adding city and state searching. I opted not to do a partial search as this is on the events search so otherwise anyone typing "New" looking for some set of events would suddenly get all of the New York, New Jersey, etc. events 

Added tests for different cases. You can test by city itself, state, state abbreviation, "city, state", and "city state" with any number of extra spaces.

Not sure if there's a better way to handle the state data. I looked for a crate for address info but didn't find one so opted for the json approach.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/address.rs
```diff
@@ -0,0 +1,38 @@
+use serde_json;
+use utils::errors::DatabaseError;
+
+#[derive(Deserialize)]
+pub struct UnitedStates {
```
mikethetike: Probably better to just call this `State` or `StateProvince`. I'd also add a `Country` lookup at the same time to allow for internationalization

db/src/data/us-states.json
```diff
@@ -0,0 +1,53 @@
+[
```
mikethetike: I wonder if it's worth adding a submodule to a git repo that has this data already. I did a quick search and found [this](https://github.com/JohnBendfeldt/country-province-data/blob/master/countries.json) repo for an NPM module.

db/src/models/address.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub name: String,
+    pub abbreviation: String,
+}
+
+pub struct Address {
```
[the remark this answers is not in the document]
mikethetike: This might even be a candidate for a DDD service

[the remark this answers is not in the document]
Timo614: DDD service? Some sort of remote domain data service? One reason we may want to keep it internal to the crate is that we have to do the lookup with each event search request and since the data isn't likely to change.

db/src/models/address.rs
Timo614: Sounds good.

db/src/data/us-states.json
Timo614: Sounds good, yeah makes it easier for us to use some of the other data as needed in the future from there.

How were you thinking of structuring where the submodules would go? Something like `db/src/external/{submodules_here}`

Then just having the include_str! macro that calls this just reference that location instead? If so sounds good will make that modification.

Timo614: I wonder if drone will just work with the submodule or if we'll need to modify it so it properly does the update to fetch the code. I'll give it a shot and see how things go.

mikethetike: Think it's a fairly simple change to the clone command

db/src/models/address.rs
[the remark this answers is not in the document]
mikethetike: More just a naming thing. This is in the `models` folder, I'd say it's more of a service as described in the Domain Driven Design books. Perhaps it should live in a `services` folder

Timo614:
Should be set now.

I've done the following:
- Switched to a submodule for the domain data on country info
- Modified the country lookup service a bit in light of the new data
- Added the ability to search by country in addition to the city and state
- Cleaned up the implementation a bit

PR #1463, opened 2019-08-21
Improve location event search

Timo614:
### References Issues:

Asana issue
Improvement on https://github.com/big-neon/bn-api/pull/1461

### Description:

https://github.com/big-neon/bn-api/pull/1461 was recently merged in and it works for the most part at doing city, state, and country searches across their names and abbreviations for the events. I realized that the current implementation has a bug though for state lookups where the state actually matches a country code. So Morocco for example is "MA" but so is "Massachusetts" the US state. The logic would default to checking the US country for the state if one isn't found but since Morocco is found the logic currently stops there. 

With these changes you search for MA now it'll search the db for events in "MA" (state and country), "Morocco" (country), or "Massachusetts" (state) (defaults to assuming US country code unless a user is passed into the API and then their last order's venue is used to determine default country for the province code lookup functionality).

Also encapsulated and better tested some of the country lookup logic.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Need to make a quick modification today prior to this merging. Would be good to limit the number of separate components searched for a match to avoid bad data leading to long searches for countries.

Timo614:
Going to look into / fix the db CPU issue from the release today with this branch as well. Guessing it's related to the logic to lookup the last venue for the user since logged in users alone were experiencing this -- will look for optimizations.

Timo614:
I've pushed up a change which I think should fix some of the instability we've noticed since the change.

- The logic avoids doing a country search on empty queries (previously it was using a sanitized search query field with the value '%' on empty) which was wasteful and now doesn't perform any logic when there's no query provided at all
- Caches the country lookup service as part of the service locator logic 
- Limits segments searched during country and state lookup

There is one flaw with the new approach but it should be going away very soon. The drip logic to send a reminder drip to a phone number loads up a service locator from a config so it can use the shortener service. Since the service locator loads up a cached copy of the country lookup that causes it to load per drip message to phone recipients. I'll be filing a PR to remove the in code drips sometime over the next week (was a transitional step for customer.io) so once that's in we won't have any per request callers of ServiceLocator::new (only the one stored on the appstate shared between requests).

PR #1466, opened 2019-08-30
Settlement report modifications

Timo614:
### References Issues:
Asana issue

### Description:
This branch attempts to add automatic settlement functionality as per the asana requirements. It adds a CLI command to generate settlement processing domain actions but normally they'd be created during creation / updating.

### Concerns:

- The migrations are not very backwards compatible as the structure of these has changed a bit since the original implementation. I drop the settlement_transactions and modify the settlements to remove the non null user_id. Asana made it sound like the original implementation didn't work for generating so from that it sounds like we shouldn't have any data in production if someone could confirm via this prior to production: `select * from settlements;` and just seeing if anything comes back.

### Notes:

- I removed some endpoints for the settlements that are of no use to us now (preparing, etc) so we will need to update the bn-api-node repo to take that into account.
- Opted to change the settlement_transactions approach to a settlement_entries one that breaks apart by ticket type as per the desired output. Order items link back to their settlement_entry_id.

#### Timezone
Timezone can be modified. It defaults to "America/Los_Angeles" if not provided. This could lead to periods where the settlement is not processed due to the settlement window moving.
For example if it was PT but then switched to ET the window it records will be until 11:59:59 PM PT and then once updated it'll start at 12:00:00 AM ET which is 9:00:00 PM PT the previous night so some overlap. Modifications in the opposite direction offset-wise will lead to the windows not overlapping.
To avoid this from becoming an issue if a previous settlement period exists the logic uses its end time as the start of the new settlement window. So if it ran from 12:00 AM ET to 11:59 PM ET and then switched to PT the first week would run from 12:00 AM ET to 11:59 PM PT. Subsequent weeks will have the correct dates. This is just done to avoid any chance of the overlap or break in the schedule windows.

#### Questions / followup issues / out of scope:
- ~Changes to the settlement type may lead to missed data or duplicate data (if rolling then suddenly post event or vice versa). How do we plan on handling that sort of situation? Should we clear past settlements and rerun them over the entire period? So if we switched from post event to rolling the subsequent settlement should include all rolling transactions yet to be reported from earlier? Rolling to post event once a post event happens it shouldn't include any already reported (from the rolling setting earlier) transactions?~ Normally a clear transaction is done -- from discussion below we're going to think about this further / out of scope here
- ~What behavior should occur if a settlement has 0 transactions? I'm still recording it in the DB as it shows the job does run but should the API filter out any 0 transaction settlements? Or let them show on the frontend but have it clearly marked 0 transactions somehow so the organizations don't start thinking the job itself is broken?~ Recording these as 0 transactions if nothing happens in them
- ~How do we want to handle an event having its end date changed after a settlement has been processed for the post-event settlement? Figure out what data wasn't included in the initial export and settle those new transactions since?~ not possible
- ~Should first settlement include all previous settlements? Or do we want to process all of the old settlements as if we had done such previously? IE: use the first order date for an organization to determine settlement~ No, it only uses transactions from the period (updated the logic and tests)
- ~Right now settlements only run if there has been at least one order for a given organization (so newer organizations don't have a giant list of empty settlements associated). Is there some way we should indicate on an organization they have moved on / deactivated so we stop processing settlements?~ Not currently an issue / out of scope at the moment
- ~Are the fees listed here correctly calculated? I based it off of the original implementation https://github.com/big-neon/bn-api/blob/1.0.224/db/src/models/settlements.rs#L254 which used the fee order item values directly via the fee item unit_price_in_cents (not broken apart by client and company). Or is this not the intention here but rather to just show the client fees alone and not company? Also should we be including the credit card fee?~ Should be using the client fees, updated the function and the tests
- ~How should refunds be handled? If the user refunds prior to the settlement should we not include that in the online sale total? What should we do if a settlement occurs and a refund occurs in a subsequent settlement period?~ Should handle refunds that happen in the settlement period but any after have to be manually handled (we could update a previous settlement as an order item now has its settlement id if we wanted in the future creating a settlement adjustment but likely out of scope)

## Release Details:

We'll need to run the CLI command to generate the missing domain actions as it's created normally when organizations are added.

```
cargo run --bin api-cli schedule-missing-domain-actions
```

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/bin/api-cli.rs
```diff
@@ -47,96 +44,33 @@ pub fn main() {
...
-                .about("Backpopulate temporary user data"),
+            SubCommand::with_name("schedule-missing-domain-actions")
+                .about("Creates any missing reoccurring domain actions"),
         )
         .get_matches();
 
     match matches.subcommand() {
         ("sync-purchase-metadata", Some(_)) => sync_purchase_metadata(database, service_locator),
         ("sync-spotify-genres", Some(_)) => sync_spotify_genres(config, database),
-        ("regenerate-interaction-records", Some(_)) => regenerate_interaction_records(database),
```
Timo614: I've removed these two one-time CLI commands / general cleanup since they won't be of use in the future. We'll want to make sure we've run them in production already but they went out with last release so should be safe to remove this here (so long as we run them prior to this merging if not already).

Krakaw:
Just a few quick notes:

1. Once a company is set to a settlement type, they can't update that later (it's defined in their contract)
2. When we run the first settlement maybe we have a custom start / end date so we can show either a 0 value settlement or a custom value settlement that will get us up to current dates so the settlements can continue from there.
3. We need a way to add `Manual Adjustments` to the settlements. Such as an out of band refund (Refund via Stripe that we need to claw back / setup fees / etc etc)

Fees and refunds I will comment on when I get into the office.

Good luck for the hurricane! Stay safe!

Timo614:
> When we run the first settlement maybe we have a custom start / end date so we can show either a 0 value settlement or a custom value settlement that will get us up to current dates so the settlements can continue from there.

What sort of flow were you thinking here? That we don't run settlement jobs for an organization until they've setup an initial settlement? Were you thinking they'd go to the admin section and then specify the dates there for the initial one?

We could also go to do a settlement and if we notice that one hasn't been run yet we:
- Run the settlement we intended to do (in the settlement window Monday, 12:00 am - Sunday, 11:59 pm (LOCAL TIME ZONE of the Org))
- Run an additional settlement from the beginning of time to the start of the settlement window we just did to catch all prior to the settlement logic

> We need a way to add Manual Adjustments to the settlements. Such as an out of band refund (Refund via Stripe that we need to claw back / setup fees / etc etc)

Sounds good will add that.

Timo614:
> Once a company is set to a settlement type, they can't update that later (it's defined in their contract)

Just noticed there was an asana issue for adding this functionality added for this (to modify at will for super admins). Can add it back sometime today but we may want to hammer out the rules there with switching it.

@harrisonlemon can you confirm the behavior for the following? Can adjust here or in follow up issues as needed for any changes.

- Changes to the settlement type may lead to missed data or duplicate data (if rolling then suddenly post event or vice versa). How do we plan on handling that sort of situation? Should we clear past settlements and rerun them over the entire period? So if we switched from post event to rolling the subsequent settlement should include all rolling transactions yet to be reported from earlier? Rolling to post event once a post event happens it shouldn't include any already reported (from the rolling setting earlier) transactions?
- What behavior should occur if a settlement has 0 transactions? I'm still recording it in the DB as it shows the job does run but should the API filter out any 0 transaction settlements? Or let them show on the frontend but have it clearly marked 0 transactions somehow so the organizations don't start thinking the job itself is broken?
- How do we want to handle an event having its end date changed after a settlement has been processed for the post-event settlement? Figure out what data wasn't included in the initial export and settle those new transactions since?
- Should first settlement include all previous settlements? Or do we want to process all of the old settlements as if we had done such previously? IE: use the first order date for an organization to determine settlement
- Right now settlements only run if there has been at least one order for a given organization (so newer organizations don't have a giant list of empty settlements associated). Is there some way we should indicate on an organization they have moved on / deactivated so we stop processing settlements?
- Are the fees listed here correctly calculated? I based it off of the original implementation https://github.com/big-neon/bn-api/blob/1.0.224/db/src/models/settlements.rs#L254 which used the fee order item values directly via the fee item unit_price_in_cents (not broken apart by client and company). Or is this not the intention here but rather to just show the client fees alone and not company? Also should we be included the credit card fee?
- How should refunds be handled? If the user refunds prior to the settlement should we not include that in the online sale total? What should we do if a settlement occurs and a refund occurs in a subsequent settlement period?

Timo614:
The tests were failing because of the facebook crate and some issues it has with deadcode, naming, etc. I've made some adjustments and had the code ignore the dead code where possible so the tests can pass. If the facebook future changes come in before this I'll revert the commit to reduce conflicts.

Timo614:
Sounds good re the comments above about the questions -- will make those modifications today.

Timo614:
I'm going to push up another quick change to how this is displayed to make it easier for the frontend (grouping up by the event_id). I had originally kept it based on the original where the events are returned by a separate array on the response but I think grouping up by the events will make displaying across platforms more consistent and less error prone (having to then do the grouping up by hand).

Timo614:
@harrisonlemon I was just thinking and one more case just popped out at me that's not currently covered by the report. Based on our above conversation the tickets that are refunded prior to this running are not reflected in the report (they are refunded afterall). If just the fee was refunded (and not the face) though the report would look pretty weird because the total wouldn't make sense without surfacing a revenue_share quantity I'd think.

Wondering what we should do there. I don't think we currently allow only the fee to be refunded for a ticket order just thinking ahead since I'm guessing that's something we'll be addressing some time over the next few months given the previous discussions we've all had about refund improvements.

harrisonlemon:
hi @Timo614  That is an excellent question. We currently don't allow for just fee to be refunded, but this is something that we'd like to support down the line. 

Since this is a bit of corner case situation, I think we are ok leaving the report as-is (esp for now) knowing that the qty*rev share will look off if only the fee is refunded, and not the face ticket price. I think adding an additional rev share column might make the report a bit busy. We can explain the fee difference directly to the client if this comes up, especially in the short term

Timo614:
Sounds good, I adjusted the response so the total reflects when just the fee is refunded so once we do support that it should be backwards compatible. Also added the fee_sold_quantity to the settlement_entries so we can more easily see the state at the time the settlement was run (versus trying to figure out had a fee been refunded when the total was calculated if later on we ever iterate on the design to flag that a different quantity of fees was refunded here or break it out further).

api/src/bin/api-cli.rs
Timo614: Sounds good, will add these back now.

Timo614: Added these back (confirmed they still worked locally as well)

PR #1467, opened 2019-08-30
Fix facebook web login

mikethetike:
### Description:
Fixed a bug where a field intended for future use did not have a serde default

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
This is the problem: https://github.com/serde-rs/serde/issues/1030#issuecomment-522278006

PR #1473, opened 2019-09-10
Add genre field to artist table and added script to create and initialise database 

toiletspider:
### Description:
Artist Database: Add genre to artists table in database.

Created shell script to create and initialise database. This should decrease the fragmented setup for the database creation.

### Migrations
* 20190905000000_add_genre_field_to_artist

### Environment Variables
 * No change


db/migrations/20190905000000_add_genre_field_to_artist/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table artists
```
mikethetike: Please add to the summary of the PR

db/src/models/artists.rs
```diff
@@ -35,6 +35,7 @@ pub struct Artist {
     pub created_at: NaiveDateTime,
     pub updated_at: NaiveDateTime,
     pub other_image_urls: Option<Vec<String>>,
+    pub genres_id: Option<Uuid>,
```
mikethetike:
```suggestion
    pub main_genre_id: Option<Uuid>,
```

db/migrations/20190905000000_add_genre_field_to_artist/up.sql
```diff
@@ -0,0 +1,2 @@
...
+    add genres_id UUID NULL REFERENCES genres(id);
```
mikethetike:
```suggestion
    add main_genre_id UUID NULL REFERENCES genres(id);
```

db/migrations/20190905000000_add_genre_field_to_artist/up.sql
toiletspider: Added migrations names

Timo614: We should add a down migration for this change as well.

toiletspider: Added migration name to the RP description

db/migrations/20190905000000_add_genre_field_to_artist/up.sql
toiletspider: Renamed table colom

db/src/models/artists.rs
toiletspider: Renamed table colom

Krakaw:
@toiletspider please also add the issue link that this PR references.

PR #1474, opened 2019-09-11
Adjust user venue lookup query for performance issues

Timo614:
### References Issues:
Ongoing performance issue

### Description:
This branch attempts to fix an ongoing performance issue we're having with event search. The problem relates to the way the last venue is fetched for the user. Mike noticed that moving the conditions into the join statement appears to resolve the issue. Likely some problem with the way the left joins are otherwise fetching more rows than they should prior to filtering.

There may be additional improvements we can make on the query but patching this as we noticed it fixed our issue for now.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/users.rs
```diff
@@ -160,27 +160,39 @@ impl User {
...
                         .and(ticket_instances::status.eq_any(vec![
                             TicketInstanceStatus::Purchased,
                             TicketInstanceStatus::Redeemed,
-                        ])))
-                    .or(orders::status.eq(OrderStatus::Paid).and(
+                        ])),
+                )),
+            )
+            .left_join(order_items::table.on(order_items::event_id.eq(events::id.nullable())))
+            .left_join(
```
mikethetike: I think this can be an inner_join

Timo614: Quickly tried the inner join and tests fail with no returned country for the users. I think it limits the entire query not just the records associated with the order_items left join.

Issue #1477, opened 2019-09-11
Domain event publishing: support timezone for user creation


Timo614:
We currently do not store the user's timezone. To figure out the timezone for customer.io the import logic originally used the user's last event they went to, ordered an order from, or expressed interest in.

The problem with this approach is the customer is created prior to their orders so while it's fine for imports new customers have nothing stored for this field as they haven't had a chance to purchase anything.

We should either figure out the timezone from IP data or via passing in from the frontend.

I'm going to file a PR to remove the timezone field for now which we can add back once we support it here. I would suggest a new column on the user table where we cache this and we can run a script to then figure it out for any previous user's via the original approach (pending improvements on that query).

PR #1478, opened 2019-09-11
Remove timezone from customer creation publish event

Timo614:
### References Issues:
References: #1477 

### Description:
This branch removes the timezone from customer creation's publish event. The reasoning is two part:

1. The query being used has been causing production issues so will need to be adjusted
2. We should likely gather this information at the time of customer creation either via the frontend or via an IP lookup service as these customers won't have events associated so the logic it was using only worked for imported users

#1477 will be used to readd the timezone field to the event just removing this as a stopgap temporarily while we sort out the associated issue to reduce likelihood of production problems / it would otherwise hold up 

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
I pushed up an additional change just to remove the method that triggered all the issues so we don't inadvertently end up using it.

I created https://github.com/big-neon/bn-api/issues/1477 as a way for us to readd this functionality in a more performant way (not sure how we want to schedule it) but this right now has some serious flaws.

PR #1482, opened 2019-09-13
Box office sales report modifications

Timo614:
### References Issues:
Asana https://app.asana.com/0/1136728345391137/1132088822459437

### Description:
The asana issue in question is asking for a box office sales report. We actually built this out based on an earlier spec so I'm only modifying it and fixing some bugs related to it and how the fees are calculated.

#### Note:
- We currently do not take fees from the box office so I can't fully test the behavior around them. I do have the tests add event and ticket level fees though so once we support it they will break (requiring us to edit them to reflect the new fee costs).
- I'm removing refunded quantity from the quantity when displaying -- this may cause a report to show a different total if they rerun it for a certain time period after additional refunds (as it's based on order date).
- Using client fee (not company share) here for the revenue share similar to the settlement report

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
> Looks good. Does the frontend need updated

The frontend right now is just a placeholder mentioning how it's not available yet -- the bn-web side was put off before so it'll need to be handled. The Asana issue has separate tasks though so should be fine tracking there I think.

PR #1486, opened 2019-09-19
Code / hold / event code char limit

Krakaw:
### Description:
Removes the minimum and maximum limits for the access, promo and event codes
### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
Just a note, there was a constraint on the codes table that did a length check, but I couldn't find the migration that introduced it. I ran this: `ALTER TABLE codes DROP CONSTRAINT IF EXISTS codes_redemption_code_check;` because this PR has also gone in as a hotfix.

PR #1491, opened 2019-09-24
Further improve unit test coverage

Timo614:
### References Issues:
Relates to: #1112

### Description:
This adds additional tests to improve unit test coverage. I also fixed settlements to only allow modification of the settlement type to super admins.

I've added coverage for many of the files going through each of the lcov's results and adding where necessary. There were quite a few methods not covered so I tried to focus on those first but added some additional cases for some of the other tests. I'm rerunning the lcov tests now (it takes a very long time since it clears the files and builds from scratch) so should be able to update the PR once this is done. Hopefully this should help with any future refactoring we may ever end up doing.

I can try to augment this further if we don't merge it right away otherwise can make additional PRs time permitting in the future as we iterate on the associated main coverage issue.

#### There are still significant holes in our coverage mostly around the following models:
- orders
- paging
- reports
- ticket_instances
- ticket_types
- users
- wallets

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/wallets.rs
```diff
@@ -21,25 +21,6 @@ pub struct Wallet {
     updated_at: NaiveDateTime,
 }
 
-#[derive(Serialize, Copy, Clone)]
-pub struct WalletId(Uuid);
```
mikethetike: Was hoping to use this pattern more but never got around to it, so happy that it's gone so that it matches the rest

Timo614:
One gotcha with this card is it fixes a bug I noticed with the settlement type selection that we allowed regular admins (not supers only) to edit the field. A bn-web component will be needed to lock down the selection of the type for this scope when this merges.

To make this less risky / self contained without introducing modifications requiring frontend changes I'm going to move the scope I added here for this action to an admin level capability and once the frontend has support for using the scope we can easily restrict it further. Will push that up now so we can merge this to increase our coverage.

PR #1492, opened 2019-09-25
Optimize fans query

Krakaw:
### Description:
This is more a hotfix PR for improving the fan search.

The fans query is causing pod / node issues from the response times. This serves to reduce the time it takes to load the fans list.

I removed the user interested count from the query as well as the tests, this can be reverted when we enhance this PR by using the organization_interactions.
### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/organizations.rs
```diff
@@ -759,18 +759,23 @@ impl Organization {
...
+                        orders::table.on(order_items::order_id
+                            .eq(orders::id)
+                            .and(orders::status.eq(OrderStatus::Paid))),
+                    )
+                    .on(order_items::event_id
+                        .eq(events::id.nullable())
+                        .and(order_items::item_type.eq(OrderItemTypes::Tickets))),
             )
-            .left_join(
+            .inner_join(
```
mikethetike: I think this needs to move into the brackets above

PR #1493, opened 2019-09-25
Adjust refundable flag to take box office orders into account

Timo614:
### References Issues:
https://app.asana.com/0/1138868228851596/1139476696018190

### Description:
This issue fixes a bug that was noticed in asana related to box office orders being ineligible for refunds. It also fixes another issue with the logic where a previously refunded ticket may appear as refunded incorrectly (the refunded_tickets join was not joining on order_items so any refunding of that ticket instance_id would trigger it to show).

There's another issue for the wrong user being used on the order pages: https://github.com/big-neon/bn-web/issues/1724 so that's not being handled here. The API side properly mentions it's a box office order and gives the on behalf of user id back to the frontend so just a matter of us using the right ID I believe.

I wasn't able to test transferring tickets as it appears we no longer allow that via the web portal and I don't know how to use the mobile app locally. Not sure if we've had issues there though as the asana issue just mentions how the user should be able to function normally not that they have run into problems outside of this bug with refunding.

#### Screenshots / behavior
Screen Shot 2019-09-25 at 2 31 25 PM
Screen Shot 2019-09-25 at 2 31 29 PM
Screen Shot 2019-09-25 at 2 31 34 PM
Screen Shot 2019-09-25 at 2 31 42 PM

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Took two screenshots of the behavior after I set this up as such:
- Setup a ticket type of only 2 tickets
- Setup an order and refunded one ticket, transferred the other
- Placed a subsequent order for the refunded ticket

## Initial order with one marked as refunded one transferred
Screen Shot 2019-09-26 at 4 47 44 PM

## New order showing the same ticket was used and it's set to purchased
Screen Shot 2019-09-26 at 4 47 56 PM

## Subsequent transfer of both tickets to a new user (showing both are capable of being transferred again)

Screen Shot 2019-09-26 at 4 59 37 PM
Screen Shot 2019-09-26 at 4 59 47 PM

PR #1497, opened 2019-09-27
Fix logic for admin in event show

Krakaw:
### Description:
Fix the logic for showing past events for users with privileges
### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/events.rs
```diff
@@ -275,9 +275,10 @@ pub fn show(
     };
 
     let event_ended = event.event_end.unwrap_or(times::infinity()) < dates::now().finish();
-    if (!is_user_admin && event.publish_date.unwrap_or(times::infinity()) > dates::now().finish())
-        || event.deleted_at.is_some()
-        || event_ended
+    if !user_has_privileges
+        && (event.publish_date.unwrap_or(times::infinity()) > dates::now().finish()
```
Krakaw:
```suggestion
        || (event.publish_date.unwrap_or(times::infinity()) > dates::now().finish()
```

PR #1500, opened 2019-10-03
Improve activity endpoint performance

Timo614:
### References Issues:
https://app.asana.com/0/1136728345391137/1142348999982865

### Description:
This branch attempts to fix the underlying issues with the activity endpoint. I've improved some of the logic to reduce the number of queries, to avoid subqueries, etc.

I've made some adjustments to the activity to remove the some extra logic that wasn't needed. Originally the activity returned records when the user made a note for another user, etc. but we don't even show the profile unless they have an order so they were inaccessible (unless they made one and then they had a ton of records associated that aren't related to their actual customer).

Note:

I experimented a bit with some additional indexes (on top of the ones I added in the migrations) but they seemed to actually slow down the query. Not sure if in production they may be more beneficial though -- for now I'm holding off but may be worth seeing how performance is affected by these later.

```
CREATE INDEX index_order_items_order_id_event_id ON order_items(order_id, event_id); 
CREATE INDEX orders_users_on_behalf_of_user_id_user_id on orders (coalesce(on_behalf_of_user_id,user_id));
```

### Migrations
 * db/migrations/20191003020555_index_improvements

### Environment Variables
 * No change


Timo614:
It would be useful if someone with a larger production-like dataset could confirm if things look good performance wise as develop and beta both are much smaller. It should perform a lot better but just want to confirm at that scale if we may still be missing an index.

Was able to optimize it quite a bit compared to master:

### In branch
Screen Shot 2019-10-03 at 1 14 08 PM
Screen Shot 2019-10-03 at 1 14 26 PM

### Master (plus reverting index changes)
Screen Shot 2019-10-03 at 1 25 56 PM
Screen Shot 2019-10-03 at 1 26 33 PM

Krakaw:
The code looks good, I'm trying to verify with a local instance of prod data, but running into import issues. Will update as soon as I have fixed the issues.

PR #1505, opened 2019-10-08
Adjust free payment logic to properly set voucher as external order type for box office orders

Timo614:
### References Issues:
https://app.asana.com/0/1138868228851596/1132088822459437

### Description:
While testing the box office report on develop Jay ran into an issue with the logic returning null for a field that should have been populated. The bug appears to be an issue with how we handle free payments for box office orders. We have a flag passed in that it's an external order but we weren't setting the `external_payment_type` on the order in that scenario. The fix here is to just set that flag.

I've added a migration to update the existing orders.

### Migrations
 * `db/migrations/20191008130051_fix_box_office_voucher_orders_external_payment_type`

### Environment Variables
 * No change


Timo614:
Set as DNM pending access to develop db. My guess is the orders likely predated the external_payment_type so may just not have info about how they were generated. Will have to confirm once I can take a look at them.

Somewhat unrelated to the first bug but seems to make sense to deal with all these issues related to this field being unpopulated at once.

Timo614:
The results found above actually were just my query not excluding the draft orders so we should be set here. Will push up one modification to the migration to just ignore Drafts and should be good to go.

PR #1506, opened 2019-10-08
Improve orders endpoint

Timo614:
### References Issues:
Loosely relates to https://app.asana.com/0/1136728345391137/1142083339118197 

### Description:
Noticed the orders endpoint was rather slow on develop so set to improving it by removing a significant amount of n+1 queries while still DRYing up the logic to avoid two separate methods to fetch the DisplayOrder logic.

The biggest offender was order search. We pull down the whole list of orders and then individually called the logic to build the display order fetching all of their items, etc. with this new approach their IDs are determined from the search criteria and then the display orders are built from that minimizing the number of queries performed.

The logic also fixes: 
- A bug where a box office user's orders list would include every order they've ever made. I confirmed this doesn't cause any issue for box office users by logging in as one with only that role: I still see the manage orders section as they're using the admin orders API for that event in the side panel (not their own list of orders to manage).
- Super admin role was missing from some logic that treated admins differently so quickly added it in a few places (`User#is_admin()`, etc).

I don't think there should be too much risk associated with this improvement but we may want to push it out separately from some of the other improvements to avoid confusion if a bug does occur.

Gists revisions to show diff from master (initial) and this branch (the final state):
Orders endpoint (several orders belonging to other users via box office purchases removed): https://gist.github.com/Timo614/81c66616eabe00bc6f81b57ccfce42c4/revisions
Event orders list (no difference):
https://gist.github.com/Timo614/176859bee0be67bac92201861bce31a5/revisions
Event orders list with query (no difference):
https://gist.github.com/Timo614/41389c8ad26594f30c10753ccb7bec7e/revisions

### Screenshots
Orders endpoint: 299ms -> 115ms
Order management for event no query: 630ms -> 136ms
Order management for event with query: 276ms -> 112ms

#### Before
Screen Shot 2019-10-08 at 6 20 23 PM
Screen Shot 2019-10-08 at 6 18 35 PM
Screen Shot 2019-10-08 at 6 19 51 PM

#### After
Screen Shot 2019-10-08 at 6 05 41 PM
Screen Shot 2019-10-08 at 6 06 58 PM
Screen Shot 2019-10-08 at 6 07 29 PM

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/orders.rs
```diff
@@ -1848,174 +1772,327 @@ impl Order {
...
+                o.on_behalf_of_user_id,
+                o.paid_at,
+                o.platform,
+                o.checkout_url,
+                o.checkout_url_expires,
+                ARRAY_AGG(DISTINCT p.payment_method) FILTER (WHERE p.payment_method IS NOT NULL) as payment_methods,
+                ARRAY_AGG(DISTINCT p.provider) FILTER (WHERE p.provider IS NOT NULL) as providers,
+                CAST(COALESCE(SUM(oi.unit_price_in_cents * oi.quantity), 0) as BigInt) as total_in_cents,
+                CAST(COALESCE(SUM(oi.unit_price_in_cents * oi.refunded_quantity), 0) as BigInt) as total_refunded_in_cents,
+                ARRAY_AGG(SUBSTRING(orgs.allowed_payment_providers::text from 2 for char_length(orgs.allowed_payment_providers::text) - 2)) FILTER (WHERE orgs.allowed_payment_providers IS NOT NULL) as allowed_payment_providers,
```
Timo614: I was unable to deserialize an array of arrays so ended up returning it as a string and deserializing manually by splitting on the comma and parsing into the enum.

PR #1507, opened 2019-10-09
Restrict which platforms a ticket type is available on

mikethetike:
### Description:
Add three flags: `app_sales_enabled`, `web_sales_enabled`, `box_office_sales_enabled` to allow a ticket type to only be available on those platforms

### Migrations
 * db/migrations/20191004121212_restrict_ticket_type_sales

### Environment Variables
 * No change


mikethetike:
I don't really want to mess with the sales pipeline. I think it's sufficient to restrict it on visibility to the front end

PR #1509, opened 2019-10-11
Facebook attempt 3

mikethetike:
### Description:
Added a few more steps to test Facebook. Still unable to complete the flow because Facebook need to grant us `create_event` but hopefully that will be available after the next review

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/external/facebook.rs
```diff
@@ -318,10 +320,42 @@ pub fn create_event(
...
+            let res: DatabaseError = e.into();
+            return Err(res.into());
+        }
+    };
+
     let fb_event = FBEvent::new(
         data.category.parse()?,
         event.name.clone(),
-        event.additional_info.clone().unwrap_or("".to_string()),
+        event.additional_info.clone().unwrap_or(event.name.clone()),
```
Timo614: Noticed above we validate to confirm additional_info has a value so could switch this to just an `unwrap()`. Not a huge deal either way though.

mikethetike: Thanks

PR #1510, opened 2019-10-12
Slug fetch logic

Timo614:
### References Issues:
https://app.asana.com/0/1138868228851596/1139843346254324

### Description:
These changes add a new `/slugs/{id}` endpoint which is used to fetch data based on slugs. The response changes depending upon the slug type of the slug. I've included some examples below. Decided on this approach after discussion with Keith re the requirements here. This approach allows for one slug endpoint to return a variety of data and then the frontend can use the type returned to figure out how to display the page. Let me know if that's not ideal though.

Logic:
- URL returned by events show has been modified to point to the new `/events`-less frontend format
- Allows for redirects (via a frontend, can adjust as needed used a json response with a redirect field to designate this) if multiple slugs point to the same event, organization, or venue types (https://moz.com/learn/seo/duplicate-content)
- Associates multiple venues with a slug via a habtm setup for city type and properly deregisters from cities when venue is modified
- Returns an enum response with a type so the slug loading page can handle the data correctly
- Uses order search to keep things DRY with parameters to properly filter down (should allow us to add pagination easily in the future if ever desired)
- While slug type may seem a bit redundant given the main_table parameter it was done this way to allow multiple events to link back to a city so they could be fetched together for the response
- Is backwards compatible for old slug links for `/events/{id}` -- this is to reduce friction as I'm sure we'll have callers there for the time being

Note:
- The create and update responses no longer return the slug since they are returning the Event itself. I don't think we use the slug in the response there aside from integration tests but worth mentioning.
- If a city was configured to have a different name than other cities (San Fran vs San Francisco, etc.) they would be treated as separate cities for the purposes of linking venues. Keith mentioned we're using an API to populate the values so they should all be standard but worth keeping in mind. We may want to iterate on this in the future to make sure they are normalized but outside of scope here.
- The migration for the city slug does not check against state and country differences (like the logic does once things are in motion) -- opted to avoid making the query complex as we don't have multiple instances of the same city in that environment.
- Like the events migration for adding slugs this adds additional text to the back of the auto generated slugs for venues and organizations. This was done to reduce collision risk during the initial migration (I noticed some names were duplicated across the venues and organizations). We can easily modify these slugs in the db later but I noticed in production we had quite a few duplicate org and venue names so there will be collisions initially / worth having those characters.

Outside of scope:
- Does not currently handle updating the slug if the event, org, venue, name changes but the logic does support redirects for once we implement that. Could add if it's worth it but want to confirm with the team first.
- We don't have a sitemap yet so didn't add a method for fetching all of the slugs. Should be pretty easy for us to pull down based on the slug_type to get a list of city slugs and then grab all of the primary slugs for events, organizations, and venues.

### Examples

Response is similar to events index but with additional information. Returns type regardless. Let me know if we want to change the show response for an event. I kept it as it has been to reduce friction with existing implementations but could modify it as needed.

#### Organization by slug:
```
 curl --header "Content-Type: application/json" http://localhost:9000/slugs/jazzy-1571069504
```
https://gist.github.com/Timo614/d75365e690f1f6c9179fd8e711de6ff1

#### Venue by slug:
```
 curl --header "Content-Type: application/json" http://localhost:9000/slugs/test-venue-1571069504
```
https://gist.github.com/Timo614/b5c9bad5cf65678914568647cd8ad1af

#### Event by slug:
```
 curl --header "Content-Type: application/json" http://localhost:9000/slugs/it-s-my-party-at-test-venue-1571069504-city-K0M9L
```
https://gist.github.com/Timo614/5fa8bf6f2b62e609fa4e5c30e055e452

#### City by slug (using generated data so "City" was set as the city field hence its slug):
```
 curl --header "Content-Type: application/json" http://localhost:9000/slugs/city
```
https://gist.github.com/Timo614/a7f4cd4defe5bf2e26265c744e6fa0f0

#### Tested deleting the slugs and confirming the migration logic repopulated it:
```
delete from slugs where slug_type <> 'Event';

-- Ran insert and update logic for slugs

bigneon=# select slug, name from venues v join slugs s on v.slug_id = s.id;
            slug             |         name          
-----------------------------+-----------------------
 test-venue_1571069504-26f01 | Test venue_1571069504
 test-venue_1571069504-74978 | Test venue_1571069504
 test-venue_1571069504-86488 | Test venue_1571069504
(3 rows)

bigneon=# select slug, name from organizations o join slugs s on o.slug_id = s.id;
          slug          |       name       
------------------------+------------------
 jazzy_1571069504-5d269 | Jazzy_1571069504
(1 row)
```

### Migrations
 * db/migrations/20191009173405_add_slugs

### Environment Variables
 * No change


Timo614:
> Looks good. Would it not be better to have a City table with a single slug and a many-many to Venue? The find_first_for_city slug might be a bit tricky to work with.

Was thinking that originally but opted to go with this approach re: Keith's asana comments as it allows for us to just tag items as needed across any dimension (with a new slug_type). We'd need to then somehow link the existing venues to the city either by using a city_id field in place of the name or by doing a search at the time the page is loaded. Given how much was already changing seemed like the safest bet to just stick with the tag approach for now on the Venues table. May be worth revisiting in the future though.

Timo614:
Note: when this goes live to production it would be good if someone with write access could rename some of the org and venue slugs to remove the appended extra characters where collisions aren't occurring. 

The logic was done similar to the original events migration where additional letters were appended to the end of the tag to avoid collisions. I was going to not at first but I noticed several venues and organizations with the same names so opted to do this to avoid any unforeseen behavior. 

It won't hurt anything to keep the extra characters just there's no need for them in many cases.

PR #1515, opened 2019-10-16
Better error handling and skip over old ticket transfer domain events

mikethetike:
### Description:
We had a migration at one point that added extra domain events that weren't necessary, this code handles that case where there are orphans

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/domain_events.rs
```diff
@@ -497,13 +507,13 @@ impl NewDomainEvent {
...
             .to_db_error(ErrorCode::InsertError, "Could not insert domain event")?;
 
-        jlog!(Info, &format!("Domain Event: {} `{}` on {}:{}", self.event_type,
-            self.display_text, self.main_table, self.main_id.map(|i| i.to_string()).unwrap_or_default()),{
-                "domain_event_id": result.id,
-                "event_type": self.event_type.clone(),
-                "main_table": self.main_table.clone(),
-                "main_id": self.main_id,
-                "event_data": self.event_data
+        jlog!(Info, & format ! ("Domain Event: {} `{}` on {}:{}", self.event_type,
```
mikethetike: ??!!

PR #1518, opened 2019-10-17
Ice add sitemap.xml to endpoint

toiletspider:
### References Issues:
SEO Phase 3 - Implement Sitemap.xml for better Google Page Indexing

### Description:
Create a api endpoint: `/sitemap.xml` which will return a sitemap for search engines containing all active events urls that are not private. Each time a event is added, a domain event is triggered which will ping google to read the updated sitemap.xml.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/events.rs
```diff
@@ -1098,6 +1098,20 @@ impl Event {
         )
     }
 
+    pub fn find_all_active_events(conn: &PgConnection) -> Result<Vec<Event>, DatabaseError> {
+        DatabaseError::wrap(
+            ErrorCode::QueryError,
+            "Error loading event via venue",
```
Krakaw:
```suggestion
            "Error loading all active events",
```

Cargo.lock
```diff
@@ -273,12 +273,12 @@ dependencies = [
 
 [[package]]
 name = "bigneon_api"
-version = "1.0.256"
+version = "1.0.258"
```
mikethetike: Rather exclude this file from future commits

api/src/controllers/sitemap_gen.rs
```diff
@@ -0,0 +1,26 @@
...
+use db::Connection;
+use domain_events::executors::GenerateSitemapEventExecutor;
+use errors::BigNeonError;
+use std::env;
+use utils::gen_sitemap;
+
+pub fn index((connection): (Connection)) -> Result<HttpResponse, BigNeonError> {
+    //find all active events
+    let events = Event::find_all_active_events(connection.get())?;
+    let _urls = events
```
mikethetike:
```suggestion
    let urls = events
```

api/src/domain_events/executors/generate_sitemap.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
+
+impl GenerateSitemapEventExecutor {
+    pub fn perform_job(
+        &self,
+        action: &DomainAction,
+        conn: &Connection,
+    ) -> Result<(), BigNeonError> {
+        Ok(())
```
mikethetike: Is this for future use?

api/src/controllers/sitemap_gen.rs
```diff
@@ -0,0 +1,26 @@
...
+        .collect::<Vec<_>>();
+    let sitemap_xml = gen_sitemap::create_sitemap(&_urls)?;
+
+    Ok(HttpResponse::Ok()
+        .content_type("text/xml")
+        .body(sitemap_xml)
+        .into())
+}
+
+fn create_url(slug: &str) -> String {
```
mikethetike: Rather get the env::var from a Config passed into the `index` method

api/src/routing.rs
```diff
@@ -405,6 +405,9 @@ pub fn routes(app: &mut CorsBuilder<AppState>) -> App<AppState> {
         r.method(Method::GET).with(venues::index);
         r.method(Method::POST).with(venues::create);
     })
+    .resource("/sitemap.xml", |r| {
+        r.method(Method::GET).with(sitemap_gen::index);
```
mikethetike: rename this controller to just `sitemap` so that it matches the route (kinda)

api/src/utils/gen_sitemap.rs
```diff
@@ -0,0 +1,38 @@
+use errors::BigNeonError;
+use sitemap::structs::UrlEntry;
+use sitemap::writer::SiteMapWriter;
+use sitemap::Error;
+use std::io::{Cursor, Read};
+
+pub fn create_sitemap(urls: &[String]) -> Result<String, BigNeonError> {
+    let mut output = Cursor::new(Vec::new());
+    let sitemap_writer = SiteMapWriter::new(&mut output);
+    let mut urlwriter = sitemap_writer.start_urlset()?; //.expect("Unable to write urlset");
```
mikethetike: These comments can be removed

db/src/models/enums.rs
```diff
@@ -154,7 +154,8 @@ string_enum! { DomainEventTypes [
     TicketTypeCreated,
     TicketTypeSalesStarted,
     TicketTypeSoldOut,
-    TicketTypeUpdated
+    TicketTypeUpdated,
+    GenerateSitemap
```
mikethetike: This event should be in past tense and try to keep them in alphabetical order

db/src/models/events.rs
```diff
@@ -1034,6 +1034,20 @@ impl Event {
         )
     }
 
+    pub fn find_all_active_events(conn: &PgConnection) -> Result<Vec<Event>, DatabaseError> {
+        DatabaseError::wrap(
```
mikethetike: You can also use the `.to_db_error()` function which is cleaner, depending on your preference

api/src/routing.rs
toiletspider: the crate that I'm using is also called sitemap. To avoid importing with "as" the controller name was changed

api/src/utils/gen_sitemap.rs
toiletspider: Comments have been moved to a ApplicationError

api/src/domain_events/executors/generate_sitemap.rs
toiletspider: yes, it have been implemented

db/src/models/events.rs
toiletspider: DatabaseError::wrap is the most used. Could change all errors in another pull request for uniformity.

PR #1520, opened 2019-10-17
Add cargo audit back

mikethetike:
### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
Will reintroduce when the audit doesn't fail

PR #1523, opened 2019-10-18
Fix display order query around refunds

Timo614:
### References Issues:
https://app.asana.com/0/1138868228851596/1145552426833993

### Description:
The display order logic was modified recently to remove some inefficiencies with how it loads up orders. Previously it did a single query for every order in the order list but now loads them all at once and returns their data. In the process of doing so I also removed some other n+1 queries where possible to reduce the number we had to load. Unfortunately I had switched the payments to a left join to aggregate up the different payment types and payment methods. This had the effect though of causing multiple records to be returned as soon as a refund was made. Ended up making a fix similar to other spots where we've done this in the queries / reports where the logic left joins on a grouped up version of the table by order id. I've added a unit test for this case as well.

#### Examples

With 3x $30 tickets, $0.10 fee, $1.00 event fee

##### Before (with bug)
Screen Shot 2019-10-18 at 2 22 56 PM

Screen Shot 2019-10-18 at 2 22 46 PM

##### After
Screen Shot 2019-10-18 at 2 31 56 PM

Screen Shot 2019-10-18 at 1 58 00 PM

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/orders.rs
```diff
@@ -1834,18 +1834,26 @@ impl Order {
...
             LEFT JOIN events e ON oi.event_id = e.id
             LEFT JOIN organizations orgs ON e.organization_id = orgs.id
-            LEFT JOIN payments p ON p.order_id = o.id
+            LEFT JOIN (
+                SELECT
+                    p.order_id,
+                    ARRAY_AGG(DISTINCT p.provider) FILTER (WHERE p.provider IS NOT NULL) as providers,
+                    ARRAY_AGG(DISTINCT p.payment_method) FILTER (WHERE p.payment_method IS NOT NULL) as payment_methods
+                FROM payments p
+                WHERE p.status = 'Completed'
```
mikethetike: Just checking whether this should include "Refunded" as well?

Timo614: Sounds good -- I think right now we handle refunds via the payment processor that the initial payment occurs with and the logic uses the first returned (used to do this previous as well) as the API endpoint for DisplayOrder only returns that first payment method or payment processor but this would future proof it so will quickly do that (but going to keep the endpoint set to return just the one processor / payment method to avoid any risk / frontend changes needed).

PR #1530, opened 2019-10-22
Rust 1.38.0 support

Timo614:
### References Issues:
N/A

### Description:
This branch simply updates our rust version to 1.38.0. I'll be pushing up another branch in the future to update actix but wanted to keep these separate since that upgrade isn't blocked here.

Just requires some minor formatting changes and the drone updated to use that version.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Good catch -- I haven't used docker before but from their dockerfile links (https://hub.docker.com/_/rust) it sounds like I'll only need to modify the logic here (https://github.com/big-neon/bn-api/blob/master/Dockerfile#L1) to use: "1.38-slim" in place of the "1.33.slim"

Will need someone who uses docker to confirm things look good though as I'm not too familiar with that setup.

Timo614:
Can someone approve the drone build? Asking the team if they can get me admin access for my other github user account (drone needs a lot of read / write access so using my nine account so it's limited to the bigneon org).

Timo614:
I'll take a swing at the drone issue today / OpenSSL problems -- it builds locally without issue so seems most likely to be an issue with the environment.

Timo614:
It looks like the dependencies updated in https://github.com/big-neon/bn-api/pull/1481 were the culprits so once I updated reqwest and expo-server-sdk drone was able to proceed. Waiting for the build to finish and will rerequest a review.

Things look good on my end testing locally (did have a failure with globee but I'd wager that's an environment issue as drone passed with the testing).

PR #1538, opened 2019-10-24
Added stub functions

Krakaw:
### Description:
Added a stub handler so we could implement function endpoints before the functionality has been finalized.

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/status.rs
```diff
@@ -7,6 +7,11 @@ use log::Level::*;
 
 static mut IS_OK: bool = false;
 
+// This function always returns a 200
+pub fn stub(_connection: (Connection)) -> Result<HttpResponse, BigNeonError> {
```
Timo614: May be worth just removing the connection parameter just to avoid any overhead with its extractor if it's not used.

Krakaw: Good thinking

PR #1540, opened 2019-10-25
Add genre list endpoint that contains the name and id of the genre

toiletspider:
### Description:
Add genre list endpoint that contains the name and id of the genre

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/genres.rs
```diff
@@ -8,3 +8,17 @@ pub fn index(connection: Connection) -> Result<HttpResponse, BigNeonError> {
     Ok(HttpResponse::Ok()
         .json(json!({"genres": &genres.into_iter().map(|g| g.name).collect::<Vec<String>>()})))
 }
+
+#[derive(Serialize)]
+struct GenreListItem {
+    #[serde(default, deserialize_with = "deserialize_unless_blank")]
+    id: String,
```
Timo614: It ends up making the Uuids into strings when it's serialized so you don't need this to be a String here.

Can also probably remove those default serde tags since it's not deserializing.

api/src/controllers/genres.rs
```diff
@@ -8,3 +8,17 @@ pub fn index(connection: Connection) -> Result<HttpResponse, BigNeonError> {
...
+    #[serde(default, deserialize_with = "deserialize_unless_blank")]
+    name: String,
+}
+
+pub fn list(connection: Connection) -> Result<HttpResponse, BigNeonError> {
+    let genres = Genre::all(connection.get())?;
```
Timo614: It may be worth just adjusting the index method if that's not already used by anything (I think it was added ages ago with a first genres iteration but may never have been consumed by the frontend). Could get datadog to see if anything has called it in the last month via the logs and if not just replacing its simple list of genres names with this output.

api/src/controllers/genres.rs
toiletspider: Changed id to UUID and removed deserialize macro

api/src/controllers/genres.rs
toiletspider: /genres/ is indeed not being used. Changed endpoint to use url instead

api/src/controllers/genres.rs
```diff
@@ -2,9 +2,16 @@ use actix_web::HttpResponse;
...
 
 pub fn index(connection: Connection) -> Result<HttpResponse, BigNeonError> {
     let genres = Genre::all(connection.get())?;
     Ok(HttpResponse::Ok()
-        .json(json!({"genres": &genres.into_iter().map(|g| g.name).collect::<Vec<String>>()})))
+        .json(json!({"genres": &genres.into_iter().map(|g| GenreListItem{id: g.id, name: g.name }).collect::<Vec<GenreListItem >>()})))
+}
+
+#[derive(Serialize)]
+struct GenreListItem {
```
mikethetike: If you make this pub you won't have to recreate it in the test

PR #1560, opened 2019-11-05
Facebook cleanup

mikethetike:
### Description:
Finally a working Facebook build. There may be some test code that still needs to be removed, but this should be prod ready

### Migrations
 * db/migrations/20191105121212_add_facebook_event_id

### Environment Variables
 * No change


db/Cargo.toml
```diff
@@ -19,7 +19,7 @@ path="src/bin.rs"
 
 [dependencies]
 backtrace = "0.2"
-diesel = {version="1.4", features = ["postgres", "uuid", "chrono","numeric", "serde_json", "r2d2"]}
+diesel = {version="1.4", features = ["postgres", "uuid", "chrono","numeric", "serde_json", "r2d2", "64-column-tables"]}
```
mikethetike: The events table now has more than 32 columns, so we have to add this feature. It's been mentioned that this can increase compile time, so let's keep an eye out for that

PR #1561, opened 2019-11-05
Revert rust 1.38.0 upgrade for now

Timo614:
### Description:
Reverts 1.38.0 -- this is just in case we're unable to fix the docker issue for now so QA can still test what's in flight.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Closing in favor of #1562

Krakaw:
If the build doesn't work then let's re-open this.

PR #1597, opened 2019-11-18
Event level access improvements

Timo614:
### References Issues:
https://app.asana.com/0/1149047845558135/1141186260130601

### Description:
PR discussed improving how invites are handled. 

> If the user has already accepted the invitation to the org, any further invitations to events will automatically be approved.
> If the user hasn't accepted the first invitation, any one of the invitations will work and accept all event invitations that have been created for that organization.

Logic modified:
- Adjusted how invites are accepted to allow multiple to exist at once for a organization user
- Blocked the ability to send an invite to an existing organization user (someone can just modify the role from the team page)
- Adjusted the logic for inviting existing promoters / readonly promoters so they are immediately accepted without an email
- Adjusted the logic for the event users endpoint to properly scope the roles to just that event in the response

Note:
- Currently it still sends invite emails for every promotion as they pile up. The logic could be adjusted to only send an email for the first invite if that's better but wasn't part of the original requirements.

Out of scope: 
- We may want to alter the roles with the invite to a singular role. The backend allows multiple rows to be set but the frontend only ever displays one (it shows an empty dropdown if the user is part of multiple) 

### Migrations
 * db/migrations/20191118161545_add-event-users-index - migration to add an index by the user_id used for the deletion of event user records in situations where the user's role is set to a cross organization role

### Environment Variables
 * No change


db/tests/unit/organization_invites.rs
```diff
@@ -113,6 +221,46 @@ fn create_with_validation_errors() {
...
+    );
+    match invite.commit(connection) {
+        Ok(_) => {
+            panic!("Expected validation error");
+        }
+        Err(error) => match &error.error_code {
+            ValidationError { errors } => {
+                assert!(errors.contains_key("user_id"));
+                assert_eq!(errors["user_id"].len(), 1);
+                assert_eq!(errors["user_id"][0].code, "uniqueness");
```
mikethetike: At a later stage we should maybe create an enum for these

PR #1608, opened 2019-11-26
[Hotfix] Fix Push Notifications Data field

mikethetike:
### Description:
Expo is returning Bad Request for our push notifications since the 15th of November. The only change I've seen is the addition of this data field. My theory is that it was putting in an explicit `data: null` instead of leaving the field off entirely. This PR hopefully fixes that

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
Can we close this in favour of merging in the full deploy that has this fix in?

mikethetike:
Yes

PR #1615, opened 2019-11-28
Get latest FB token if there are multiples

mikethetike:
### Description:
There may be cases, mainly in test environments, where there is more than one facebook token stored for a single user. Ideally we would want to throw an error in this case, but I'm not sure where the problem originally initiates. This code now will receive the last token created if there are multiples

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/external_logins.rs
```diff
@@ -91,6 +91,7 @@ impl ExternalLogin {
                 .filter(external_logins::user_id.eq(user_id))
                 .filter(external_logins::site.eq(site))
                 .filter(external_logins::deleted_at.is_null())
+                .order_by_desc(external_logins::created_at)
```
Krakaw:
```suggestion
                .order_by(external_logins::created_at.desc())
```

mikethetike:
Thanks, was switching to another branch as was too hasty to push

PR #1622, opened 2019-12-03
Add rows affected check for multiple processes

mikethetike:
### Description:
Currently if there are multiple processes running domain events, there is a chance that they will publish multiples. This checks for rows affected in the `ON CONFLICT` clause of the insert statement and doesn't process the webhook if that is the case

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Approved -- had thought if there was a conflict though wouldn't it error out and fail the entire logic?

mikethetike:
@Timo614 No, it is has an on_conflict clause which prevents the insert. It may have been better to propagate the error and let the caller decide. Will see how this works out and if it fails QA I'll use that approach

PR #1624, opened 2019-12-03
Adjust logic for different platforms

Timo614:
### References Issues:
https://app.asana.com/0/1151234524692625/1139530207175514

### Description:
Fixes issue with platforms for orders on the system. We were historically only setting a platform if the user agent matched a set number of ones we're searching for and we'd mark the rest as `null` in the database. This logic now sets Box Office orders as `BoxOffice`, sets unknown orders as `Web`, and the others are determined in the same way as before.

Example after making a box office and web purchase:
Screen Shot 2019-12-03 at 5 11 41 PM

### Migrations
 * `db/migrations/20191203195932_fix-order-platforms` fixes the existing order platforms set on the system

### Environment Variables
 * No change


Timo614:
Updated the PR's code and stripe is suddenly failing for odd reasons. Perhaps something bad with develop or a temporary issue on the Stripe side?

PR #1625, opened 2019-12-04
Email: fixed history sent_at date not showing

toiletspider:
The sent_at at field for a broadcast is now set when a broadcast are started. This will allow the front end to show the history of the broadcast email sent. It is currently showing null.

### References Issues:
Email Ticket Buyers Directly within Studio

### Migrations
 * No Migrations

### Environment Variables
 * No change


mikethetike:
What is the reason for setting the sent_at time?

toiletspider:
> What is the reason for setting the sent_at time?

I have updated the initial comment

db/src/models/broadcasts.rs
```diff
@@ -114,6 +114,18 @@ impl Broadcast {
...
     }
 
+    pub fn set_sent_at(
+        id: Uuid,
+        datetime: NaiveDateTime,
+        connection: &PgConnection,
+    ) -> Result<Broadcast, DatabaseError> {
+        let broadcast = Broadcast::find(id, connection)?;
+        diesel::update(&broadcast)
+            .set(broadcasts::dsl::send_at.eq(datetime))
```
mikethetike: Add the `updated_at` field here please

api/src/domain_events/executors/broadcast_push_notification.rs
```diff
@@ -66,7 +67,10 @@ impl BroadcastPushNotificationExecutor {
         .map(|aud| aud.0)
         .collect_vec();
 
+        // set the broadcast count (the number of messages that will be sent)
         Broadcast::set_sent_count(broadcast_id, audience.length() as i64, conn)?;
```
mikethetike: Not involved in this PR, but the original intention I had for this field was to be updated as each email is sent, so that the user can see the progress of the broadcast. Doing this at this stage could result in an incorrect value, e.g. if some of the emails fail. 

Alternatively we could alter the logic of the `broadcasts.show` method to query the number and status from the `domain_actions` that are linked to this broadcast

db/src/models/broadcasts.rs
Krakaw: This is the wrong field, this is to specify when to send the broadcast.

Krakaw:
This is not required, `send_at` is optional and correct as is.

PR #1639, opened 2019-12-16
Lockdown ticket pricing deletion for ones with sales

Timo614:
### References Issues:
https://app.asana.com/0/1151234524692625/1149281525498849

### Description:
This branch addresses the associated issue by locking down ticket pricing / scheduled price changes for ones with sales so they can no longer be deleted. Previously we allowed them to be deleted and it caused confusion where the clients were unaware of how the ticket pricing ever existed when they checked their event audit so this change was requested.

The logic here also:
- Prevents the creation of a new ticket pricing if no changes have been made during update (just returns the cloned ticket pricing struct)
- Adds a previous_ticket_pricing_id field which tracks the record this ticket pricing was generated from when replaced during update
- Uses that aforementioned field to link back and figure out if any previous iteration of the ticket pricing records had sales

Note: this is a forward only change as legacy records will not know they had sales in previous iterations as we did not link the new to the old records as they were replaced. Jessi mentioned this was fine.

### Example

Some price changes scheduled, first is prevented from being deleted as sales exist:
Screen Shot 2019-12-16 at 10 58 26 AM

After some sales for the second:
Screen Shot 2019-12-16 at 10 59 18 AM

Change the price triggering a new price schedule to be created (linking to previous) and delete third:
Screen Shot 2019-12-16 at 10 59 40 AM

Add new price schedule not changing the others:
Screen Shot 2019-12-16 at 11 00 10 AM

### Migrations
 * db/migrations/20191213201353_ticket-pricing-add-previous-ticket-pricing-id

### Environment Variables
 * No change


Timo614:
The web side was merged it sounds like but this side needs a code review if someone could take a look when you have a chance.

db/src/models/ticket_pricing.rs
```diff
@@ -42,6 +42,39 @@ pub struct TicketPricingEditableAttributes {
...
 impl TicketPricing {
+    pub fn associated_with_active_orders(&self, conn: &PgConnection) -> Result<bool, DatabaseError> {
+        select(exists(
+            order_items::table
+                .inner_join(orders::table.on(orders::id.eq(order_items::order_id)))
+                .filter(orders::status.eq(OrderStatus::Paid).or(orders::expires_at.ge(dsl::now)))
+                .filter(
+                    sql("
+                        order_items.ticket_pricing_id in
+                        (WITH RECURSIVE ticket_pricing_r(id) AS (
```
Krakaw: Very cool.

PR #1647, opened 2020-01-02
Modify card expiration date to 2030 to avoid issues around expired card use

Timo614:
### References Issues:
N/A

### Description:
This just updates the expiration date to 2030 for the integration tests to avoid the expired card errors we were seeing.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Should confirm the tests pass prior to merging -- I had it set to 2025 for testing locally but switched to 2030 just to give a bit more time. Figured could just use drone to confirm that date won't have any issues with the Stripe API (given how far away it was).

PR #1656, opened 2020-01-10
Add parent name to ticket types response

Timo614:
### References Issues:
https://app.asana.com/0/1151234524692625/1110258272725761

### Description:
It was mentioned in the associated asana issue that there was some issue getting the start date for some ticket types. Child ticket types have null start dates so that issue was unrelated to any API bugs. It was noticed that the new view UI would need to call for the parents of each ticket type in order to display their names. This patch simply does that query prior to returning the response so they can just render the parent's name more easily.

Screen Shot 2020-01-10 at 3 32 42 PM

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
This doesn't add it to the response on bn-api-node. I'm not sure what needs to be changed there will leave that for the frontend.

PR #1661, opened 2020-01-15
Update event broadcasts as well as their related domain action

msbodetti:
### References Issues:
https://app.asana.com/0/1153926226800754/1155691976771297/f
https://app.asana.com/0/1156907105711882/1157220402023137
### Description:
As well as adding the route to update, we needed to update the pending domain actions to reschedule them if it is allowed.
### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
When a broadcast is created, it creates a `DomainAction` that actually executes the broadcast. We need to add the update to the domain action as well for this to have any real effect.

msbodetti:
Thanks for the feedback @Krakaw. I'll bring this up in standup today as well so everyone is on the same page regarding this.

Krakaw:
Closing this to fix some tests.

PR #1663, opened 2020-01-15
Abandoned cart retargeting emails

Timo614:
### References Issues:
https://app.asana.com/0/1153926226800754/1153920343475198

### Description:
The associated PR has a desire to add abandoned cart retargeting. There are quite a few requirements here so best to check the asana issue for the full list. In general the idea is that once a cart has been abandoned users can get an email 24 hours later asking them if they'd like to complete it. We only intend to send one of these per event to customers and limit to 1 per 7 day periods to avoid spamming them for other event emails.

Note:
- "You already have tickets in your cart" is the message I'm showing regardless of what event the tickets belong to when you go to duplicate and they already exist in the cart, correct? I did update the other error text for multiple events in one cart but want to confirm since it wouldn't make sense to use that I'd think.
- Limited retargeted carts to ones belonging to events with no override status or with one set to one of the following EventOverrideStatus::PurchaseTickets, EventOverrideStatus::TicketsAtTheDoor, EventOverrideStatus::Free
- Opted to do the ticket type and redemption code checks on an individual order level versus trying to get creative with the query. I don't think we'll have too many abandoned carts to dig through and the query would be rather complex if I tried to do it all at once. Can circle back if there are performance issues ever experienced and optimize further.
- If a ticket type has ended and a new one is following it fails duplication and does not automatically create a cart with the new child ticket type even if that has inventory (ie: Early Bird -> Regular situations etc with limited release since different pricing etc)
- Price schedule changes or changes to the current ticket type (like additional fee) may cause the cost to be different for the new order (not validating against the old price to new)
- Filtering out box office abandoned carts from being included
- We may want to iterate on the valid_for_duplicating method for orders at some point. There were quite a few conditions to place on the logic here so I opted to DRY it up to use for both the retargeting selection and eventual duplication. For the former though it makes sense to cache ticket types since many are reused across multiple customer orders so passed it in as a cache to the DRYed up method. Feels a bit dirty but seemed like the best tradeoff without needing to repeat logic between the two or take a potential performance hit.
- A lot of logic is changing here. I'm going on holiday for the next week so it may make sense to hold off on merging this until I get back even if QA approves. There's also the chance for a performance hit when the logic runs given how much it has to fetch from the DB, etc. Locally it seems fine but I don't have a DB nearly as large as production. May be worth Mike giving the logic a shot locally.

### Examples

New text requested for scenario where you have event in cart and are buying from other event:
Screen Shot 2020-01-15 at 4 38 28 PM

Regenerating domain actions (logic added for new one plus some of the old refactored a bit) doesn't create more than intended (note: this was from a fresh dev clean db and mocha run so added both and multiple runs didn't add subsequent, also ran at 4:45 ET and its set to run the first one at 5 ET in 11 minutes):
```
bigneon=# select domain_action_type, created_at, scheduled_at from domain_actions order by created_at desc limit 5;
         domain_action_type         |         created_at         |        scheduled_at        
------------------------------------+----------------------------+----------------------------
 RetargetAbandonedOrders            | 2020-01-15 21:45:54.920179 | 2020-01-15 22:00:00
 SendAutomaticReportEmails          | 2020-01-15 21:45:54.913434 | 2020-01-16 12:00:00
 Communication                      | 2020-01-15 21:37:51.762161 | 2020-01-15 21:37:52.469239
 Communication                      | 2020-01-15 21:30:36.698886 | 2020-01-15 21:30:37.356597
 SendPurchaseCompletedCommunication | 2020-01-15 21:30:34.817937 | 2020-01-15 21:30:35.643017
(5 rows)
```

After running the task once:
```
bigneon=# select domain_action_type, created_at, scheduled_at, status from domain_actions order by created_at desc limit 5;
    domain_action_type     |         created_at         |        scheduled_at        | status  
---------------------------+----------------------------+----------------------------+---------
 RetargetAbandonedOrders   | 2020-01-15 22:00:00.392619 | 2020-01-15 23:00:00        | Pending
 RetargetAbandonedOrders   | 2020-01-15 21:45:54.920179 | 2020-01-15 22:00:00        | Success
 SendAutomaticReportEmails | 2020-01-15 21:45:54.913434 | 2020-01-16 12:00:00        | Pending
 Communication             | 2020-01-15 21:37:51.762161 | 2020-01-15 21:37:52.469239 | Success
 Communication             | 2020-01-15 21:30:36.698886 | 2020-01-15 21:30:37.356597 | Success
(5 rows)
```

## Release Details:
Run `cargo run --bin api-cli schedule-missing-domain-actions`

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/orders.rs
```diff
@@ -131,6 +132,311 @@ struct MatchData<'a> {
...
+
+        let mut carts_to_retarget = Vec::new();
+        for cart in abandoned_carts {
+            if cart.valid_for_duplicating(Some(&ticket_types_map), conn)? {
+                carts_to_retarget.push(cart.clone());
+            }
+        }
+
+        // From the remaining carts, filter out:
+        // 1. Users who have received a targeted email in past 7 days
```
mikethetike: This filtering can happen at Customer.io, might be easier to leave that logic there

Timo614:
> One concern is that if one of these actions fails and doesn't create the next action, we won't be signalled and the process could stall for a long time. Perhaps it's time to look into a cron kind of scheduler.

Agreed. Especially for jobs like this and the ticket counts report which are more consistent (versus the ones that run each day at a given time in the event timezone etc). Perhaps we should build out some basic tooling for Super level users to given insight into the domain actions at a glance (what's queued, passed, failed, stuck, etc.).

> There was another way we could have approached this. We could send an event to customer.io when the user starts a cart and then one when it is completed (This is already in place). Then using customer.io's logic we could trigger the sending if we haven't received the order completed event. The problem with that approach is that there isn't really a way to check if the tickets are still available, so I'm happy with going with your approach here

Yeah was thinking this before as well. I couldn't think of any good way to really get around that since they'd have no context of our system state.

I'll make the adjustments mentioned above moving the filtering into customer.io.

db/src/models/orders.rs
```diff
@@ -131,6 +132,311 @@ struct MatchData<'a> {
...
+                Some(cart.id),
+                None,
+                None,
+            )
+            .commit(conn)?;
+        }
+        Ok(carts_to_retarget)
+    }
+
+    pub fn valid_for_duplicating(
```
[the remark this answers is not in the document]
Timo614: Thinking this over:
- The user may have existing items in their cart at the time we're running this check (a rare situation though)
- We may run into issues where people are purchasing on the site and the inventory is taken during the checks. Probably not enough for anyone to notice but if it was down to the last inventory it could prevent it from being purchased by users briefly.
- We'd end up creating a ton of new carts for users although since we'd clear them they'd be reused on the user's next purchase so isn't terrible.

One thing I could do though is just remove the check from the duplicating code and only do it when preparing to send the initial email.

The only issue above that concerns me for the check in general would be the having items in the cart already since it may be a risky to mess with an active session. I think the code will be a bit cleaner though if I just rely on duplicating to check applicability for the actual duplicate logic though so will go with that approach now and can always circle back to adjust further.

db/src/models/orders.rs
```diff
@@ -131,6 +132,311 @@ struct MatchData<'a> {
...
+        let mut carts_to_retarget = Vec::new();
+        for cart in abandoned_carts {
+            if cart.valid_for_duplicating(Some(&ticket_types_map), conn)? {
+                carts_to_retarget.push(cart.clone());
+            }
+        }
+
+        // From the remaining carts, filter out:
+        // 1. Users who have received a targeted email in past 7 days
+        // 2. Users who have received a targeted email for this event in the past
```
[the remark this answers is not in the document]
Timo614: Sent Mike a message and looked over customer.io but I don't think this scenario can be handled by the segmenting tooling there.

I'm going to make the other modification moving the 7 days check into customer.io over here (makes it more customizable as well) but will leave this check in place.

db/src/models/orders.rs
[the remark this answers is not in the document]
Timo614: Opting to hold off on switching to recreating the cart at this time as it's a much bigger dig. It looks like we don't do validations on the ticket type at the time of updating so you can add a ticket type to the cart that is cancelled, has passed its sales date, or even hasn't reached its start date yet. 

I went to add them but it led to a bunch of errors across our tests as we have quite a few tests that add those invalid ticket types to the cart for the purposes of building data. We should probably circle back with a separate ticket to add in those validations and confirm we're not missing more in that cart saving flow.

db/src/models/domain_events.rs
```diff
@@ -209,11 +220,12 @@ impl DomainEvent {
...
+        front_end_url: &str,
         data: &mut HashMap<String, Value>,
         order: Order,
     ) -> Result<(), DatabaseError> {
         if let Some(event) = order.events(conn)?.pop() {
-            Event::event_payload_data(&event, data, conn)?;
+            Event::event_payload_data(&event, front_end_url, data, conn)?;
         }
         data.insert("webhook_event_type".to_string(), json!("purchase_ticket"));
         data.insert("order_number".to_string(), json!(order.order_number()));
```
Krakaw: Will the order_number suffice, or do we need to also include the order_id so that we can call the `/cart/{id}/duplicate` endpoint?

Timo614: That's a good point -- will add the order ID to the payload.

Timo614: I've added the order ID to the payload so should be set here.

mikethetike:
If the user has added something to their cart, and then try to restore a cart, it'd be nice to return a 409 conflict instead of the 422. Then the front end can ask the user if they want to replace it. This code returns a correct error:
image
But it gets wiped out by this blanket `map`:
image

Timo614:
I'll adjust that map so it allows the conflict to go through today.

Timo614:
I've adjusted the response to return conflict if the cart already contains items as per the above feedback.

```
curl --header "Content-Type: application/json" --request POST http://localhost:9000/cart/e47b6413-9571-40f7-a3d4-39c0234e335f/duplicate -i --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI4Mzk0YmI4My0yZjMwLTQ5NmYtYjNlYS0yNGIxYjQyNDc2Y2UiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTgwNDg2OTMzfQ.IdWnoPsFjkB4XC2diZ9j3Yj1SWcFJDorfQ_2CUMfKAQ"
HTTP/1.1 200 OK
content-length: 2011
access-control-expose-headers: x-app-version
vary: Origin
x-app-version: 1.9.2
date: Fri, 31 Jan 2020 16:08:30 GMT

{"id":"16fadefb-0af9-4ef1-bf1a-98371d09fbd2","date":"2020-01-31T15:54:25.619730","expires_at":"2020-01-31T16:09:25.722525","seconds_until_expiry":55,"status":"Draft","items":[{"id":"3aacebb6-9f45-4e4d-98e4-ab64777730cb","parent_id":null,"ticket_type_id":"bcb28657-b633-41d7-b899-68e3ff6c7c58","ticket_pricing_id":"ba102925-4655-4a5d-9f92-d641b789654e","quantity":2,"refunded_quantity":0,"unit_price_in_cents":3000,"item_type":"Tickets","description":"It's my party - Default_Pricing_1580159151","redemption_code":null,"cart_item_status":"Valid","event_id":"66e89938-eae6-4d45-83d1-cf64ee23af79"},{"id":"c205b110-8d58-4920-873b-ec182bdf40b9","parent_id":"3aacebb6-9f45-4e4d-98e4-ab64777730cb","ticket_type_id":null,"ticket_pricing_id":null,"quantity":2,"refunded_quantity":0,"unit_price_in_cents":10,"item_type":"PerUnitFees","description":"Ticket Fees","redemption_code":null,"cart_item_status":"Valid","event_id":"66e89938-eae6-4d45-83d1-cf64ee23af79"},{"id":"1e0213e9-0699-42aa-ac6b-64b5c5efc7eb","parent_id":null,"ticket_type_id":null,"ticket_pricing_id":null,"quantity":1,"refunded_quantity":0,"unit_price_in_cents":100,"item_type":"EventFees","description":"Event Fees - It's my party","redemption_code":null,"cart_item_status":"Valid","event_id":"66e89938-eae6-4d45-83d1-cf64ee23af79"}],"limited_tickets_remaining":[],"total_in_cents":6120,"total_refunded_in_cents":0,"user_id":"8394bb83-2f30-496f-b3ea-24b1b42476ce","user":{"id":"8394bb83-2f30-496f-b3ea-24b1b42476ce","first_name":"System","last_name":"Administrator","email":"nine@tari.com","phone":"8883","profile_pic_url":null,"thumb_profile_pic_url":null,"cover_photo_url":null,"is_org_owner":false},"order_number":"1d09fbd2","paid_at":null,"checkout_url":null,"allowed_payment_methods":[{"method":"Card","provider":"Stripe","display_name":"Card"}],"order_contains_other_tickets":false,"valid_for_purchase":true,"platform":"Web","is_box_office":false,"payment_method":null,"payment_provider":null,"on_behalf_of_user":null,"on_behalf_of_user_id":null}
```

Then duplicating again (which should fail due to the cart already existing and having items):

```
curl --header "Content-Type: application/json" --request POST http://localhost:9000/cart/e47b6413-9571-40f7-a3d4-39c0234e335f/duplicate -i --header "Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiI4Mzk0YmI4My0yZjMwLTQ5NmYtYjNlYS0yNGIxYjQyNDc2Y2UiLCJpc3MiOiJiZy1zYW1wbGUtaXNzdWVyIiwiZXhwIjoxNTgwNDg2OTMzfQ.IdWnoPsFjkB4XC2diZ9j3Yj1SWcFJDorfQ_2CUMfKAQ"
HTTP/1.1 409 Conflict
content-length: 49
access-control-expose-headers: x-app-version
vary: Origin
x-app-version: 1.9.2
date: Fri, 31 Jan 2020 16:08:32 GMT

{"error":"You already have tickets in your cart"}
```

PR #1664, opened 2020-01-17
Update event broadcasts as well as their related domain action

Krakaw:
### References Issues:
https://app.asana.com/0/1153926226800754/1155691976771297/f
https://app.asana.com/0/1156907105711882/1157220402023137
### Description:
Added route to update broadcasts. Some special attention is required with the new validation and updating code. The flow is as follows:

I have removed the ability to update the status via the API other than specifically cancelling, status should only be updated by the model.

If a new `send_at` is set it checks that it is not in the past, it also checks that the broadcast isstill pending. If the send_at is `Some(None)` it sets the send_at to now, otherwise it sets it to the date provided. If there is a new date set, it also updates the `DomainAction`(s) linked to the broadcast.

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/broadcasts.rs
```diff
@@ -235,6 +269,43 @@ impl Broadcast {
             }
         }
     }
+
+    fn send_at_has_not_passed(
+        send_at: Option<NaiveDateTime>,
+        new_send_at: Option<NaiveDateTime>,
+        status: BroadcastStatus,
+        _connection: &PgConnection,
+    ) -> Result<Result<(), ValidationError>, DatabaseError> {
```
mikethetike: This method doesn't look like it can ever return an `Err` result. It might be worth removing the outer `Result<..., DatabaseError>`

Krakaw: You are correct, removed.

PR #1667, opened 2020-01-23
Delete user and add adwords key

mikethetike:
### Description:
I duffed this one and had to add the google ad words key in here as well. 
This PR allows a Superuser to soft delete a user. The user will not be able to log in

### Migrations
 * db/migrations/20200117121212_adwords_key/up.sql 
 *  db/migrations/20200122001212_add_user_deleted_at/up.sql

### Environment Variables
 * No change


api/src/controllers/auth.rs
```diff
@@ -79,8 +79,11 @@ pub fn token(
     // Generic messaging to prevent exposing user is member of system
     let login_failure_messaging = "Email or password incorrect";
 
-    let user = match User::find_by_email(&login_request.email, connection.get()) {
-        Ok(u) => u,
+    let user = match User::find_by_email(&login_request.email, false, connection.get()).optional() {
+        Ok(u) => match u {
+            Some(usr) => usr,
+            None => return application::unauthorized_with_message(login_failure_messaging, None, Some(login_log_data)),
```
Timo614: Not a huge deal either way but why go with the optional approach here versus just letting the 404 fall into the below error handling since they return the same error?

api/src/controllers/auth.rs
```diff
@@ -126,7 +129,6 @@ pub fn token_refresh(
         &user.id,
         &refresh_request.refresh_token,
     )?;
```
Timo614: Should we add a check here for the deleted_at field being present and preventing the user from refreshing the token? And also above when they login via the app?

api/src/controllers/cart.rs
```diff
@@ -412,29 +412,25 @@ fn checkout_external(
 
     let mut guest: Option<DbUser> = None;
 
-    if email.is_some() {
-        guest = DbUser::find_by_email(email.as_ref().unwrap(), conn).optional()?;
+    if let Some(ref e) = email {
+        // Guest can be a deleted user
+        guest = DbUser::find_by_email(e, true, conn).optional()?;
```
Timo614: Should we allow box office purchases for deleted users? If we soft deleted them and aren't allowing login that forces them to have to use guest list checking off only. I wonder if we'd want a soft deleted user to ever purchase again on their email as well.

api/src/controllers/user_invites.rs
```diff
@@ -21,9 +21,9 @@ pub fn create(
     (state, connection, parameters, auth_user): (State<AppState>, Connection, Json<UserInviteRequest>, AuthUser),
 ) -> Result<HttpResponse, BigNeonError> {
     let connection = connection.get();
-
+    let email = parameters.email.trim().to_lowercase();
     // User already exists, so no need to invite them
-    if let Some(_) = User::find_by_email(parameters.email.as_str(), connection).optional()? {
+    if let Some(_) = User::find_by_email(&email, true, connection).optional()? {
```
Timo614: I noticed for organization invites we're not including the deleted users but are including them here. Wondering if we'd want to make this `false` because those users would never be able to login.

api/src/controllers/auth.rs
Timo614: Ah I see below in the extractor code -- they'd get an error as soon as they attempted to use their token so not a big deal if they can refresh it.

api/src/controllers/auth.rs
mikethetike: Yeah, I think I may have been on auto-pilot for this one

api/src/controllers/cart.rs
mikethetike: Yeah I thought about this. It seemed like a weird flow having to explain to a user at the box office why they couldn't buy a ticket, and then having them stand there to reactivate their account or use a different email. Also, any tickets that the user had in their account when they are deleted would still be on the guest list, so I thought it was better to just allow the box office for deleted users

api/src/controllers/user_invites.rs
mikethetike: TBH I don't even know if this code is dead or not. I considered deleted it altogether

PR #1674, opened 2020-01-28
Speed up fan details

mikethetike:
Removed event_count field and the heavy query that came along with it

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/src/models/users.rs
```diff
@@ -700,101 +699,60 @@ impl User {
...
             tickets_owned: i64,
+            #[sql_type = "BigInt"]
             revenue_in_cents: i64,
-            event_count: i64,
         }
-        let result: R = query
-            .get_result(conn)
+        let mut result: Vec<R> = query
+            .get_results(conn)
             .to_db_error(ErrorCode::QueryError, "Could not load profile for organization fan")?;
```
Krakaw: We'll need to test the results here if we want to error on no results found.
`if tickets_sales == 0 && tickets_owned == 0 && revenue_in_cents == 0 { return Err(NotFound) }`

db/src/models/users.rs
```diff
@@ -700,101 +699,63 @@ impl User {
...
 
+        let result = result.remove(0);
+        if result.ticket_sales == 0 && result.tickets_owned == 0 && result.revenue_in_cents == 0 {
+            return Err(DatabaseError::no_results("No profile information for this organization"));
```
Krakaw:
```suggestion
            return DatabaseError::no_results("Could not load profile for organization fan");
```

Krakaw: The DieselError::NotFound appends the `, NotFound` into the error message.
```suggestion
            return Err(DatabaseError::no_results("No profile information for this organization, NotFound"));
```

PR #1678, opened 2020-02-07
Create magic token and link for automatic login

mikethetike:
### Description:
Adds a magic token and link to UserCreated and Transfer* domain event webhooks for use by Customer.io. I also created an API-CLI command `update-customer-io-webhooks` which takes a`site_id` and `api_key` and creates the necessary web hooks for communication.

I also refactored the clap set up on the cli for neatness. 

I decided to move the main webhook data construction from `models::DomainEventPublisher` to the API project because I was ending up moving a lot of the services and structs to the DB crate just to reference them. It makes more sense IMHO to have it in API since a lot of the services needed are web specific.

UPDATE: I've added tokens for temporary users. If they use the token for refreshing, then a new user will be created

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
Left a review request for myself -- will make sure I take a look tomorrow at the code. Sorry for the delay here.

api/src/bin/api-cli.rs
```diff
@@ -157,6 +167,48 @@ fn regenerate_interaction_records(org_id: Option<&str>, database: Database) {
     }
 }
 
+fn update_customer_io_webhooks(site_id: Option<&str>, api_key: Option<&str>, database: Database) {
+    info!("Updating/ensuring Customer.io webhooks");
+    let connection = database.get_connection().expect("Could not connect to database");
+    let conn = connection.get();
+    let publishers = DomainEventPublisher::find_all(conn).unwrap();
+    use bigneon_db::models::DomainEventTypes::*;
+    let event_types: Vec<DomainEventTypes> = vec![
```
Timo614: Not a huge deal but may be worth switching this to use a const from instead the domain event or publisher models. Thinking it may end up being something we often forget to update.

It looks like there was a const added at some point but it's no longer used anywhere from what I could see:
https://github.com/big-neon/bn-api/blob/mikethetike-magic-link/db/src/models/domain_event_publishers.rs#L13

It looks like the PushNotificationCreated domain event type is missing from the list as well (unless intentional): https://github.com/big-neon/bn-api/blob/f7fdd920d76ecbf02a2a484fe6fcef3c38033d94/db/src/models/domain_events.rs

api/src/bin/api-cli.rs
```diff
@@ -36,32 +40,35 @@ pub fn main() {
...
-        .subcommand(SubCommand::with_name("backpopulate-temporary-user-data").about("Backpopulate temporary user data"))
-        .subcommand(
-            SubCommand::with_name("schedule-missing-domain-actions")
-                .about("Creates any missing reoccurring domain actions"),
-        )
-        .subcommand(
-            SubCommand::with_name("generate-genre-slugs").about("Creates any missing genre and city genre slugs"),
-        )
-        .get_matches();
+    let matches = clap_app!(myapp =>
```
Timo614: Much cleaner approach to these :+1:

api/src/bin/api-cli.rs
Timo614: Are we going to add these to each scratch? Wondering if we're going to flood customer.io with a lot of useless data if we don't separate the account out. Probably not a huge deal if it's just Zuzu testing but if it ends up having automated test data at some point it may get tricky with how many records we introduce into the account.

It is nice to be able to turn these on where needed for scratches though especially for newer customer.io email needs, etc.

mikethetike: Could be, but I don't think it'll be any more than what we're getting in Sendgrid

mikethetike:
UPDATE: I've added tokens for temporary users. If they use the token for refreshing, then a new user will be created

mikethetike:
UPDATE: I added a check that prevents admin and org users from getting a magic link

PR #1679, opened 2020-02-07
[DNM] Promo codes use orders not individual tickets as the count

Krakaw:
### References Issues:
https://app.asana.com/0/657855478437518/1155201645123121
### Description:
Uses a count of `order.id` to reduce the promo code availability instead of individual tickets. Also added a new function re refund an entire order (or what is remaining)
### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
Closing for further validation.

PR #1688, opened 2020-02-19
WIP: Chat bot playbook

Timo614:
### References Issues:
N/A

### Description:
WIP: This branch adds the ability to run chat workflows for the purposes of our APP chat bot. It's WIP as it's massive so not all of the main functionality we've described is included in scope here and will need to be iterated on. In addition, this is built off of the redis websockets branch so we'll want to merge that first and rebase. I haven't been able to full test things yet as I need to build out a custom client to interact with the API for our purposes here. I've been doing some basic curl testing and the code compiles, however, it's just the actual chat I can't fully test yet (hoping to resolve this sometime soon as it'll be a pain to debug issues otherwise).

I'll flesh out this PR further as well just wanted to get the initial MVP out there as I've been working on it for some time now.

Questions:
- What permissions should we have for editing these? Currently I'm allowing only admins and supers and only using high level permissions ChatWorkflowWrite, ChatWorkflowRead, ChatWorkflowDelete (versus scopes on an individual ChatWorkflowItem level)

Note:
- These are global chat workbooks currently not tied to a given event, etc. figure we can iterate on that as needed
- We don't have a means to trigger these based on criteria / segments currently. MVP has app trigger them via an endpoint (re discussion with Keith) and then it proceeds to use a chat web socket.
- I only added domain event logging for actions on the chat workflow itself, not for its individual items and responses but could add if there's value. These are the events that fire when the API is hit to modify the chat workflow.
- Currently once a chat workflow is published it's not locked down from editing, since it's super admin only it seemed fine for now but could adjust with future requirements.
- Deletes cascade so deleting a parent in the tree can lead to its children getting deleted.
- When rendering the chat tree I've marked an element as `multiple_references` if it already renders earlier in the recursive tree rendering logic. We may want to iterate on how we display this. Was to prevent infinite loops during rendering if there's a link that leads users back to an already selected choice.
- I added the initial two chat bot conversations via an APLI CLI job. It's pretty ugly as it's a lot of data and it has to link up all the pieces of the chat. Hopefully the frontend will be able to display these in a nicer manner.
- Probably worth removing the chat_workflow_item from the process response logic as we can find it via the chat session / may quickly update that sometime soon

Remaining items / Out of Scope / We'll need to iterate to include:
- UTM based tracking / link display logic (I've been mostly focused on the first two chat items)
- Criteria / segmentation logic for triggering based on arbitrary requirements or state of system
- Display interaction logging data (data is stored on interactions but not currently surfaced)
- Chat websocket logic swallows errors at the moment / will need some way to surface them (perhaps just peppering error! macros about)

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
I've rebased this over the latest develop so it no longer contains commits from the websockets branch and fixed the conflicts caused by that.

PR #1689, opened 2020-02-20
Adjust promo code times used to be transaction based not quantity based

Timo614:
### References Issues:
https://app.asana.com/0/1158142652422330/1155201645123121

### Description:
The team noticed that `max_uses` for redemption codes was based on the quantity of tickets sold not the number of times the redemption code was used. This branch attempts to fix that by altering it to use the number of orders versus the quantity of tickets sold.

Note:
- If an order expires it is no longer considered used
- If an order has its line item fully refunded it no longer is considered used (if I bought 2 tickets and refunded 2)
- I discovered a few bn-web issues while testing this. Included them in the asana issue for if we want to circle back with another issue for them.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
I'm adding one additional change -- adding an optional user to the redemption code info endpoint and having it return number of purchases so that the limit per user can be respected by the frontend. I mentioned in asana that issue and another web one I saw while doing this but not sure if we'll prioritize those as they don't really break the flow too heavily (versus this bug which shuts off a redemption code a lot earlier than intended by the team).

Timo614:
Should be all set, pushed up that commit so we surface the user_purchased_ticket_count now which should allow the frontend to properly limit the max number of tickets to purchase if a user is making multiple orders (right now it always list the limit per user as the max even if previous purchases have brought the user closer to reaching the limit).

Screen Shot 2020-02-21 at 11 48 21 AM

Timo614:
By mistake closed the underlying branch, I've reopened the PR.

Timo614:
Having some issues getting the branch tests to pass due to Github API issues that have been happening all day. I'm restarting the tests as it appears that they may have fixed the issues at this point.

PR #1692, opened 2020-02-25
Additional endpoint caching, switch to middleware approach

Timo614:
### References Issues:

### Description:

This branch modifies the previous caching approach to use middleware instead so it's more DRY. I've done it on the resource level (limiting to GET requests) so that we can have finer controller over which endpoints are cached. In the future we could expand the middleware to keep track of the caching duration, etc. if we wanted custom durations per endpoint as well.

Adding endpoint caching to the GET requests for the:
- artists search, show, index
- events show, index
- genres index
- regions show, index
- venues show, index
- sitemap

### Examples

Hitting the venues index endpoint:
```
curl http://localhost:9000/venues
{"data":[{"address":"address","city":"city","country":"the best country","created_at":"2020-02-25T20:01:55.651962","google_place_id":null,"id":"35b444ab-b76d-4992-8c4f-2612ab3472e0","is_private":false,"latitude":null,"longitude":null,"name":"Test venue_1582660906","phone":"5555555555","postal_code":"2222","promo_image_url":null,"region_id":null,"slug_id":"9330dc88-2b60-4bee-bdfc-935768f2ca3f","state":"CA","timezone":"America/Los_Angeles","updated_at":"2020-02-25T20:01:55.697736"}],"paging":{"dir":"Asc","limit":100,"page":0,"sort":"","tags":{},"total":1}}
```

Hitting it multiple times it's clear from the redis cli logs that if it fails to find the record it creates it with a timeout (currently set by env) and then subsequent requests pull that information and use it until it expires. Once it expires the subsequent request is stored, etc.
```
1582661929.209519 [0 127.0.0.1:54328] "GET" "{\"method\":\"GET\",\"path\":\"/venues\"}"
1582661929.212660 [0 127.0.0.1:54328] "PING"
1582661929.212750 [0 127.0.0.1:54328] "SET" "{\"method\":\"GET\",\"path\":\"/venues\"}" "{\"data\":[{\"id\":\"35b444ab-b76d-4992-8c4f-2612ab3472e0\",\"region_id\":null,\"is_private\":false,\"name\":\"Test venue_1582660906\",\"address\":\"address\",\"city\":\"city\",\"state\":\"CA\",\"country\":\"the best country\",\"postal_code\":\"2222\",\"phone\":\"5555555555\",\"promo_image_url\":null,\"created_at\":\"2020-02-25T20:01:55.651962\",\"updated_at\":\"2020-02-25T20:01:55.697736\",\"google_place_id\":null,\"latitude\":null,\"longitude\":null,\"timezone\":\"America/Los_Angeles\",\"slug_id\":\"9330dc88-2b60-4bee-bdfc-935768f2ca3f\"}],\"paging\":{\"page\":0,\"limit\":100,\"sort\":\"\",\"dir\":\"Asc\",\"total\":1,\"tags\":{}}}"
1582661929.212900 [0 127.0.0.1:54328] "PEXPIRE" "{\"method\":\"GET\",\"path\":\"/venues\"}" "10000"
1582661931.150821 [0 127.0.0.1:54328] "PING"
1582661931.150943 [0 127.0.0.1:54328] "GET" "{\"method\":\"GET\",\"path\":\"/venues\"}"
1582661932.062420 [0 127.0.0.1:54328] "PING"
1582661932.062554 [0 127.0.0.1:54328] "GET" "{\"method\":\"GET\",\"path\":\"/venues\"}"
1582661946.408116 [0 127.0.0.1:54328] "PING"
1582661946.408220 [0 127.0.0.1:54328] "GET" "{\"method\":\"GET\",\"path\":\"/venues\"}"
1582661946.409396 [0 127.0.0.1:54328] "PING"
1582661946.409488 [0 127.0.0.1:54328] "SET" "{\"method\":\"GET\",\"path\":\"/venues\"}" "{\"data\":[{\"id\":\"35b444ab-b76d-4992-8c4f-2612ab3472e0\",\"region_id\":null,\"is_private\":false,\"name\":\"Test venue_1582660906\",\"address\":\"address\",\"city\":\"city\",\"state\":\"CA\",\"country\":\"the best country\",\"postal_code\":\"2222\",\"phone\":\"5555555555\",\"promo_image_url\":null,\"created_at\":\"2020-02-25T20:01:55.651962\",\"updated_at\":\"2020-02-25T20:01:55.697736\",\"google_place_id\":null,\"latitude\":null,\"longitude\":null,\"timezone\":\"America/Los_Angeles\",\"slug_id\":\"9330dc88-2b60-4bee-bdfc-935768f2ca3f\"}],\"paging\":{\"page\":0,\"limit\":100,\"sort\":\"\",\"dir\":\"Asc\",\"total\":1,\"tags\":{}}}"
1582661946.409622 [0 127.0.0.1:54328] "PEXPIRE" "{\"method\":\"GET\",\"path\":\"/venues\"}" "10000"
1582661947.106461 [0 127.0.0.1:54328] "PING"
1582661947.106558 [0 127.0.0.1:54328] "GET" "{\"method\":\"GET\",\"path\":\"/venues\"}"
1582661947.981397 [0 127.0.0.1:54328] "PING"
1582661947.981548 [0 127.0.0.1:54328] "GET" "{\"method\":\"GET\",\"path\":\"/venues\"}"
```

This is also visible by looking at the response times via the dev console.

Without caching (first request):
Screen Shot 2020-02-25 at 3 29 24 PM

Subsequent request:
Screen Shot 2020-02-25 at 3 29 32 PM

### Migrations
 * No Migrations

### Environment Variables
 * No change

### DIFF
--- a/api/src/middleware/cache_resource.rs
+++ b/api/src/middleware/cache_resource.rs
@@ -1,9 +1,13 @@
+use actix_web::http::header::{HeaderName, HeaderValue};
+use actix_web::http::{HttpTryFrom, Method};
 use actix_web::middleware::{Middleware, Response, Started};
-use actix_web::{http::Method, FromRequest, HttpRequest, HttpResponse, Result};
+use actix_web::{FromRequest, HttpRequest, HttpResponse, Result};
 use extractors::*;
 use helpers::*;
 use server::AppState;
 use std::collections::BTreeMap;
+
+const CACHED_RESPONSE_HEADER: &'static str = "X-Cached-Response";
 
 #[derive(Clone)]
 pub struct CacheResource {}
@@ -55,7 +59,7 @@ impl Middleware<AppState> for CacheResource {
         Ok(Started::Done)
     }
 
-    fn response(&self, request: &HttpRequest<AppState>, response: HttpResponse) -> Result<Response> {
+    fn response(&self, request: &HttpRequest<AppState>, mut response: HttpResponse) -> Result<Response> {
         if request.method() == Method::GET {
             let mut query = BTreeMap::new();
             let resource = request.resource().clone();
@@ -84,6 +88,11 @@ impl Middleware<AppState> for CacheResource {
                     .inner
                     .clone()
                     .and_then(|conn| caching::set_cached_value(conn, &config, &response, &query).ok());
+            } else if cached.is_some() {
+                response.headers_mut().insert(
+                    &HeaderName::try_from(CACHED_RESPONSE_HEADER).unwrap(),
+                    HeaderValue::from_static("1"),
+                );
             }
         }
 
--- a/api/src/middleware/cache_resource.rs
+++ b/api/src/middleware/cache_resource.rs
@@ -1,15 +1,17 @@
-use actix_web::http::header::{HeaderName, HeaderValue};
-use actix_web::http::{HttpTryFrom, Method};
+use actix_web::http::header::*;
+use actix_web::http::{HttpTryFrom, Method, StatusCode};
 use actix_web::middleware::{Middleware, Response, Started};
-use actix_web::{FromRequest, HttpRequest, HttpResponse, Result};
+use actix_web::{Body, FromRequest, HttpRequest, HttpResponse, Result};
+use bigneon_http::caching::*;
 use extractors::*;
 use helpers::*;
+use serde_json::Value;
 use server::AppState;
 use std::collections::BTreeMap;
+use uuid::Uuid;
 
 const CACHED_RESPONSE_HEADER: &'static str = "X-Cached-Response";
 
-#[derive(Clone)]
 pub struct CacheResource {}
 
 impl CacheResource {
@@ -18,8 +20,27 @@ impl CacheResource {
     }
 }
 
+pub struct CacheConfiguration {
+    cache_response: bool,
+    served_cache: bool,
+    error: bool,
+    user_id: Option<Uuid>,
+}
+
+impl CacheConfiguration {
+    pub fn new() -> CacheConfiguration {
+        CacheConfiguration {
+            cache_response: false,
+            served_cache: false,
+            error: false,
+            user_id: None,
+        }
+    }
+}
+
 impl Middleware<AppState> for CacheResource {
     fn start(&self, request: &HttpRequest<AppState>) -> Result<Started> {
+        let mut cache_configuration = CacheConfiguration::new();
         if request.method() == Method::GET {
             let mut query = BTreeMap::new();
             let resource = request.resource().clone();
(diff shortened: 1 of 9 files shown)


api/src/middleware/cache_resource.rs
```diff
@@ -0,0 +1,92 @@
...
+impl CacheResource {
+    pub fn new() -> CacheResource {
+        CacheResource {}
+    }
+}
+
+impl Middleware<AppState> for CacheResource {
+    fn start(&self, request: &HttpRequest<AppState>) -> Result<Started> {
+        if request.method() == Method::GET {
+            let mut query = BTreeMap::new();
```
Timo614: Using a BTreeMap here to ensure the keys are in alphabetical order when it's serialized for use as the redis key. Previously during testing I noticed the HashMap provided by `request.query()` was serializing in random order causing cache misses on identical requests so opted for this approach.

I've added the method and the path to the serialized BTreeMap so that key is unique across different endpoints. The method could be removed I suppose in light of the fact it's limited to just GET requests but probably not a big deal either way.

api/src/middleware/cache_resource.rs
```diff
@@ -0,0 +1,92 @@
...
+            let path_text = "path".to_string();
+            let path = resource_def.unwrap().pattern().to_string();
+            let method_text = "method".to_string();
+            let method = request.method().to_string();
+            query.insert(&path_text, &path);
+            query.insert(&method_text, &method);
+            let state = request.state().clone();
+            let config = state.config.clone();
+            let user = OptionalUser::from_request(request, &());
+            if user.is_ok() && user.unwrap().0.is_none() {
```
Timo614: We could optionally still cache logged in user info just include the user_id (or none) in the serialized query that's used as the key. For now I opted not to since the cache timeout is so low it's just added work with little benefit but if there's value we could go that direction.

Not sure of a good way to share the cache across logged in users as this happens prior to authorization. It's fine if we did add a per user cache as that user would have been authorized on a previous hit of the endpoint so we know the response is valid for them but if it's a shared cache across authorized users we'd need some way to introduce an authorization step in the mix (which varies for permissions by endpoint). Perhaps if we moved authorization and authorization into a middleware we could just have that run first but it would require us to then load up the organization in the middleware so could get a bit tricky. A problem for another day though.

Timo614:
Sounds good, added the header.

Normal response from API:
Screen Shot 2020-02-26 at 11 10 36 AM
Screen Shot 2020-02-26 at 11 10 44 AM

Cached response from API:
Screen Shot 2020-02-26 at 11 11 00 AM
Screen Shot 2020-02-26 at 11 11 05 AM

Timo614:
It looks like the tests are failing due to branch.io issues so will rerun them (they've had issues most of the day).

Timo614:
I've pushed up a commit to fix the guest_list test instability. It uses a subset of the ID for searching so it ends up occasionally returning multiple records. I have the logic check against the ids for all those records returned now so it shouldn't cause so many false positive test failures on drone.

api/src/routing.rs
```diff
@@ -19,16 +20,19 @@ pub fn routes(app: &mut CorsBuilder<AppState>) -> App<AppState> {
...
     .resource("/artists/{id}/toggle_privacy", |r| {
         r.method(Method::PUT).with(artists::toggle_privacy);
     })
     .resource("/artists/{id}", |r| {
+        r.middleware(CacheResource::new());
         r.method(Method::GET).with(artists::show);
         r.method(Method::PUT).with(artists::update);
     })
     .resource("/artists", |r| {
+        r.middleware(CacheResource::new());
```
mikethetike: I think we should explicitly set which query params to cache on, otherwise an attacker could pass in random params to fill up our cache

api/src/middleware/cache_resource.rs
mikethetike: I think we should cache for logged in and logged out. Users may log in and then refresh the page constantly waiting for tickets. The other thing to consider is that I think OptionalUser::from_request() does a database query so we will still be bound to DB CPU performance

api/src/middleware/cache_resource.rs
```diff
@@ -0,0 +1,101 @@
...
+                // if there is a error in the cache, the value does not exist
+                let cached_value = cache_database
+                    .clone()
+                    .inner
+                    .clone()
+                    .and_then(|conn| caching::get_cached_value(conn, &config, &query));
+                if let Some(response) = cached_value {
+                    // Insert self into extensions to let response know not to set the value
+                    request.extensions_mut().insert(self.clone());
+                    return Ok(Started::Response(response));
```
mikethetike: It might be worth setting some HTTP Headers + etag to let the client cache this as well

api/src/routing.rs
Timo614: I imagine we'll have a pretty short expiration period for the cache to avoid issues like events going on sale and not appearing the response. Something along the lines of a couple seconds or so. I think since the keys expire it shouldn't be a problem with the random parameters. 

If we're being hit with a DDOS attack we'd probably have bigger issues on our hands and the cache filling up a bit would be less of a concern.

I could add a whitelist but I think we'll end up running in cache collisions as we iterate on the endpoints in question adding new fields. Could be a hard thing to debug in those cases since it won't be clear what's going on at first. May be worth seeing how things work out in practice for a bit and confirm. We could always disable the cache or lower its expiration if there's an attack and then circle back with a whitelist if that was an issue during it.

api/src/middleware/cache_resource.rs
Timo614: Sounds good will add some.

Timo614:
Will add the per user caching and additional headers here. :+1:

Timo614:
I've added the cache control and etags headers and adjusted the logic to cache per user_id data. I adjusted the ENV vars to allow myself to test these locally in chrome.

`REDIS_CACHE_PERIOD_MILLI=30000`
`CLIENT_CACHE_PERIOD=10` 

Original request:
Screen Shot 2020-02-28 at 2 48 10 PM

Cache control request:
Screen Shot 2020-02-28 at 2 48 16 PM

Redis cache with user id:
Screen Shot 2020-02-28 at 2 48 31 PM

Redis cache logged out:
Screen Shot 2020-02-28 at 2 49 49 PM

ETag matches (used curl since chrome somewhat hides it with the response):
```
curl 'http://localhost:9000/events?query=&page=0&status=Published' -H 'Connection: keep-alive' -H 'Accept: application/json, text/plain, */*' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.130 Safari/537.36' -H 'X-API-Client-Version: bn-api-node@2.1.13' -H 'Origin: http://localhost:3000' -H 'Sec-Fetch-Site: same-site' -H 'Sec-Fetch-Mode: cors' -H 'Referer: http://localhost:3000/' -H 'If-None-Match: 37cd18c06479ce4eaeaa79af6913b4e139e27da8' -i
HTTP/1.1 304 Not Modified
x-cached-response: 1
cache-control: public, max-age=10
etag: 37cd18c06479ce4eaeaa79af6913b4e139e27da8
access-control-expose-headers: x-cached-response, x-app-version
vary: Origin
x-app-version: 1.9.18
content-length: 0
date: Fri, 28 Feb 2020 21:52:58 GMT
```

api/src/routing.rs
mikethetike: Fair enough

mikethetike:
Pushed a change that allows different routes to use different caching strategies for users. I'm not sure we will use all of them, but they are there for future use.

api/src/middleware/cache_resource.rs
```diff
@@ -56,25 +72,39 @@ impl Middleware<AppState> for CacheResource {
...
+                        CacheUsersBy::GlobalRoles => {
+                            cache_configuration.user_key = Some(user.user.role.iter().map(|r| r.to_string()).join(","));
+                        }
+                    }
+                    if let Some(ref user_key) = cache_configuration.user_key {
+                        query.insert(&user_text, user_key.to_string());
                     }
-                    query.insert(&user_text, &user_value);
-                }
-                Err(error) => {
```
Timo614: My thinking for the error here was that given we're swallowing the error here in the event it failed here but succeeded in the controller in all of the other requests we could end up caching user specific data as the non user cache. It's an extreme edge case though.

api/src/middleware/cache_resource.rs
```diff
@@ -56,25 +72,39 @@ impl Middleware<AppState> for CacheResource {
...
+                        CacheUsersBy::None => (),
+                        CacheUsersBy::AnonymousOnly => {
+                            // Do not cache
+                            return Ok(Started::Done);
+                        }
+                        CacheUsersBy::UserId => {
+                            cache_configuration.user_key = Some(user.id().to_string());
+                        }
+                        CacheUsersBy::GlobalRoles => {
+                            cache_configuration.user_key = Some(user.user.role.iter().map(|r| r.to_string()).join(","));
```
Timo614: I think this may only work if we've applied the user role to the user record. I don't believe we do that for anything besides User, Admin, and Super. The others are on a per organization level. So it would get the difference between a regular user and a super admin but anything in between is going to be treated as a regular user.

Timo614: Just head the comment above (was going bottom to top). Sounds good, this was taken into account. :+1:

mikethetike: I originally had an `CacheUsersBy::OrganizationAccess`, but later decided it was very specific and would probably be handled by the `CacheUsersBy::UserId` if needed

PR #1706, opened 2020-03-06
Fix concurrency issues when publishing domain events

taritom:
### References Issues:
https://app.asana.com/0/1158142652422330/1163262011434617/f

### Description:
The domain event publisher process would cause deadlocks on the domain_event_publishers table and attempt to insert duplicates into domain_event_published. These two issues should be resolved.

### Migrations
 * Database schema has been updated. The new column has a default value so no special steps need to be taken.

### Environment Variables
 * No change


db/src/models/domain_event_publishers.rs
```diff
@@ -335,14 +328,42 @@ impl DomainEventPublisher {
...
+                Ok(())
+            }
+            None => DatabaseError::concurrency_error("Another process is busy with this publisher"),
+        }
+    }
+
+    pub fn renew_lock(&mut self, timeout: i64, conn: &PgConnection) -> Result<(), DatabaseError> {
+        let timeout = Utc::now().naive_utc() + Duration::seconds(timeout);
+        let result: Option<DomainEventPublisher> = diesel::update(&*self)
+            .filter(domain_event_publishers::blocked_until.eq(self.blocked_until))
```
mikethetike: Watch out for some milliseconds getting trimmed here in the db. Not sure if it happens, but could be some minor differences in the Rust representation and the SQL

taritom: I'll look into and adjust appropriately if need be.

PR #1710, opened 2020-03-09
Upgrade bigneon_api to actix 2

dunnock:
### References Issues:

https://app.asana.com/0/1158142652422330/1166895853985361/f

### Description:

This PR is upgrading ./api crate to actix 2, which would allow to reduce async bolerplate in the future. This should not change logic or blocking calls, instead adapt all implementations to actix 2 API.

This is second step out of 4 steps plan to use actix in a non blocking way. The end goal is to make actix handle more concurrent requests as well as make slow requests to not block fast ones. Fast requests Include /events/{id} and /health so that should make whole system responsive during load spikes.

Testing notes:
 - caching logic
 - authorized/not authorized general flow
 - facebook login
 - logging middleware logic
 -- authorized uuid used to be logged as Option::Option::uuid became Option::uuid
 - database transaction handling logging (middleware)
 - domain events and communications
 - event's websocket communication
 - SSR logic

Notes:
 - domain events and communications primarily use sync clients under the hood
 - all db logic is sync, including obtaining connection from the pool
 - Connection is not Sync, to be able to use web::block it needs to be Sync
 - IMHO we can minimize number of open connections converting readonly connections to not use transactions and to use async pool like deadpool with a small number of connections.
 - FromRequest for Json<T> and RequestInfo can be replaced with standard extractors
 - Authorization can be replaced with Identity

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/extractors/uuid.rs
```diff
@@ -0,0 +1,41 @@
+use crate::auth::claims;
+use crate::jwt::{decode, Validation};
+use crate::server::GetAppState;
+use actix_web::error::{self, ErrorBadRequest, ErrorUnauthorized};
+use actix_web::HttpMessage;
+
+pub(crate) struct Uuid;
```
Timo614: Wondering if it's worth renaming this struct from Uuid to something like AuthorizationUuid, or the like. Nothing wrong with the encapsulation, etc. just thinking so it's clearer what the Uuid belongs to since we have other Uuids present in the request process / would make it easier outright to understand its role at a glance.

api/src/middleware/big_neon_logger.rs
```diff
@@ -1,103 +1,155 @@
...
     }
 }
 
-impl Middleware<AppState> for BigNeonLogger {
-    fn start(&self, req: &HttpRequest<AppState>) -> error::Result<Started> {
-        self.logger.start(req)?;
-        let user = OptionalUser::from_request(req, &());
-        let ip_address = req.connection_info().remote().map(|i| i.to_string());
+pub struct RequestLogData {
+    user: Option<uuid::Uuid>, // NOTE: this used to be Option<Option<uuid::Uuid>>
```
Timo614: May be worth removing the comment here in favor of the in the PR notes just since it won't matter later on once the context is gone. Seems fine though :+1:

api/src/extractors/uuid.rs
dunnock: Yeah, I thought too. Fixed

api/src/middleware/big_neon_logger.rs
dunnock: Done, also removed one more comment about web::block as that anyways should be separately organized effort

PR #1711, opened 2020-03-09
Adjust logic to clear cache

Timo614:
### References Issues:

### Description:
Adds logic to clear the cache based on a search parameter. The logic is pretty dumb right now as it's just searching by ID but it allows the cache to remove the old records immediately after an update. I guess we could expand it to clear the events cache entirely, etc. on updates as well but not the main focus here. With the changes here I see things update locally pretty fast after a change is made to the resource. Before I saw the cached response returned for some time after a change. We may eventually want to switch to admin only endpoints that aren't cached but it may be nice for performance if we can keep a cache in front.

### Example:

Cache key exists and is returned multiple times.
```
1583779412.669506 [0 127.0.0.1:58273] "GET" "{\"method\":\"GET\",\"path\":\"/venues/7f7f6f13-bab4-46ca-833b-bedc03461256\",\"x-user-role\":\"\"}"
1583779431.270859 [0 127.0.0.1:58273] "PING"
1583779431.270986 [0 127.0.0.1:58273] "GET" "{\"method\":\"GET\",\"path\":\"/events/af27d2bd-cf44-497e-8e5a-188927639db3\",\"x-user-role\":\"User,Admin,Super\"}"
1583779431.872593 [0 127.0.0.1:58273] "PING"
1583779431.872740 [0 127.0.0.1:58273] "GET" "{\"method\":\"GET\",\"path\":\"/venues/7f7f6f13-bab4-46ca-833b-bedc03461256\",\"x-user-role\":\"\"}"
1583779438.238531 [0 127.0.0.1:58273] "PING"
```

Logic does a keys search for anything using this UUID
```
1583779438.238708 [0 127.0.0.1:58273] "KEYS" "*/af27d2bd-cf44-497e-8e5a-188927639db3*"
1583779438.238877 [0 127.0.0.1:58273] "PING"
```

Logic deletes any returned keys:
```
1583779438.238977 [0 127.0.0.1:58273] "DEL" "{\"method\":\"GET\",\"path\":\"/events/af27d2bd-cf44-497e-8e5a-188927639db3\",\"x-user-role\":\"User,Admin,Super\"}"
``` 

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/controllers/artists.rs
```diff
@@ -175,6 +176,11 @@ pub fn update(
         action.commit(connection)?;
     }
 
+    cache_database
```
Timo614: Wondering if there's a better way to handle this without needing to push the cache database into each endpoint that needs to clear some records.

One thing I was thinking of was having middleware handle the deleting by having the cache key added to the http request extensions (and if present handling the deletion of those keys) but then there's the issue of getting it into the http request extension (so we'll need to include that in the method signature then regardless).

Timo614: Another idea is to take the PUT request with the middleware and just attempt to remove any matching resources. So if a PUT request is made for /venues/0000-0000-0000-0000 then a delete request is also made for /venues/0000-0000-0000-0000. 

I'll actually switch to that approach as it's much more extensible. We'll still need the adjusted method signature for publishing and unpublishing since those won't match but lets the RESTful setup work without any adjustments.

api/src/middleware/cache_resource.rs
```diff
@@ -197,6 +200,12 @@ impl Middleware<AppState> for CacheResource {
                     }
                 }
             }
+        } else {
+            // Method besides GET requested, PUT/POST/DELETE so clear any matching path cache
+            cache_database
```
Timo614: I'm going to make one more addition here to add a check to see if the response is an error. If it's an error the logic won't delete the cache. Thinking this way the logic prevents folks from spamming this and clearing the cache while they have auth errors, etc.

api/src/middleware/cache_resource.rs
```diff
@@ -101,6 +101,8 @@ impl Middleware<AppState> for CacheResource {
                         query.insert(&user_text, user_key.to_string());
                     }
                 }
+            } else {
+                query.insert(&user_text, "".to_string());
```
Timo614: I noticed the caching logic here was previously faulty for the non user type. It would cache the response with an empty string value but we were doing the lookup with no parameter at all.

api/src/middleware/cache_resource.rs
```diff
@@ -96,11 +107,54 @@ impl Middleware<AppState> for CacheResource {
...
                         }
+                        CacheUsersBy::OrganizationScopePresence(load_type, scope) => {
+                            if let Some(connection) = request.extensions().get::<Connection>() {
+                                let connection = connection.get();
+                                match load_type {
+                                    OrganizationLoad::Path => {
+                                        // Assumes path element exists
+                                        let organization_id: Uuid =
+                                            request.match_info().get(&"id".to_string()).unwrap().parse().unwrap();
+                                        let organization = match Organization::find(organization_id, connection) {
```
Timo614: This isn't ideal but I think it's benefit of caching this data outweighs any negatives. I was thinking about potentially moving the scopes into a middleware but it doesn't give a lot of freedom for situations where we have a few scopes depending upon functionality, etc.

This logic attempts to fetch the organization and then goes and sees if the passed in scope is present for that user. This was the only way I could think of the best means to determine in the start of the middleware so we could return a cached result.

Timo614:
Can see the newest behavior with the org fans endpoint here:
```
1583860982.603111 [0 127.0.0.1:51192] "GET" "{\"limit\":\"20\",\"method\":\"GET\",\"page\":\"0\",\"path\":\"/organizations/800da891-cbf2-4999-a932-61927e15676a/fans\",\"query\":\"\",\"sort\":\"LastInteracted\",\"x-user-role\":\"org:fans-t\"}"
1583860989.864575 [0 127.0.0.1:51192] "PING"
1583860989.864780 [0 127.0.0.1:51192] "GET" "{\"limit\":\"20\",\"method\":\"GET\",\"page\":\"0\",\"path\":\"/organizations/800da891-cbf2-4999-a932-61927e15676a/fans\",\"query\":\"\",\"sort\":\"LastInteracted\",\"x-user-role\":\"org:fans-t\"}"
1583860996.441840 [0 127.0.0.1:51192] "PING"
1583860996.441976 [0 127.0.0.1:51192] "GET" "{\"limit\":\"20\",\"method\":\"GET\",\"page\":\"0\",\"path\":\"/organizations/800da891-cbf2-4999-a932-61927e15676a/fans\",\"query\":\"\",\"sort\":\"LastInteracted\",\"x-user-role\":\"org:fans-t\"}"
1583861079.767479 [0 127.0.0.1:51192] "PING"
1583861079.767661 [0 127.0.0.1:51192] "GET" "{\"limit\":\"20\",\"method\":\"GET\",\"page\":\"0\",\"path\":\"/organizations/800da891-cbf2-4999-a932-61927e15676a/fans\",\"query\":\"\",\"sort\":\"LastInteracted\",\"x-user-role\":\"org:fans-f\"}"
1583861091.820599 [0 127.0.0.1:51192] "PING"
1583861091.820788 [0 127.0.0.1:51192] "GET" "{\"limit\":\"20\",\"method\":\"GET\",\"page\":\"0\",\"path\":\"/organizations/800da891-cbf2-4999-a932-61927e15676a/fans\",\"query\":\"\",\"sort\":\"LastInteracted\",\"x-user-role\":\"org:fans-f\"}"
1583861094.471224 [0 127.0.0.1:51192] "PING"
```

Can see it returning the role string there t or f depending upon if the user had permissions. The false ones are not stored in the caching as they returned a 401 response.

I've held off from other endpoints behind the admin section as I think we should be a bit cautious as we roll these out initially in case there are some unknown bugs / once we're more comfortable can slowly roll out caching to other endpoints (we'll want to make sure the scope is enough if we reuse this caching strategy as some roles may see different responses to the same endpoints despite having the same scope access).

PR #1713, opened 2020-03-11
Auto expiring holds that release inventory

Timo614:
### References Issues:
https://app.asana.com/0/1158142652422330/1165647302888857

### Description:
The web side of holds appears to indicate that holds automatically release their inventory. This has never been true -- holds have never had any sort of logic in place that releases that as soon as the date passes. This pull request attempts to rectify this discrepancy by having the API run a domain action set to the end_at date for a given hold if one is set. If one is not set then it doesn't do anything in regards to that. 

Note:
- The domain action is updated on creation and update
- The logic does not affect child holds as they have their own end_at date so wouldn't make sense to be recursive

### Examples

Two holds, a parent and child with sales on the parent set to end 30 minutes apart:
Screen Shot 2020-03-11 at 1 39 15 PM
Screen Shot 2020-03-11 at 1 39 23 PM
Screen Shot 2020-03-11 at 1 39 36 PM

2:00 PM hits, inventory is returned for the parent but not the child:
```
{"level":"INFO","time":"[2020-03-11][18:00:53]","target":"server","message":"Domain Event: HoldQuantityChanged `Hold quantity decreased from 90 to 2` on Holds:2fec9925-2d0a-4223-86c5-4995937d1b69","domain_event_id":"6b78bb1b-d1bc-4bb7-8401-6afbdd1f9bf7","event_data":{"new_quantity":2,"old_quantity":90},"event_type":"HoldQuantityChanged","main_id":"2fec9925-2d0a-4223-86c5-4995937d1b69","main_table":"Holds"}
{"level":"INFO","time":"[2020-03-11][18:00:53]","target":"server","message":"Domain Event: HoldAutomaticallyReleased `Hold Test hold released` on Holds:2fec9925-2d0a-4223-86c5-4995937d1b69","domain_event_id":"f1a82458-2fae-44ff-b0ca-85fa3add6269","event_data":{"created_at":"2020-03-11T17:12:38.438444","deleted_at":null,"discount_in_cents":1100,"email":null,"end_at":"2020-03-11T18:00:00","event_id":"4b0c37ea-1e1d-4ba7-948c-35866ec797a3","hold_type":"Discount","id":"2fec9925-2d0a-4223-86c5-4995937d1b69","max_per_user":0,"name":"Test hold","parent_hold_id":null,"phone":null,"redemption_code":"ABC12349DJD","ticket_type_id":"90f5116a-91ca-4cc1-8f54-7a8353c5d9f1","updated_at":"2020-03-11T17:38:52.089161"},"event_type":"HoldAutomaticallyReleased","main_id":"2fec9925-2d0a-4223-86c5-4995937d1b69","main_table":"Holds"}
```

Screen Shot 2020-03-11 at 1 41 25 PM

2:30 PM hits, inventory is returned for the child as well:
```
{"level":"INFO","time":"[2020-03-11][18:30:36]","target":"server","message":"Domain Event: HoldQuantityChanged `Hold quantity decreased from 2 to 0` on Holds:250bd90e-4756-4fbc-89f5-b40123974ebf","domain_event_id":"ea6d365f-3be8-4f45-898b-a31b580f0375","event_data":{"new_quantity":0,"old_quantity":2},"event_type":"HoldQuantityChanged","main_id":"250bd90e-4756-4fbc-89f5-b40123974ebf","main_table":"Holds"}
{"level":"INFO","time":"[2020-03-11][18:30:36]","target":"server","message":"Domain Event: HoldAutomaticallyReleased `Hold Test hold released` on Holds:250bd90e-4756-4fbc-89f5-b40123974ebf","domain_event_id":"03fb1830-2e0a-47d7-a5ec-ec7f5f0d30d5","event_data":{"created_at":"2020-03-11T17:35:44.433788","deleted_at":null,"discount_in_cents":1100,"email":null,"end_at":"2020-03-11T18:30:00","event_id":"4b0c37ea-1e1d-4ba7-948c-35866ec797a3","hold_type":"Discount","id":"250bd90e-4756-4fbc-89f5-b40123974ebf","max_per_user":0,"name":"Test hold","parent_hold_id":null,"phone":null,"redemption_code":"ABC12349DJDE","ticket_type_id":"90f5116a-91ca-4cc1-8f54-7a8353c5d9f1","updated_at":"2020-03-11T17:35:44.433788"},"event_type":"HoldAutomaticallyReleased","main_id":"250bd90e-4756-4fbc-89f5-b40123974ebf","main_table":"Holds"}
```

Screen Shot 2020-03-11 at 1 42 05 PM

## Release Details:
This command should be run which will cause all old expired holds to immediately expire their inventory and setup other holds that have end dates to be ready to expire when they come to pass.

```
cargo run --bin api-cli schedule-missing-domain-actions -h
```

Note the `-h` flag. This is not normally run with the scheduling of missing domain actions as it shouldn't be needed more than once / we may want to retire that portion of the command as soon as it's done in production.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
Purely as a thought experiment, as I haven't checked the process completely in bn-api, but could we not expire the holds in the same way that we expire reserved tickets. So when we do a reserve ticket lookup, we check for any ticket_instance that has a hold_id where the expiry < now? Kind of like an implicit expiry instead of explicitly forcing the expiration?

Just thinking out loud here, I may be way off base.

Timo614:
@Krakaw I think that's a good idea but it has a few issues which make it less desirable. Namely we'd have to add somewhat SQL wherever we do counts (every report, endpoints that display quantity, etc) to properly show the correct current counts. We'd also have to modify checkout to handle allowing these to be purchased which wouldn't be that bad. It does somewhat add some complication across the board as well with SQL queries requiring the holds table as a join going forward to check if it's a hold ticket or not, etc.

I think the domain actions work pretty well though -- they are only made if the hold expires, there is only one per hold, it has a scheduled date of the end_at so it's immediate in regards to the release, and it requires the least amount of major changes in the system to support since they just become normal tickets again. 

I could be convinced otherwise though if we think it's a better idea to go that route with the hold tickets. I do think the reserved approach for the normal tickets makes the most sense given there are just so many of them, any delay in on sale even seconds may cause problems during a spike of interest (we want these back on sale immediately), and so on. but those concerns are less pronounced here given even a 30 second delay in releasing the hold inventory wouldn't be a big deal (and it should never be that long) and the data model ends up being a lot cleaner since we can rely on the existence of the join to signify an existing relationship / less junk data we have to work around.

Timo614:
I'll push up a PR today to switch to that other approach and we can make the call there. It won't be as straightforward looking over the data model but I don't think it'd be too difficult to get working.

Krakaw:
Nah I think your current approach is perfect, it was just a thought.

Timo614:
I'm rebasing to fix conflicts -- I've also adjusted the CLI a bit to match the new format since it had rebase conflicts and changed the optional sync holds flag to just `-h`

PR #1718, opened 2020-03-16
Toggleable refund access by organization

Timo614:
### References Issues:
https://app.asana.com/0/1158142652422330/1166418437480933

### Description:

This pull request includes logic that fixes the associated asana issue by allowing organizations to have their ability to initiate refunds be toggled by admins and supers.

Adjustments made:
- Create route for organization allows is_allowed_to_refund boolean which when not provided defaults to false
- Update route for organization allows is_allowed_to_refund boolean which when not provided does not update
- Logic that checks applicability to refund an order now checks if the user is an admin or if the organization has is_allowed_to_refund enabled or returns an unauthorized error
- Logic that toggles frontend refund functionality via order details updated to modify refundable flag depending upon if the user is an admin or if the organization is_allowed_to_refund

Note:
- I've defaulted these all to false -- I noticed that we wanted to remove the functionality from across the board via some other discussion so this starts us at that point. We'll need to enable refunds in any environments we want them such as scratches.

### Examples:

As a system administrator I can refund one ticket despite the organization having is_allowed_to_refund set to false:

Screen Shot 2020-03-16 at 4 22 48 PM

As the organization owner, I can't click the checkbox next to the other ticket as it is disabled by the is_allowed_to_refund flag:

Screen Shot 2020-03-16 at 4 24 15 PM

I've temporarily switched the flag in the db to true, this allows me to select the checkbox:

Screen Shot 2020-03-16 at 4 26 47 PM

Now I've set it back to false in the db it fails when it attempts to submit because I lack the permissions (normally you'd be blocked by the disabled checkbox above):

Screen Shot 2020-03-16 at 4 26 59 PM

Switching the organization to is_allowed_to_refund = true allows the refund to succeed:

Screen Shot 2020-03-16 at 4 27 08 PM

### Migrations
 * `migrations/20200316174006_add_is_allowed_to_refund_to_organizations`

### Environment Variables
 * No change

### DIFF
--- a/api/src/controllers/events.rs
+++ b/api/src/controllers/events.rs
@@ -1084,7 +1084,8 @@ pub fn guest_list(
     //TODO refactor GuestListQueryParameters to PagingParameters
     let conn = connection.get();
     let event = Event::find(path.id, conn)?;
-    user.requires_scope_for_organization_event(Scopes::EventViewGuests, &event.organization(conn)?, &event, conn)?;
+    let organization = event.organization(conn)?;
+    user.requires_scope_for_organization_event(Scopes::EventViewGuests, &organization, &event, conn)?;
 
     let query_string = query.clone().query;
     let changes_since = query.clone().changes_since;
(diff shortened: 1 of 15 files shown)


Krakaw:
Hey @Timo614 it's totally my fault I didn't assign the card to myself, I thought to expand on the initial spec slightly and pushed this up https://github.com/big-neon/bn-api/pull/1719 it allows for a per user per org revocation / addition or permissions, so that if we need a doorperson to do an emergency refund we can hack'n'slash their perms on the fly. Let me know what you think.

Timo614:
Sounds good

PR #1719, opened 2020-03-17
Add per org user permissions to revoke per org refunds

Krakaw:
### References Issues:
https://app.asana.com/0/1158142652422330/1166418437480933
### Description:
This PR allows you to add additional scopes to an org user or revoke specific scopes. It adds a new field to the `organization_users` table `additional_scopes` that can contain a `json_serde::Value` of 

```json
{ "additional": [], "revoked": ["order:refund"]}
```

```bash
api-cli organization_uuid [-u|--user user_id] [-a|--additional csv,scopes] [-r|--revoke csv,scopes] [-c|--clear]
```
### Migrations
 * No Migrations

### Environment Variables
 * No change


db/migrations/20200316115923_add_additional_scopes_to_organization_users/up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE organization_users
+    -- the additional_scopes is a json object of {"additional": ["ScopeEnum"], "revoked": ["ScopeEnum"]}
+    ADD COLUMN additional_scopes JSONB NULL;
```
mikethetike: Maybe add a default `{"additional":[], "revoked":[]"}`

Krakaw: My thinking of leaving it null might have been a micro/premature optimization, in that if there is a null value we can just ignore.

Krakaw: Though thinking about it, it probably would've simplified the code a fair amount.

PR #1720, opened 2020-03-17
Further caching improvements

Timo614:
### References Issues:

### Description:
We've been working on caching improvements. It was mentioned today that the box office was showing old available ticket data. It looks like they use the same event show as normal users with some additional information.

This pull request addresses our needs here by doing the following:
- Adds a new `CacheBypass` header that when set bypasses the cache
- Adjusted the logic to clear restful paths on modifications to no longer use a wildcard (it was searching for `*/events*` which would clear all of the caches under events) -- the same exact path is sufficient for now since it handles the main update / show case and index / post (although we may need to circle back for better cache clearing in the future but this is helpful)
- As per Keith given the events show endpoint is pretty intense in terms of the logic it's showing (different scope checks, etc) for now we're going to use a new cache strategy PublicUsersOnly which only caches users who either are logged out or who belong to no organizations at all
- Fixes issue with log spam when cache key not found / get now returns an Option where the none is the lack of a key
- Fixes issue from actix 2 modifications around query string and the cache key / the logic was updated to use the HttpRequest uri method and parse it with the Url class. The HttpRequest uri doesn't include a host so it ended up skipping over that logic and caches everything at a given path in the same key.

I was able to confirm this worked by logging in as an admin and org admin and confirming my redis monitor was not seeing cache hits and then trying again as a normal user and as a logged out and confirming it did see cache hits.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Timo614:
I've rebased this over the latest changes which modified the cache middleware a bit for actix 2.

Note: I found a new bug in develop from the latest changes where the logic is not respecting any query differences so it returns the same cache for any use of the path. The issue was the URI being returned by actix-web's HttpRequest was not a full URL so when it did the URL parsing it would fail and never add any of those query elements to the key.

Timo614:
@Krakaw any clue if the app uses the availability information at all or has in an previous iteration?

Planning on removing it from the event show response since we'll be updating the bn-web as well but want to make sure I'm not about to break some other consumers.

Timo614:
@Krakaw Actually I'll handle splitting out the event show into two endpoints (so we can have availability be uncached) in a separate PR and just set this one up so we can merge it when possible.

api/src/middleware/cache_resource.rs
```diff
@@ -13,9 +13,11 @@ use futures::future::{ok, Ready};
 use itertools::Itertools;
 use serde_json::Value;
 use std::collections::BTreeMap;
+use url::form_urlencoded;
 use uuid::Uuid;
 
 const CACHED_RESPONSE_HEADER: &'static str = "X-Cached-Response";
+const CACHE_BYPASS_HEADER: &'static str = "CacheBypass";
```
Krakaw: Total nitpick
```suggestion
const CACHE_BYPASS_HEADER: &'static str = "Cache-Bypass";
```

api/src/middleware/cache_resource.rs
```diff
@@ -116,6 +129,23 @@ impl CacheResource {
                         CacheUsersBy::UserId => {
                             cache_configuration.user_key = Some(user.id().to_string());
                         }
+                        CacheUsersBy::PublicUsersOnly => {
+                            if let Some(connection) = request.extensions().get::<Connection>() {
+                                let connection = connection.get();
+                                let is_public_user = match user.user.is_public_user(connection) {
```
Krakaw: I'm thinking should we not add this field to the user object and then fetch it as a subquery in the Users::find function, that way instead of 2 user queries on every request, we just have the one?

Timo614:
Sure sounds good, will quickly make those changes this morning.

api/src/middleware/cache_resource.rs
Timo614: Trying to sort out the best way to handle this.

Thinking that instead of going the approach mentioned I'll switch the way the AuthUser functions to include it and have the auth extractor logic use the sql that returns the subquery so it's still one request but using it in place of the find in that scenario. Originally was going to go the direction mentioned above re: having User::find but given it would mess with QueryableByName across the repo I was struggling to find a good way to approach that.

PR #1726, opened 2020-03-18
[High Risk] Split event show data into two endpoints to allow caching of event resource

Timo614:
### References Issues:

### Description:
We intended to cache the event show endpoint but began running into issues while testing in develop. The problem is we use the event show data to display whether or not there are tickets left for sale. We also include individual user level data in the form of limited tickets remaining info and if the user was interested which I've moved off of the event show endpoint response. 

Not sure if anyone has a better name or a different concept around how they pictured the split here. I was hesitant to include the interest info in the new endpoint but it seemed really wasteful to have to pull down another endpoint just for that.

Note: 
- I haven't modified how we display availability here just am splitting the endpoint. I figure we can circle back later and handle that properly once we've had a chance to get a UX review on how we want to display etc. and any copy. My gut would tell me the frontend should get fuzzy numbers up until we're at like 20 or less left and then we give exact to aid in the sales process.
- I noticed we also listed whether the user was interested in an event on the event index -- I've removed this -- will this be a problem? It doesn't appear to be used by the frontend `user_is_interested`. The frontend does use the event show version which has been replaced with the availability data.

Putting high risk as this will be a breaking change. I've asked in the asana issue if Keith is aware of the app ever using this data from event show so will adjust as needed. Worst case we could keep it in the event show just know any consumer will be getting delayed data while we transition but for now will keep it a breaking change.

### Examples:

Event Show API with fields removed: https://gist.github.com/Timo614/3adceccafbcc11647cc43843cc57e18e
Availability info: https://gist.github.com/Timo614/caa9cda450d021a799713d003350511e

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
Currently we use the `user_is_interested` field in the app, but only with event show. So pulling it out of index is fine. Let's just use the limit_per_person as the available unless the actual available is less than that?

Krakaw:
@coka we'll need to cover this change in the app before it goes live, also for backwards compatibility I need to dig in a little more before this goes live. This will solve a lot of problems though.

Krakaw:
Something I noticed while going through these functions is this query within a loop http://github.com/big-neon/bn-api/blob/3c77d2f0549dafcc73e591a5c0f0aa776aa1ac30/api/src/controllers/events.rs#L452-L452 all the ticket_type data should already be pulled, no need to re-pull it.

api/src/controllers/events.rs
```diff
@@ -308,42 +316,27 @@ pub struct EventParameters {
...
-        && !(query.private_access_code.is_some()
-            && event.private_access_code.clone().unwrap() == query.private_access_code.clone().unwrap().to_lowercase())
+        && !(private_access_code.is_some()
+            && event.private_access_code.clone().unwrap() == private_access_code.clone().unwrap().to_lowercase())
     {
         match user {
             Some(ref user) => user.requires_scope_for_organization(Scopes::OrgReadEvents, &organization, connection)?,
             None => {
-                return application::unauthorized_with_message("Unauthorized access of private event", None, None);
+                return Err(application::unauthorized_with_message::<HttpResponse>(
```
dunnock: BTW, it seems here you don't get value of using this helper. Error will be logged anyways with request data:
```
Err(AuthError::unauthorized("Unauthorized access of private event"))
```

Timo614:
Yeah we could do that -- they may end up with bad info but I guess some info is better than breaking things. Only downside is we may not notice some frontend logic still relying on the frontend which may cause issues with the cached value at random times and prove to be hard to reproduce. We just have to be careful and aware we may run into that and I think we'll be fine though. I'll quickly add things back today and switch the helper as Dunnock mentioned above.

Timo614:
One note of caution -- the cached results for if the user is interested will be based on the data fetched for whichever user was requesting it at the time of the caching. It's not really personal info but may confuse folks if any widgets use that interested info and are relying on the cached data we're adding back.

Krakaw:
> One note of caution -- the cached results for if the user is interested will be based on the data fetched for whichever user was requesting it at the time of the caching. It's not really personal info but may confuse folks if any widgets use that interested info and are relying on the cached data we're adding back.

Let's just set the is_interested to false for this endpoint, we won't be using it in the app / on bn-web, and the widget doesn't utilize it, so it should be low risk.

api/src/controllers/events.rs
```diff
@@ -355,50 +343,49 @@ pub async fn show(
...
-
-    let box_office_pricing = query.box_office_pricing.unwrap_or(false);
     if box_office_pricing {
         match user {
             Some(ref user) => {
                 user.requires_scope_for_organization(Scopes::BoxOfficeTicketRead, &organization, connection)?
             }
             None => {
-                return application::unauthorized_with_message("Cannot access box office pricing", None, None);
+                return Err(application::unauthorized_with_message::<HttpResponse>(
```
dunnock: this one is actually same as above:
```
return Err(AuthError::unauthorized("Cannot access box office pricing").into());
```

PR #1731, opened 2020-03-20
Added collections to support collectibles functionality.

taritom:
### Description:
Adds collections to support collectibles product.

### Migrations
New DB tables, no special actions needed.
db/migrations/20200310015532_collections

### Environment Variables
Added config to alter routes based on product.
example: PRODUCT_CONTEXT=BigNeon


api/src/controllers/collection_items.rs
```diff
@@ -0,0 +1,78 @@
...
+        Json<CreateCollectionItemRequest>,
+        User,
+    ),
+) -> Result<HttpResponse, ApiError> {
+    let conn = connection.get();
+    // todo: verify user owns at least one collectible of this type
+    // todo: decide what a reasonable limit for number of collectibles in a collection is
+    let collection = Collection::find(path.id, conn)?;
+    if collection.user_id != user.id() {
+        application::forbidden::<HttpResponse>("User does not have access to this collection")?;
```
dunnock: It's clear exit point, hence better to state it explicitly (minus turbo-fish):
```
return application::forbidden("User does not have access to this collection");
```

api/src/controllers/collection_items.rs
```diff
@@ -0,0 +1,78 @@
...
+    Ok(HttpResponse::Created().json(&created_item))
+}
+
+pub async fn index(
+    (connection, path, user): (Connection, Path<PathParameters>, User),
+) -> Result<HttpResponse, ApiError> {
+    let conn = connection.get();
+    let collection = Collection::find(path.id, conn)?;
+    if collection.user_id != user.id() {
+        application::forbidden::<HttpResponse>("User does not have access to this collection")?;
```
dunnock: same as above

api/src/controllers/collection_items.rs
```diff
@@ -0,0 +1,78 @@
...
+pub async fn update(
+    (connection, path, collection_item_update_attr, user): (
+        Connection,
+        Path<PathParameters>,
+        Json<UpdateCollectionItemAttributes>,
+        User,
+    ),
+) -> Result<HttpResponse, ApiError> {
+    let conn = connection.get();
+    let item = CollectionItem::find(path.id, conn)?;
```
dunnock: What about joining on collection to get user_id right from the item?

api/src/controllers/collection_items.rs
```diff
@@ -0,0 +1,78 @@
...
+    let collection = Collection::find(item.collection_id, conn)?;
+    if collection.user_id != user.id() {
+        application::forbidden::<HttpResponse>("User does not have access to this collection")?;
```
dunnock: same
```
return application::forbidden("User does not have access to this collection");
```

api/src/controllers/collection_items.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+pub async fn delete(
+    (connection, path, user): (Connection, Path<PathParameters>, User),
+) -> Result<HttpResponse, ApiError> {
+    let conn = connection.get();
+    let item = CollectionItem::find(path.id, conn)?;
+    let collection = Collection::find(item.collection_id, conn)?;
+    if collection.user_id != user.id() {
+        application::forbidden::<HttpResponse>("User does not have access to this collection")?;
```
dunnock:
```
return application::forbidden("User does not have access to this collection");
```

api/src/controllers/collection_items.rs
taritom: Agree. Surprised my lint didn't tell me of the unneeded type hint.

.drone.yml
```diff
@@ -50,6 +50,7 @@ env: &env
     CONNECTION_POOL_MIN: 1
     BRANCH_IO_BRANCH_KEY:
         from_secret: BRANCH_IO_BRANCH_KEY
+    PRODUCT_CONTEXT: BigNeon
```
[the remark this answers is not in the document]
taritom: Would like to keep it separate. In the case we have a conflict of functionality with the same route this is makes it easy to deal with it. Can also shim with different request/response structs (verbiage change such as collectible instead of ticket type) while executing the same logic underneath. This also makes it easy to identify what functionality is explicitly related to a vertical on the code base like collectibles or another in the future.

api/src/config.rs
```diff
@@ -452,6 +461,16 @@ impl Config {
...
         let static_file_path = env::var(&STATIC_FILE_PATH).map(|s| Some(s)).unwrap_or(None);
 
+        let product_context = match env::var(&PRODUCT_CONTEXT)
+            .unwrap_or("bigneon".to_owned())
+            .to_lowercase()
+            .as_str()
+        {
+            "collectibles" => ProductContext::Collectibles,
+            "bigneon" => ProductContext::BigNeon,
+            _ => ProductContext::BigNeon,
```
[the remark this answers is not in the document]
taritom: Considered doing that but opted for the less loud option. I will change it to a panic since people are cool with it.

api/src/controllers/collection_items.rs
taritom: I could. If we need maximum performance I'd create INSERT/UPDATE/DELETE statements utilizing JOINs and SELECTs and read the result to decide if it erred.

dunnock: Those DB routes are blocking, so I would minimize time spent in a blocking routes as much as possible. If that's reasonably easy optimization I would opt doing that.

.drone.yml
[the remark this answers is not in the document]
dunnock: This is the only point left I am not sure, maybe @mikethetike might advice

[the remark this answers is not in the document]
mikethetike: I think the ENV is fine for now. Let's give it a try. We can always change it later

api/src/controllers/collection_items.rs
taritom: Deferring for now. Can be a concerted effort in the future to minimize all round trips to postgres throughout the db layer.

dunnock:
@taritom it has some conflicts in .drone.yml

taritom:
> @taritom it has some conflicts in .drone.yml

PR #1732, opened 2020-03-20
Add email only registration via Magic Link 

mikethetike:
### Description:
Adds the ability to allow users to register via email only. They will be sent an email with a link that has a refresh token which they can then log in with. 

This can be turned off via config, in which case the endpoint will return HTTP 405

### Migrations
 * No Migrations

### Environment Variables
 * EMAIL_TEMPLATES_USER_REGISTERED_MAGIC_LINK
 * EMAIL_ONLY_REGISTRATION_ALLOWED


api/src/communications/mailers/user.rs
```diff
@@ -89,3 +92,42 @@ pub fn invite_user_email(config: &Config, user: &User, conn: &PgConnection) -> R
...
+    custom_data.insert("refresh_token".to_string(), json!(refresh_token));
+    let link = deep_linker.create_with_custom_data(&desktop_url, custom_data)?;
+
+    let source = CommAddress::from(config.communication_default_source_email.clone());
+    let destinations = CommAddress::from(email.to_string());
+    let title = format!("Welcome to {}", SITE_NAME);
+    let template_id = config.email_templates.user_registered_magic_link.to_string();
+    let mut template_data = TemplateData::new();
+    template_data.insert("download_link".to_string(), link);
+    template_data.insert("refresh_token".to_string(), refresh_token);
```
Timo614: Haven't reviewed just saw the text about a refresh token in email -- maybe we expire it somehow if it's not used before a certain date? Like perhaps we encrypt the refresh token and an expiration date and then when they go to use it give them the new refresh token that's decrypted if it hasn't expired? Not sure if there are plans for this already or if there's a better way to handle this. Just worried about giving these really long lasting tokens out via email in plaintext. Do the magic links ever expire? Or intended to last forever?

mikethetike: These refresh tokens expire after 2 hours:
```
 let refresh_token = match user.create_magic_link_token(
        state.service_locator.token_issuer(),
        Duration::minutes(120),
        false,
        conn,
    )?
```

api/src/controllers/users.rs
```diff
@@ -299,6 +302,49 @@ pub async fn register(
...
+            ErrorCode::DuplicateKeyError => {
+                return application::unprocessable("A user with this email already exists");
+            }
+            _ => return Err(e.into()),
+        },
+    };
+
+    let refresh_token = match user.create_magic_link_token(
+        state.service_locator.token_issuer(),
+        Duration::minutes(120),
```
taritom: Probably move to config either file or possibly DB to be adjusted by a super admin.

api/src/communications/mailers/user.rs
taritom: Also these cannot be generated for admin type roles. If that is a future requirement would certainly want a consumable instead of a token.

dunnock: @Timo614 @taritom as it expires are you ok with this solution? 
We can make timeout configurable later any time when needed.

taritom: Yes.

Timo614: Sounds good

PR #1745, opened 2020-03-31
Surface total uses of codes for dashboard

Timo614:
### References Issues:
https://app.asana.com/0/1158142652422330/1163466546088745
#1741 

### Description:
I adjusted some logic to make available null for codes in the event it was an unlimited use event. Previously we were returning a negative number there for the unlimited as the max per user limit was set to 0 which indicating unlimited uses. Didn't realize the frontend was using this value to determine the total uses of the codes and it was relying on the bad data in the available number unintentionally in the unlimited case. This was caused because the frontend did the following equation: `max uses - available = total uses` which turns into `0 - (-3) = 3`. With my fix it became effectively `0 - 0 = 0` so we'd show no uses of the code.

The fix here is just to surface the total uses in the codes and holds index responses. They can continue to calculate the hold if they want by subtracting out the values (holds have inventory pulled aside so it's not an unlimited thing ever and will always be right) or can update to the new calculated field but did that for consistency. The frontend will need to update to use the new field as part of these changes.

### Migrations
 * No Migrations

### Environment Variables
 * No change


dunnock:
hi @Timo614 , so it seems by the change information will only be available via `/events/{id}/holds` which requires `Scopes::HoldRead`. So UI won't be able to use it on the event page for user. Perhaps add to `/redemption_codes/{code}` too?

Timo614:
@dunnock this is only used for the dashboard -- we shouldn't surface it for the user likely since it wouldn't be of much use there and it may give some insight into how well the code is working if they pinged it often and compared the total used number over time.

Timo614:
It's also available to the codes via the CodesWithAvailability or whatever it is struct which uses serde attributes to flatten the struct into the response with the code data. The events#codes controller uses that model so ends up getting included.

dunnock:
Sorry, it seems I confused this subject, I was thinking on ticket_type.limit_per_person...

PR #1748, opened 2020-04-01
Add forced rollback transaction on connection drop

dunnock:
### Description:

Use official versions of diesel and r2d2 connection pool. 

Currently used diesel fork https://github.com/mikethetike/diesel/tree/r2d2-transaction-check is resolving 2 problems:
1. PooledConnection might keep open transaction when returned to the pool, which might lead to issues next time it retrieved from the pool
2. It is based on master, which have sql_query.into_boxed(). This is not available yet in diesel:1.4.4

Resolution for p1. PR exploits the fact that PooledConnection is returned to the pool via Drop which is standard practice. As ConnectionType is a wrapper for PooledConnection, custom Drop impl on ConnectionType will check if there are any pending transactions and if so it will try rollback those. Worst case issuing `DISCONNECT`.

Solution for p2. Copied into_boxed() implementation from diesel:master to db/utils/boxed_query.rs. As soon as diesel:1.5 released this code will become redundant.

### Migrations
 * No Migrations

### Environment Variables
 * No change


api/src/database/connection_type.rs
```diff
@@ -7,3 +9,32 @@ pub enum ConnectionType {
     Pg(PgConnection),
     R2D2(R2D2PooledConnection),
 }
+
+// When connection dropped check if there is currently hanging transactions
+impl Drop for ConnectionType {
+    fn drop(&mut self) {
+        if let Self::R2D2(ref mut conn) = self {
+            // This is guarnteed checkpoint before returnning PooledConnection to the Pool
```
Krakaw, line 17:
```suggestion
            // This is guaranteed checkpoint before returning PooledConnection to the Pool
```

PR #1751, opened 2020-04-02
Removed the lock from dockerfile

Krakaw:
### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
> We'll need to keep an eye on any possible build issues

Yeah it failed in the CD pipeline without the lock file. I also think we can optimize the Dockerfile because I don't think the pre-caching works anyway.

PR #1754, opened 2020-04-03
Small optimization on diesel build

dunnock:
### Description:

Turned off unused features from diesel. This is only small optimization. That decreased dependencies from 370 down to 355.

I have tried to get rid of `64-column-tables` feature, but gave up after spending few hours trying to convert events table.

### Migrations
 * No Migrations

### Environment Variables
 * No change


Krakaw:
What about adding a new field to events called: `additional_json` and adding the following fields and using a `From` and `Into` to bundle those fields in and extract them out on save and load?
```
cover_image_url
video_url
top_line_info
additional_info
promo_image_url
```

PR #1755, opened 2020-04-03
Reduce diesel build time 

dunnock:
### Description:

Diesel and db take more than 70% of build time and in diesel docs it is said that when using `64-column-tables` build times will significantly increase. We are using `64-column-tables` only to support event which has 35 columns, rest of objects are less than 32 columns.

In this PR event model is amended, following fields moved to jsonb field `additional_json`:
```
    cover_image_url,
    video_url,
    top_line_info,
    additional_info,
    promo_image_url,
```

The challenge though is to preserve back compatibility on API level. `Event`, `NewEvent`, `EventEditableAttributes` models all exposed in the API endpoints. This PR separates DB model and externally available data, introducing 4 new crate-local structs for communication with Diesel, while all existing structs and methods stay unchanged. DomainEvents jsons in event_data field preserve back compatibility too.

Build time on my PC with clean build reduced from `7m 37s` down to `5m 44s`

### Migrations
 * No Migrations

### Environment Variables
 * No change


db/migrations/20200403133417_events_additional_info/up.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE events
+    -- the additional_json is a json object of {
+	--   cover_image_url,
+    --   video_url,
+    --   top_line_info,
+    --   additional_info,
+    --   promo_image_url,
+	-- }
+    ADD COLUMN additional_json JSONB not null default '{}'::jsonb;
```
Krakaw, line 9:
```suggestion
    ADD COLUMN additional_json JSONB NOT NULL default '{}'::jsonb;
```

db/src/models/events/mod.rs
```diff
@@ -766,16 +766,19 @@ impl Event {
...
+        let update = EventEditableAttributesData::from(event);
+        let data: EventData = DatabaseError::wrap(
             ErrorCode::UpdateError,
             "Could not update event",
-            diesel::update(self)
-                .set((&event, events::updated_at.eq(dsl::now)))
+            diesel::update(&EventId::from(self))
+                .set((&update, events::updated_at.eq(dsl::now)))
                 .get_result(conn),
         )?;
```
Krakaw, line 777: Should we replace this `DatabaseError::wrap` with a `.to_db_error`

dunnock: yeah replaced for events, there might be more in other modules though

db/src/models/events/model.rs
```diff
@@ -0,0 +1,357 @@
...
+
+        let mut changed = false;
+
+        let mut current: EventAdditionalJson = events::table
+            .filter(events::id.eq(id))
+            .select(events::additional_json)
+            .get_result(conn)
+            .to_db_error(ErrorCode::QueryError, "Could not load additional_json for event")?;
+
+        macro_rules! check_and_update {
```
Krakaw, line 333: Very cool. Are you using the macro here instead of a closure to keep the `changed` var in scope?

dunnock: Actually this is just DRY principle, as that thing repeats 5 times. Thanks!

